Issue #150, opened 2026-04-28
[retrieval] Heat kernel authority signal via precomputed eigenbasis (signed Laplacian)


robotrocketscience:
## Summary

Compute graph-based authority scores per query using the **heat
kernel** `exp(-tL)` over the signed normalized Laplacian, evaluated
through a **precomputed top-K=200 eigenbasis** rather than full
matrix exponential. Mix into the ranking score log-additively at
weight 1.0. Use the existing broker-confidence attenuation hook
(`store.propagate_valence`) to dampen scores flowing through
low-confidence intermediates.

This is the primary authority-scoring lever — the largest single
NDCG@10 gain in synthetic-graph evaluation.

Depends on: signed-Laplacian issue (linked).

## Motivation

BM25 alone is authority-blind: it ranks beliefs by topic relevance
but doesn't see graph structure. PageRank, Personalized PageRank,
and the heat kernel are three options for layering authority on
top. Synthetic-graph evaluation found heat kernel beats both
PageRank variants at this scale:

- Heat kernel has an explicit bandwidth knob `t` (PR has fixed
  α=0.85 damping).
- Heat kernel on the signed Laplacian propagates negative
  authority through `CONTRADICTS` edges; PR on directed
  adjacency does not.
- Heat kernel via eigenbasis is both faster and slightly higher
  quality than Krylov-based `expm_multiply` at our scale.

Polynomial methods (Taylor, Chebyshev) also work in principle but
fail without a spectral-radius probe because signed Laplacians can
have eigenvalues outside `[0, 2]` (Kunegis 2010). The eigenbasis
form sidesteps that issue entirely. A follow-up issue can wire a
polynomial path with adaptive degree for fast-changing corpora.

## Algorithm

Given precomputed eigenpairs `(Λ, U)` from the signed-Laplacian
issue:

```
heat_kernel_score(seeds: np.ndarray, t: float = 8.0) -> np.ndarray:
    """Return heat-kernel authority scores for all beliefs."""
    return U @ (np.exp(-t * Λ) * (U.T @ seeds))
```

`seeds` is the BM25-ranked top-25 distribution over beliefs, with
each seed weighted by its BM25 score and the vector L1-normalized.

The full per-query authority score is then attenuated by the
existing broker-confidence mechanism (a belief's score is multiplied
by `α/(α+β)` of the intermediates it flows through). The existing
hook can be reused without modification.

Final ranking score per query / belief:

```
[...]

Issue #148, opened 2026-04-28
[retrieval] Augmented BM25F (incoming-edge anchor text) + vectorized BM25 sparse matvec


robotrocketscience:
## Summary

Augment FTS5 BM25 retrieval with **incoming-edge anchor text** as a
weighted secondary stream, and replace the per-document Python BM25
loop with a **single sparse matvec** over a precomputed
`(n_docs × n_terms)` term-frequency matrix.

Two complementary changes, one PR. Anchor-text augmentation is the
quality lever; vectorized BM25 is the latency lever. Both touch the
same scoring path.

## Motivation

Standard BM25 over a belief's own text misses two recovery cases:

1. **Vocab-shifted beliefs.** A belief whose text uses different
   jargon than the topic at large is unreachable by topic-vocabulary
   queries — even when its citers describe it in topic-correct
   terms. Brin & Page 1998 §4.5 documented this for hyperlinked
   documents; the same pattern applies to `Edge.anchor_text` in the
   belief graph.
2. **Clean queries also under-weight authority anchors.** When
   multiple beliefs share topic vocabulary, the one cited by N peers
   under "consistent" anchor text is ranked indistinguishably from
   peers cited once.

BM25F (Robertson 2004) addresses both via per-stream weighted field
counts. We use the simplest stream-replication form: each belief's
"document" for retrieval becomes its own text plus its incoming
edges' anchor text replicated by a fixed weight.

The vectorization piece is independent. The current per-document
Python loop is fast enough at the v1.0.x scales aelfrice targets but
scales linearly with corpus size; sparse matvec on a precomputed
inverted index is constant-overhead per query and ~10× faster on
synthetic graphs at N=50k.

### BM25F augmented document

For each belief `b`, the indexed document is:

```
augmented_tokens(b) = tokens(b.content)
                   + concat(tokens(e.anchor_text) for e in incoming_edges(b)) * W
```

Weight `W = 3` ships as the default. Stream-replication (multiplying
the anchor token sequence) is BM25F-compatible because BM25's
length-normalization absorbs the increased token count correctly,
unlike post-saturation score fusion. Synthetic-graph evaluation at
N=50k under a 15%-vocab-shifted regime: rank of a vocab-shifted
relevant belief drops from a mean of ~132 to ~14 with W=3; quality
also lifts on clean queries with no vocab gap (NDCG@10 ~0.61 → ~0.77
on a noise-no-shift query mix).

[...]

PR #211, opened 2026-04-28
feat: signed normalized Laplacian + eigenbasis builder (#149)

### LINKED ISSUE #149: [retrieval] Signed normalized Laplacian + offline eigenbasis (top-K=200) builder
## Summary

Build a **signed normalized Laplacian** over the typed edge graph,
treating `CONTRADICTS` and `SUPERSEDES` edges as negative-weight
links and the supportive edge types (`SUPPORTS`, `CITES`,
`RELATES_TO`) as positive. This is the offline foundation for
graph-based authority scoring: by itself it does nothing visible;
its eigendecomposition feeds the heat-kernel issue (linked).

Pure offline construction. No query-time cost. Ships as a builder +
serialization format with no integration into `retrieve()`. The
heat-kernel issue consumes it.

## Motivation

A graph authority signal that ignores opposition treats `b
CONTRADICTS a` as a positive structural link from `a` to `b`,
because adjacency-only methods don't see edge polarity. PageRank,
random walks, and heat kernels on the unsigned adjacency therefore
**propagate authority through contradictions** — the wrong direction.

Kunegis 2010 introduced the signed normalized Laplacian as the
correct generalization for graphs with negative edges:
contradictions propagate distrust, not trust. The construction is a
small modification of the standard symmetric Laplacian and is
mechanical (no learning, no parameters beyond edge-type weights).

aelfrice's edge schema makes this immediately applicable: the
existing relation families have well-defined polarity.

## Algorithm

Edge weights by type:

```
SUPPORTS, CITES, RELATES_TO, DERIVED_FROM   →  +1.0
CONTRADICTS                                  →  -1.0
SUPERSEDES                                   →  -0.5
```

Build the symmetric weighted adjacency `W_sym = W + W.T` from the
typed adjacency. Per Kunegis 2010, the signed normalized Laplacian
is constructed using the **absolute** degree:

```
deg_i = sum_j |W_sym[i, j]|
D = diag(deg)
L = I - D^(-1/2) W_sym D^(-1/2)
```

The absolute-value degree is what makes the negative weights
propagate as opposition rather than collapse into the normalization.

Eigendecomposition `L = U Λ U.T` is computed offline and stored.
Top-K=200 eigenpairs are sufficient for downstream consumers
because the heat-kernel filter `exp(-tL)` exponentially suppresses
high-frequency modes (at bandwidth `t=8` the K>200 modes contribute
< 1e-3). Synthetic-graph evaluation at N=50k: K=200 produces
essentially exact heat-kernel scores (within 1e-3 of the full
expm).

## Implementation sketch

New module `src/aelfrice/graph_spectral.py`:

- `build_signed_adjacency(store) -> scipy.sparse.csr_matrix`
- `build_signed_normalized_laplacian(W) -> scipy.sparse.csr_matrix`
- `compute_eigenbasis(L, k=200) -> (eigvals: np.ndarray, eigvecs: np.ndarray)`
- `save_eigenbasis(path, eigvals, eigvecs)` /
  `load_eigenbasis(path)` for persistence

Storage: numpy `.npz` next to the SQLite file
(`<store_dir>/aelfrice/graph_eigenbasis.npz`). Rebuild on the same
invalidation callback `RetrievalCache` already subscribes to
(extend the registry rather than duplicate it).

## Synthetic-graph evidence

(Effect is observable only through the heat kernel; this issue's
artifact is the eigenbasis. The downstream issue measures NDCG.)

| size N | adjacency build | Laplacian build | top-200 eigsolve |
| --- | --- | --- | --- |
| 10k | ~0.05s | ~0.05s | ~4.1s |
| 50k | ~0.1s | ~0.1s | ~35s |

Memory: ~80 MB at N=50k (200 × 50k float64 eigenvectors). Scales
linearly in N. At N≥10⁶ block-Lanczos approximate eigsolvers are
the path forward; this issue ships the dense top-K solver from
scipy and a follow-up issue handles N≥10⁶.

## Acceptance criteria

1. `build_signed_adjacency(store)` returns a symmetric sparse matrix
   `W_sym` with `W_sym[i, j] == W_sym[j, i]` for all (i, j).
2. Edge-type → weight mapping is exposed as a module-level constant
   so a future issue can sweep alternative weights without changing
   call sites.
3. `build_signed_normalized_laplacian(W)` produces a matrix `L`
   satisfying:
   - `L = I - D_abs^(-1/2) W D_abs^(-1/2)` where
     `D_abs = diag(sum_j |W[i,j]|)`.
   - `L` is symmetric: `L == L.T` to within float tolerance.
   - The all-ones vector is not necessarily an eigenvector (signed
     graphs differ from unsigned; this is correctness, not a bug).
4. `compute_eigenbasis(L, k=200)` returns the smallest K eigenpairs
   (`scipy.sparse.linalg.eigsh(L, k, which="SM")`).
5. Round-trip serialization (`save_eigenbasis` →
   `load_eigenbasis`) is byte-identical.
6. Rebuild fires on the same store-mutation callback that clears
   `RetrievalCache`. Cached eigenbasis from before the mutation is
   not consumed after.
7. Build is deterministic at fixed seed for the eigsolve.

## Test plan

`tests/test_graph_spectral.py`:
- toy-graph fixture (5 nodes, mix of positive and negative edges):
  asserts `W_sym` symmetry, `L` symmetry, eigvalues real-valued,
  smallest eigenvalue in `[0, 2]` for the unsigned-positive
  subgraph (sanity check vs the signed regime which can exceed).
- determinism: build → save → mutate → build → save → load both;
  confirm second build is fresh and diverges from cached.
- edge-type → weight mapping invariants.
- micro-benchmark gated behind `--run-perf` for the eigsolve.

All deterministic, < 500ms each except the perf-gated benchmark.

## Out of scope

- Online incremental eigenbasis updates. Full rebuild on
  invalidation is the v1.x approach; incremental updates land in
  v2.x if benchmark-justified.
- Block-Lanczos / Krylov-based approximate eigensolvers for
  N≥10⁶. Separate issue.
- Per-corpus tuning of edge-type weights. The default mapping ships
  fixed; a sweep harness is a separate research issue.
- Directed signed Laplacian variants. The symmetric form is
  sufficient for the downstream heat kernel; directed variants are
  research.

## References

- Kunegis 2010, *Spectral Analysis of Signed Graphs*.
- Chung 1997, *Spectral Graph Theory* (normalized Laplacian
  background).
- ARPACK / scipy `eigsh` documentation for the underlying solver.

robotrocketscience:
## Summary

Closes #149. Ships the offline foundation for #150 (heat-kernel
authority signal). No integration into `retrieve()`; no query-time
cost.

- New module `aelfrice.graph_spectral` implementing Kunegis 2010's
  signed normalized Laplacian `L = I - D_abs^(-1/2) W_sym D_abs^(-1/2)`
  over the typed edge graph.
- Edge polarity per spec §Algorithm:
  `SUPPORTS|CITES|RELATES_TO|DERIVED_FROM = +1.0`,
  `CONTRADICTS = -1.0`, `SUPERSEDES = -0.5`. Mapping exposed as
  module-level `EDGE_LAPLACIAN_WEIGHTS` (separate from
  `models.EDGE_TYPE_WEIGHTS` which serves valence propagation
  under a different polarity convention).
- API: `build_signed_adjacency`, `build_signed_normalized_laplacian`,
  `compute_eigenbasis`, `save_eigenbasis`, `load_eigenbasis`, plus
  `GraphEigenbasisCache` for invalidation wiring.
- `GraphEigenbasisCache` registers on the same `add_invalidation_callback`
  registry `RetrievalCache` uses; mutation invalidates both
  in-memory eigenpairs and the persisted `.npz`.
- Adds `MemoryStore.iter_all_edges()` so graph-wide builders don't
  reach into `_conn` directly.

## Test plan

- [x] `tests/test_graph_spectral.py` — 13 tests covering all 7
      acceptance criteria; toy 5-node fixture; deterministic.
- [x] `uv run pytest` — full suite **1476 passed, 5 skipped**
      (no regressions from `iter_all_edges` addition).
- [x] `uv run pyright src/aelfrice/graph_spectral.py
      tests/test_graph_spectral.py src/aelfrice/store.py` — clean.
- [x] Perf gate (`--run-perf`): `test_eigsolve_under_budget_n10k`
      — skipped in regular runs.

## Out of scope

Heat kernel consumer (#150), block-Lanczos for N≥10⁶, online
incremental eigenbasis updates, per-corpus weight sweeps.

### DIFF
--- a/src/aelfrice/graph_spectral.py
+++ b/src/aelfrice/graph_spectral.py
@@ -1,0 +1,225 @@
+# pyright: reportUnknownMemberType=false, reportUnknownArgumentType=false, reportUnknownVariableType=false, reportUnknownParameterType=false, reportMissingTypeStubs=false, reportOptionalSubscript=false, reportAttributeAccessIssue=false, reportGeneralTypeIssues=false, reportConstantRedefinition=false, reportArgumentType=false
+"""Signed normalized Laplacian + offline eigenbasis builder (#149).
+
+Constructs the signed normalized Laplacian
+``L = I - D_abs^(-1/2) W_sym D_abs^(-1/2)``
+over the typed edge graph (Kunegis 2010), where ``CONTRADICTS`` /
+``SUPERSEDES`` count as negative-weight edges and the supportive edge
+families as positive. The top-K=200 eigenpairs of ``L`` are the offline
+foundation consumed by the heat-kernel authority signal (#150). This
+module ships only the builder + serialization; no integration into
+``retrieve()``.
+
+All construction is offline — no query-time cost. Rebuild is wired to
+the same store-mutation invalidation callback that ``RetrievalCache``
+subscribes to (extending the registry, not duplicating it).
+"""
+from __future__ import annotations
+
+from dataclasses import dataclass, field
+from pathlib import Path
+from typing import Final
+
+import numpy as np
+import scipy.sparse as sp
+import scipy.sparse.linalg as spla
+
+from aelfrice.models import (
+    EDGE_CITES,
+    EDGE_CONTRADICTS,
+    EDGE_DERIVED_FROM,
+    EDGE_RELATES_TO,
+    EDGE_SUPERSEDES,
+    EDGE_SUPPORTS,
+)
+from aelfrice.store import MemoryStore
+
+# Edge-type → signed weight mapping for Laplacian construction.
+# Distinct from `models.EDGE_TYPE_WEIGHTS` (used for valence
+# propagation; different polarity convention). Per #149 spec; exposed
+# as a constant so a follow-up sweep issue can override without
+# touching call sites.
+EDGE_LAPLACIAN_WEIGHTS: Final[dict[str, float]] = {
+    EDGE_SUPPORTS: 1.0,
+    EDGE_CITES: 1.0,
+    EDGE_RELATES_TO: 1.0,
+    EDGE_DERIVED_FROM: 1.0,
+    EDGE_CONTRADICTS: -1.0,
+    EDGE_SUPERSEDES: -0.5,
+}
+
+# Top-K eigenpairs retained from the signed Laplacian. K=200 is the
+# spec default: at bandwidth t=8 the heat-kernel filter exp(-tL)
+# suppresses K>200 modes below 1e-3 contribution.
+DEFAULT_K: Final[int] = 200
+
+# Persisted .npz layout version. Bump on incompatible field changes.
+_NPZ_VERSION: Final[int] = 1
+
+
+def build_signed_adjacency(
+    store: MemoryStore,
+    weights: dict[str, float] | None = None,
+) -> tuple[sp.csr_matrix, list[str]]:
+    """Build the symmetric signed adjacency ``W_sym = W + W.T``.
+
+    Returns ``(W_sym, belief_ids)`` where ``belief_ids`` is the
+    canonical row/column ordering used by every downstream consumer
+    in this module. Edges referencing ids absent from
+    ``list_belief_ids()`` are dropped (defensive: foreign-key
+    invariant should prevent this, but a stale graph isn't worth
+    crashing the build).
+    """
+    w_map = EDGE_LAPLACIAN_WEIGHTS if weights is None else weights
+    belief_ids = store.list_belief_ids()
+    n = len(belief_ids)
+    if n == 0:
+        return sp.csr_matrix((0, 0), dtype=np.float64), belief_ids
+
+    idx = {bid: i for i, bid in enumerate(belief_ids)}
+    rows: list[int] = []
+    cols: list[int] = []
+    vals: list[float] = []
+    for e in store.iter_all_edges():
+        if e.src not in idx or e.dst not in idx:
+            continue
+        w = w_map.get(e.type, 0.0)
+        if w == 0.0:
+            continue
+        rows.append(idx[e.src])
+        cols.append(idx[e.dst])
+        vals.append(w)
+
+    W = sp.csr_matrix(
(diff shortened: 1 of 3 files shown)


robotrocketscience:
**Kulili peer review** — solid work overall, two minor flags:

### Non-blocking observations

- **`compute_eigenbasis` uses `which=\"SM\"`.** Correct for smallest-magnitude eigenpairs of the normalized Laplacian, but SM internally requires factorization which can be slow at N=50k. Worth benchmarking against `eigsh(L, k=k_eff, sigma=0, which=\"LM\")` (shift-invert mode for smallest-near-zero) on the perf gate when #150 lands and uses this. Not a fix-now item.
- **No CHANGELOG entry under `[Unreleased]`.** Every other recent merge has touched it. Consider adding a short entry before merge — heat kernel (#150) consumes this offline and a v1.6 reader will want the breadcrumb.
- **`EDGE_LAPLACIAN_WEIGHTS` is a `dict[str, float]`** (mutable). Consider `Mapping[str, float]` parameter type or `MappingProxyType` to discourage in-place mutation by callers. Not a real risk in this codebase, just hygiene.

### What I checked

- Schema match with spec § Algorithm: edge weights, `D_abs` normalization, symmetrization, K=200 default, save/load round-trip layout. All match.
- `iter_all_edges()` helper (`ff1bdcb`) is the right addition — minimal, additive, and the right place for it. The +1 deletion is just removing a now-redundant local query in the same file.
- Invalidation callback subscription + on-mutation file-unlink pattern is consistent with `RetrievalCache` and `BM25IndexCache`.
- Pyright pragma block matches the bm25.py (#148) precedent for numpy/scipy stubs.

Green from my side. Heat kernel (#150) can build on this directly.

PR #215, opened 2026-04-28
fix: suppress stale upgrade banner when installed version has caught up

### LINKED ISSUE #214: Statusline upgrade-banner persists after out-of-band pip upgrade
## Summary

The orange \`⬆ aelfrice X.Y.Z available, run: aelf upgrade\` statusline banner persists after the user actually upgrades aelfrice via the pip command \`aelf upgrade\` printed for them. Reproduced visually: banner shows \`aelfrice 1.4.0 available\` even though the running interpreter is at 1.4.0.

## Root cause

\`aelf upgrade\` is print-only — \`src/aelfrice/cli.py:_cmd_upgrade\` (lines 1451-1511) instructs the user to run a pip command but never modifies the package itself. The cache (\`UpdateStatus.update_available=True, latest=\"1.4.0\"\`) is only cleared when:

1. A subsequent \`aelf upgrade\` invocation runs a fresh PyPI check that returns \`update_available=False\` (line 1505, \`_clear_update_cache()\`), OR
2. The cache TTL expires and a background check refreshes it.

Both paths require explicit re-invocation of \`aelf upgrade\` AND a successful network call. Until then, \`statusline.format_snippet\` (\`src/aelfrice/statusline.py:64\`) shows the stale banner because it only checks the cached \`update_available\` flag — it never compares the cached \`latest\` against the actually-running \`aelfrice.__version__\`.

## Proposed fix

Make \`format_snippet\` self-correcting. After the existing \`update_available\` check, compare \`status.latest\` against the running version (via \`lifecycle.installed_version()\` + \`lifecycle.is_newer\`). If \`installed >= latest\`, suppress the banner regardless of the cached flag.

\`\`\`python
running = installed if installed is not None else installed_version()
if running and not is_newer(status.latest, running):
    return \"\"
\`\`\`

The fix is purely additive (one extra short-circuit), preserves all existing tests, and degrades safely: if \`installed_version()\` returns \"\" for any reason the existing behavior is preserved (default-visible — better to show a stale banner than miss a real update).

## Why rook

Mechanism is contained to one function. Spec is tight. No design judgment around correctness contracts. Builder pattern follows the existing \`installed: str | None = None\` injection idiom for testability.

## Acceptance

1. \`format_snippet\` accepts an optional \`installed: str | None = None\` parameter; default resolves via \`installed_version()\`.
2. When \`status.update_available=True\` and \`status.latest=\"1.4.0\"\` and \`installed=\"1.4.0\"\` (or any version >= 1.4.0), \`format_snippet\` returns \`\"\"\`.
3. When \`status.latest=\"1.5.0\"\` and \`installed=\"1.4.0\"\`, banner is shown (existing case preserved).
4. When \`installed=\"\"\` (defensive: \`installed_version()\` failed), banner is shown (preserve original behavior).
5. All existing tests in \`tests/test_statusline.py\` continue to pass without modification.

## Test plan

Add four new tests to \`tests/test_statusline.py\`:
- \`test_suppressed_when_running_version_caught_up\`
- \`test_suppressed_when_running_version_ahead_of_cache\`
- \`test_visible_when_running_version_behind_cache\`
- \`test_visible_when_installed_unknown\`

## Out of scope

- Restructuring \`aelf upgrade\` to actually run pip itself (deliberately deferred per \`_cmd_upgrade\` docstring: replacing the running package mid-process is unreliable on Windows).
- Network-driven cache refresh; the fix is purely local-state-comparison.

robotrocketscience:
## Summary

- Adds `installed: str | None = None` parameter to `format_snippet` for testability.
- After the existing `update_available` check, compares `status.latest` against the running version via `installed_version()` / `is_newer`; returns `""` when installed >= latest.
- Degrades safely: if `installed_version()` returns `""`, banner remains visible (preserves default behavior).

Closes #214

## Test plan

- [ ] `test_suppressed_when_running_version_caught_up` — installed="1.4.0", latest="1.4.0", expect ""
- [ ] `test_suppressed_when_running_version_ahead_of_cache` — installed="1.5.0", latest="1.4.0", expect ""
- [ ] `test_visible_when_running_version_behind_cache` — installed="1.3.0", latest="1.4.0", expect banner shown
- [ ] `test_visible_when_installed_unknown` — installed="", expect banner shown (safe fallback)
- [ ] All pre-existing `tests/test_statusline.py` tests pass
- [ ] Full suite (1456 tests) passes

### DIFF
--- a/src/aelfrice/statusline.py
+++ b/src/aelfrice/statusline.py
@@ -19,7 +19,7 @@ from __future__ import annotations
 import os
 from typing import Final
 
-from aelfrice.lifecycle import UpdateStatus, read_cache
+from aelfrice.lifecycle import UpdateStatus, installed_version, is_newer, read_cache
 
 # CSS orange #FFA500 in truecolor, 256-color 208 (#FF8700), basic 33.
 ANSI_RESET: Final[str] = "\x1b[0m"
@@ -65,6 +65,7 @@ def format_snippet(
     status: UpdateStatus,
     *,
     env: dict[str, str] | None = None,
+    installed: str | None = None,
 ) -> str:
     """Return the statusline snippet for the given cache status.
 
@@ -72,8 +73,20 @@ def format_snippet(
     returns 'COLOR_ON⬆ aelfrice X.Y.Z available, run: aelf upgrade
     COLOR_OFF │ '. The trailing ' │ ' is the GSD-style separator
     so the snippet composes naturally as a leading badge.
+
+    Self-correcting: even when the cache says update_available=True,
+    the banner is suppressed if the running interpreter is already at
+    or ahead of status.latest.  This prevents the banner from
+    persisting after an out-of-band upgrade that has not yet triggered
+    a fresh PyPI check.  The ``installed`` parameter overrides the
+    running version for testing; when omitted, installed_version() is
+    called.  If installed_version() returns "" (lookup failure) the
+    banner is shown, preserving the default-visible safe fallback.
     """
     if not status.update_available or not status.latest:
+        return ""
+    running = installed if installed is not None else installed_version()
+    if running and not is_newer(status.latest, running):
         return ""
     color_on = _pick_color(env)
     color_off = ANSI_RESET if color_on else ""
--- a/tests/test_statusline.py
+++ b/tests/test_statusline.py
@@ -71,6 +71,30 @@ def test_snippet_includes_upgrade_command() -> None:
     assert statusline.ICON in out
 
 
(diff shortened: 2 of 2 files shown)


robotrocketscience:
Superseded by #239 — same fix for #214 with cleaner self-correcting check in `format_snippet` (compares cached `status.latest` against `installed_version()`, suppresses banner when caught up). #239 is mergeable and CI-green; closing this duplicate to keep one source of truth.

Branch `fix/issue-214-statusline-stale-after-upgrade` left intact in case anything from it is worth cherry-picking later.

PR #245, opened 2026-04-28
fix: extend correction-anchor regex for missed correction/requirement utterances (#225)

### LINKED ISSUE #225: [v1.x] correction.py imperative-marker regex undersized — extending recovers +12pp macro-F1
## Source

Lab campaign: `aelfrice-lab/experiments/orphan-propagation-spike/` R0' (run_R0_prime.py / R0_prime_result.json), 2026-04-28.

## Finding

`src/aelfrice/correction.py` `_IMPERATIVE_RE` (line 25–28) and the supporting term tuples are undersized for real correction/requirement utterances. Extending the imperative pattern to include `supersedes|no longer|the previous` (correction-class) and `do not|must not|cannot|fails if|never` (requirement-class) recovers **+12pp macro-F1** over the current production regex on the labeled corpus, with no training and no LLM call.

Note: some of those tokens already exist in `_NEGATION_TERMS` / `_ALWAYS_NEVER_TERMS`, but they only fire as secondary signals after the imperative gate. The miss is structural — utterances like "the previous instruction supersedes …" don't trigger the imperative category at all and never reach the 2-signal threshold.

## Proposed change

1. Extend `_IMPERATIVE_RE` to include the correction-class anchors (`supersedes|no longer|the previous|instead|actually`) and requirement-class anchors (`must not|cannot|fails if`).
2. Or split into `_IMPERATIVE_RE` + `_CORRECTION_ANCHOR_RE` + `_REQUIREMENT_ANCHOR_RE` so the three categories count as distinct signals toward `CORRECTION_SIGNAL_THRESHOLD`.
3. Re-run the labeled-corpus harness in the lab campaign; gate landing on +macro-F1 ≥ +5pp on a held-out split.

## Acceptance

- Labeled-corpus macro-F1 ≥ baseline + 5pp.
- No regression on the existing precision target on the lab false-positive set.
- Deterministic, no LLM, no new dep.

## Out of scope

Trained classifier path (R1' downstream).

robotrocketscience:
## Summary

- Adds `_CORRECTION_ANCHOR_RE` matching `supersedes|no longer|the previous|instead|actually` (word-boundary, case-insensitive) as a new `correction_anchor` signal in `detect_correction`.
- Adds `_REQUIREMENT_ANCHOR_RE` matching `must not|cannot|fails if|do not|never` as a new `requirement_anchor` signal.
- Both are distinct signal categories — they don't collapse into `negation`, `always_never`, or each other. Sentences like "never push to main" now fire three signals (`always_never`, `negation`, `requirement_anchor`), all independently counted toward `CORRECTION_SIGNAL_THRESHOLD`.
- `_IMPERATIVE_RE` is unchanged.

Closes #225.

## Test plan

- [ ] CI staging-gate green.
- [ ] Sentences from the issue (`"the previous instruction supersedes the README"`, `"do not commit secrets"`, `"never push to main"`) now reach the correction/requirement label instead of falling out as `factual`.
- [ ] No precision regression on the existing positive/negative fixtures in `tests/test_correction_detector.py`.

Local: 1606 passed, 5 skipped.

### DIFF
--- a/src/aelfrice/correction.py
+++ b/src/aelfrice/correction.py
@@ -1,10 +1,12 @@
 """No-LLM correction detector.
 
 Heuristic detector that identifies user corrections / directives by
-counting signal-class hits across seven categories: imperative-verb
+counting signal-class hits across nine categories: imperative-verb
 start, always/never absolutist language, negation, emphasis, prior
-reference, declarative override, and strong directive. A text counts
-as a correction when at least two distinct signals fire (precision
+reference, declarative override, strong directive, correction anchor
+(supersedes/replaces/the previous/instead/actually), and requirement
+anchor (must not/cannot/fails if/do not/never). A text counts as a
+correction when at least two distinct signals fire (precision
 trade-off: single-signal matches have ~60% precision; the explicit
 two-signal threshold trades recall for precision so the explicit
 correct-this-belief path covers the gap).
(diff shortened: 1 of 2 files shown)


robotrocketscience:
Superseded by #249, which closed the same issue with a smaller diff.

Issue #263, opened 2026-04-28
[v2.x] legacy_unknown migration for pre-v2.0 stores


robotrocketscience:
## Summary

Pre-v2.0 stores have beliefs with no `ingest_log` row pointing at
them. The reachability check (#205) flags them all as orphans. To
make the v2.x slices work on real-world stores rather than only on
greenfield v2.0 ones, we need a one-shot migration that synthesizes
a `source_kind=legacy_unknown` log row for every existing belief at
its original `created_at` timestamp.

Per spec at `docs/design/write-log-as-truth.md` § "Costs and risks":
*"Existing v1.x stores have no ingest log. Migration synthesizes
`source_kind=legacy_unknown` rows per existing belief at the
original `belief.created_at`. Acceptable but lossy."*

## Why this matters

Without the migration, three things break on legacy stores:
- Reachability check reports many orphan beliefs that aren't
  actually orphans — they're just pre-v2.0.
- Full-equality replay (#262) excludes `legacy_unknown` rows from
  comparison, so a legacy store with no migration registers as 0%
  matched.
- The v2.x view-flip cannot flip on stores with orphan beliefs,
  because there's nothing in the log to derive them from.

## Proposed change

Add to `_MIGRATIONS` (or a one-shot pass after `_SCHEMA`) that:

1. Reads every belief from `beliefs`.
2. For each, if no `ingest_log` row references its id in
   `derived_belief_ids`, INSERTs one with:
   - `source_kind = INGEST_SOURCE_LEGACY_UNKNOWN`
   - `raw_text = belief.content`
   - `source_path = NULL`
   - `derived_belief_ids = [belief.id]`
   - `ts = belief.created_at`
   - `session_id = belief.session_id` if present, else NULL
3. Tracked via a schema_meta marker so it runs once per store.

## Why rook

Mechanical. No design judgment beyond "use the documented
`legacy_unknown` source_kind." Spec already covers the contract —
this is implementation.

## Acceptance

1. New `_maybe_synthesize_legacy_log_rows(store)` method on
   `MemoryStore`, called once at open time (gated by schema_meta).
2. After migration, reachability check reports 0 orphans on a
   pre-v2.0 store.
3. Idempotent: re-running on a migrated store no-ops.
4. New beliefs inserted post-migration go through normal
   `record_ingest` paths — migration does NOT run on them.
5. Tests in `tests/test_legacy_migration.py`:
   - fresh store: no migration triggers, no rows synthesized.
[...]

Issue #262, opened 2026-04-28
[v2.x] Implement replay_full_equality — flip-readiness probe


robotrocketscience:
## Summary

Implement `aelfrice.replay.replay_full_equality()` — the v2.x
flip-readiness probe stubbed in #205. Re-runs the deterministic
derivation function over every `ingest_log` row and compares the
synthesized beliefs to canonical `beliefs`, reporting drift.

This is the validation harness the spec at
`docs/design/write-log-as-truth.md` calls for, and the contract gate
that has to pass before v2.x can flip view authority.

## Background

#205 shipped the parallel-write phase: every belief now has at
least one log row pointing at it (reachability check). The next
step is **fidelity**: under the current rule set, does replaying
the log reproduce the canonical store?

Until #261 lands, there is no pure derivation function to invoke;
this issue is gated on it.

## Proposed change

`replay_full_equality(store, *, scope='all', limit=None)` returns a
`FullEqualityReport` with:

- `total_log_rows`: log rows considered (excluding `legacy_unknown`).
- `derived_match`: log rows where derived belief == canonical belief.
- `derived_mismatch`: rows where the synthesized belief differs from
  canonical (content_hash, type, alpha/beta, origin, or edge set).
- `derived_orphan`: log rows producing a belief that is NOT in
  canonical (something was deleted post-ingest).
- `canonical_orphan`: beliefs in canonical with no derivation in the
  current rule set (something was inserted by a path that no longer
  produces it, or the rules changed).
- `drift_examples`: up to N representative drift cases for
  human inspection (raw_text + canonical belief + derived belief).

CLI surface: extend `aelf doctor --replay` (added in #205 as a
no-op). The flag's stub becomes the real probe.

## Why queen

Several correctness contracts that the spec leaves open:

1. **What "equal" means.** Bit-equal across all fields? Or
   content_hash + type + within-tolerance alpha/beta? alpha/beta
   evolve via feedback after the initial derivation, so strict
   equality is wrong; the right equality is **shape equality at
   ingest time** — same content_hash, same type, same origin —
   ignoring posterior drift. This needs a written contract.

2. **Edge equality.** Edges are derived too. The spec's guarantee
   is unclear about whether edges must replay equal. Most edges
   come from `triple_extractor`, which is deterministic; some come
   from `propagate_valence`, which is feedback-driven and won't
   replay. Need to scope the equality check.

3. **Time-travel scope.** "What would the agent have retrieved on
[...]

PR #270, opened 2026-04-29
refactor: derivation.derive() pure function for ingest pipeline (#261)

### LINKED ISSUE #261: [v2.x prereq] Factor out deterministic belief derivation function
## Summary

Foundation for the v2.x write-log slices that come after #205. Today
the deterministic part of "derive a belief from raw text" is split
across `ingest.py`, `scanner.py`, `triple_extractor.py`,
`classification.py`, and `mcp_server.py` — each call site directly
constructs a `Belief` after running the classifier inline. There is
no callable that takes "(raw_text, source_kind, source_path,
classifier_version, rule_set_hash)" and returns the belief that
*would* be inserted, without the side effect.

This issue extracts that callable. It does not change behavior at
any existing call site; it factors out a function that the call
sites and the v2.x replay harness can both use.

## Motivation

The v2.0 first slice (#205) shipped `ingest_log` parallel-write but
left the derivation function un-factored. As a result,
`replay_full_equality` (the v2.x flip-readiness probe) is a stub —
it has nothing pure to invoke.

Filing this as the **prerequisite** for the v2.x chain. It is rook-
tier (mechanical refactor, no design judgment) and ready to pull
forward — does not depend on any further design.

## Proposed change

New module `src/aelfrice/derivation.py`:

```python
@dataclass(frozen=True)
class DerivationInput:
    raw_text: str
    source_kind: str            # one of INGEST_SOURCE_KINDS
    source_path: str | None
    raw_meta: dict | None
    session_id: str | None
    ts: str
    classifier_version: str | None
    rule_set_hash: str | None

@dataclass(frozen=True)
class DerivationOutput:
    belief: Belief | None       # None when classifier rejects (persist=False)
    edges: list[Edge]
    skip_reason: str | None     # 'persist=False', 'noise', etc.

def derive(input: DerivationInput) -> DerivationOutput:
    """Pure function: raw text in, belief/edges out, no store I/O."""
```

Each existing call site is rewritten to:
1. Build `DerivationInput`.
2. Call `derive(...)`.
3. If `output.belief` is not None: `record_ingest(...)`,
   `insert_belief(output.belief)`, insert any edges.

## Why rook

No design judgment. The split is mechanical:
- Move the classify+belief-construction code into `derivation.py`.
- Per-entry-point: replace inline construction with a `derive()`
  call.
- Tests guarantee equivalence — every existing ingest test continues
  to pass because the produced belief is the same.

## Acceptance

1. `src/aelfrice/derivation.py` exposes `DerivationInput`,
   `DerivationOutput`, and `derive()`.
2. `derive()` has zero `MemoryStore` dependencies — no store
   parameter, no I/O. Pure function of its input.
3. Each of the six v2.0 ingest entry points calls `derive()` instead
   of constructing `Belief` inline.
4. The full pytest suite passes unchanged. No new test is required
   beyond what the existing per-entry-point tests already cover —
   if `derive()` is correct, they continue to pass.
5. New unit test `tests/test_derivation.py` exercises `derive()`
   directly with synthetic inputs covering each `source_kind`.

## Out of scope

- `replay_full_equality` consumption (separate issue).
- Per-call-site behavior changes.
- Schema additions.

## Dependencies

- **Depends on:** #205 (merged) — `ingest_log` exists.
- **Blocks:** the next 4 issues in this chain (replay_full_equality,
  legacy_unknown migration, derivation worker, view-flip).

robotrocketscience:
## Summary

Closes #261 (partial — see Scope below). Adds `src/aelfrice/derivation.py` exposing a pure `derive(DerivationInput) -> DerivationOutput` callable that takes raw text + source metadata and returns a `Belief + Edge[]` (or a skip with reason). Zero `MemoryStore` dependency — same input, same output, no I/O.

The derivation half of the pipeline is now testable in isolation and provides the pure callable the v2.x `replay_full_equality` flip-readiness probe needs. Each call site that goes through `derive()` is rewritten to: build input → derive → check duplicate → record_ingest + insert.

## Changes

- New `src/aelfrice/derivation.py` (213 lines): `DerivationInput`, `DerivationOutput`, `derive()`. Two dispatch paths: classify-based (filesystem/git/python_ast/feedback_loop_synthesis/legacy_unknown) and lock-based (mcp_remember/cli_remember).
- New `tests/test_derivation.py` (319 lines, 23 tests): purity, each source_kind, lock priors, classifier rejection, unknown-kind ValueError.
- Refactored `src/aelfrice/ingest.py` and `src/aelfrice/scanner.py` (regex path) to call `derive()`.

## Scope: 2 of 6 entry points refactored

The issue body asked for all six entry points. Three were left direct-construction; reasons:

1. **`mcp_server.tool_lock` and `cli._cmd_lock`** — use `sha256("lock\x00" + content)` as belief id, a deliberately distinct namespace from `derive()`'s `sha256(source_path \x00 text)` scheme. Routing through `derive()` would compute different ids and silently break belief lookup for existing locked beliefs in production stores. **Data-identity break, not a refactor** — rejected.
2. **`triple_extractor._resolve_or_create_belief`** — uses `sha256("triple\x00normalized_phrase")` id namespace + fixed (1.0, 1.0) priors for noun-phrase entity beliefs. Same id-namespace incompatibility — rejected.
3. **`classification.accept_classifications`** — applies host-LLM classification results that have already been computed. Calling `derive()` here would re-run the regex classifier and potentially override the host's typed results. **Behavior change, not a refactor** — rejected.
4. **Scanner's opt-in `llm_router` path** — returns custom `origin` (e.g. `DOCUMENT_RECENT`) not representable in current `DerivationInput`. The default no-LLM regex path (production default) is fully routed.

These four sites need follow-up design work before they can be unified — id-namespace migration for #1/#2, classifier-replay handling for #3, `DerivationInput` extension for #4. Each is queen-tier, not mechanical refactor. Recommend filing as follow-up issues if the unification is wanted; otherwise leaving them as their own derivation paths is defensible.

## Test plan

- [x] `tests/test_derivation.py`: 23 new tests, all pass.
- [x] Full suite: 1802 passed, 6 skipped, 1 pre-existing failure deselected (`test_serve_raises_clear_error_when_fastmcp_missing` — environment-dependent, fails on `main` HEAD too).
- [x] Rebased onto current `main` (post #266 merge); no conflicts.

## Commits

- `f258fa1` — add derivation.derive() pure function
- `a76fb18` — route ingest through derive()
- `14ac869` — route scanner regex path through derive()

## Out of scope

- `replay_full_equality` consumption (#262).
- `legacy_unknown` migration (#263).
- The 4 unrouted entry points above — non-refactor work, separate issues if wanted.

### DIFF
--- a/src/aelfrice/derivation.py
+++ b/src/aelfrice/derivation.py
@@ -1,0 +1,213 @@
+"""Pure derivation layer: raw text in, Belief + Edges out, no store I/O.
+
+`derive()` is the single place that turns a classified (or fixed-prior)
+text input into a `Belief` dataclass and any accompanying `Edge` objects.
+All six v2.0 ingest entry points delegate to this function so the
+deterministic part of the pipeline is tested and auditable in one place.
+
+Ingest entry points remain responsible for:
+  - store.get_belief() duplicate checks
+  - store.record_ingest() write-log entries
+  - store.insert_belief() / store.insert_edge() persistence
+  - store.record_corroboration() for re-assertions
+  - any store.update_belief() for the lock-upgrade path
+
+What `derive()` owns:
+  - belief-id derivation
+  - content-hash derivation
+  - classifier dispatch (regex or fixed-prior)
+  - alpha / beta / type / lock_level / origin selection
+  - edge list construction (currently empty for most source_kinds;
+    the DERIVED_FROM wiring in ingest_jsonl is caller-side because it
+    spans consecutive turns and depends on prior-turn state)
+"""
+from __future__ import annotations
+
+import hashlib
+from dataclasses import dataclass
+from typing import Final
+
+from aelfrice.classification import classify_sentence
+from aelfrice.models import (
+    BELIEF_FACTUAL,
+    INGEST_SOURCE_CLI_REMEMBER,
+    INGEST_SOURCE_FILESYSTEM,
+    INGEST_SOURCE_FEEDBACK_LOOP_SYNTHESIS,
+    INGEST_SOURCE_GIT,
+    INGEST_SOURCE_KINDS,
+    INGEST_SOURCE_LEGACY_UNKNOWN,
+    INGEST_SOURCE_MCP_REMEMBER,
+    INGEST_SOURCE_PYTHON_AST,
+    LOCK_NONE,
+    LOCK_USER,
+    ORIGIN_AGENT_INFERRED,
+    ORIGIN_USER_STATED,
+    Belief,
+    Edge,
+)
+
+_BELIEF_ID_HEX_LEN: Final[int] = 16
+
+# Source-kinds that go through the regex/LLM classify_sentence path.
+_CLASSIFY_SOURCE_KINDS: Final[frozenset[str]] = frozenset({
+    INGEST_SOURCE_FILESYSTEM,
+    INGEST_SOURCE_GIT,
+    INGEST_SOURCE_PYTHON_AST,
+    INGEST_SOURCE_FEEDBACK_LOOP_SYNTHESIS,
+    INGEST_SOURCE_LEGACY_UNKNOWN,
+})
+
+# Source-kinds that create user-locked beliefs with fixed priors.
+_LOCK_SOURCE_KINDS: Final[frozenset[str]] = frozenset({
+    INGEST_SOURCE_MCP_REMEMBER,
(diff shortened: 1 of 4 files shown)


robotrocketscience:
Superseded by #273 (merged). Same target: extract pure derive() for #261.

Issue #272, opened 2026-04-29
history-scan blocks all new PRs after main merge — scope or update SCAN_RULES_B64


robotrocketscience:
## Symptom

Three open feature PRs (#268, #269, plus separately #261/#263 reruns) fail `history-scan` with `pattern hit in history`. PRs from before the most-recent main merge pass.

## Diagnosis

`staging-gate.yml` `history-scan` job runs:

```
git fetch --no-tags --prune origin '+refs/heads/*:refs/remotes/origin/*'
git log --all --remotes='origin/*' --format='%H...' > /tmp/h
git log --all --remotes='origin/*' -p > /tmp/d
```

i.e. it scans **every branch** the remote knows about. As soon as a new commit hits main containing strings that match `SCAN_RULES_B64`, every subsequent PR's history-scan picks them up via the new main commit (since main is reachable from `--all --remotes`).

The recent main merge (#253, classify-orphans) shipped real LLM-feature code in `src/aelfrice/llm.py`, `cli.py`, `doctor.py` and tests, with legitimate API/SDK identifiers in source. These appear to be what `SCAN_RULES_B64` is hitting now.

Affected open PRs (verified failing history-scan since the #253 merge):
- #268 docs/issue-220-hook-injection-audit
- #269 feat/issue-195-uncertainty-score
- #261 (rerun expected)
- #263 (rerun expected)

PRs whose runs predate the #253 merge passed (#256, #267, #266, #205).

## Options

1. **Update `SCAN_RULES_B64`** to whitelist the legitimate feature-name occurrences (or remove the offending pattern entries entirely if they were aimed at discretion violations that the new feature now legitimately requires).
2. **Scope the scan** to head/PR diff only (`git log origin/main..HEAD -p`) instead of `--all --remotes`. Keeps the discretion check on incoming work; stops cross-PR spillover.
3. **Allow patterns to match in code only outside `src/aelfrice/llm.py` and `tests/test_llm_*.py`** via path-based filtering in the scan script. Most surgical; preserves the current "no leaks anywhere" intent for non-feature paths.

(2) seems best: discretion is a property of incoming changes, not of legitimate feature code that already shipped.

## Repro

```
gh run view <recent-PR-run-id> --log-failed | grep history-scan
```

shows `pattern hit in history` with no detail (intentional — patterns are private).

## Blocker for

All open feature PRs against current main.

robotrocketscience:
[...]

PR #274, opened 2026-04-29
ci: temporary debug workflow for #272 history-scan diagnosis

robotrocketscience:
Diagnostic-only PR for #272. Adds a separate workflow that runs the same history-scan logic but emits per-pattern hit counts plus up to 20 unique file paths — never patterns, lines, or matched content.

Gated to `ci/issue-272-*` branches via `if: startsWith(github.head_ref, 'ci/issue-272')`, so it does not affect any other PR.

After we identify the offending pattern index(es), this PR is closed without merge and the workflow file is deleted on the follow-up fix branch.

Note: the existing `history-scan` job on this branch is expected to fail (that is the bug). Look at the `History Scan Debug (#272)` job's "Diagnose history hits" step output.

### DIFF
--- a/.github/workflows/history-scan-debug.yml
+++ b/.github/workflows/history-scan-debug.yml
@@ -38,14 +38,15 @@ jobs:
             [[ "$p" =~ ^# ]] && continue
             head_hits=$(grep -Ec "$p" /tmp/h || true)
             diff_hits=$(grep -Ec "$p" /tmp/d || true)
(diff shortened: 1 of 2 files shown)


robotrocketscience:
Diagnosis complete via #272 — offending content is the legitimate `Anthropic` SDK wrapper in src/aelfrice/llm_classifier.py + cli.py LLM block, plus four pre-existing main commit subjects ([onboard-llm], Claude Code adapter). Fix: drop SDK-name patterns from SCAN_RULES_B64. Closing without merge; debug workflow removed in follow-up.

PR #268, opened 2026-04-28
docs: audit aelfrice hook injection (#220)

robotrocketscience:
Closes #220 (investigation + recommendations only — no code changes).

## Summary

- Locates and characterises both hook injection paths: `UserPromptSubmit` (`<aelfrice-memory>`) and `PreToolUse` on Grep/Glob (`<aelfrice-search>`).
- Identifies a query-construction mismatch where `OR`-joined tokens are passed through `_escape_fts5_query` and interpreted as AND + literal "OR", explaining the no-match path and cross-domain contamination.
- Documents token-budget cost structure per turn and estimated combined worst-case overhead (~6 300 tokens on a 10-Grep investigation pass).
- Proposes five bounded tuning recommendations, each citing the specific file and line that would change.

## Doc sections

- Overview
- Hook 1: UserPromptSubmit `<aelfrice-memory>` — entry point, behavior, relevance filter, observed gaps
- Hook 2: PreToolUse on Grep/Glob `<aelfrice-search>` — entry point, behavior, relevance filter, observed gaps
- Token budget characterisation
- Recommended tunings (bounded, each citing file:line)
- Out of scope (echoes issue #220's own list)

PR #278, opened 2026-04-29
docs: audit aelfrice hook injection (refs #220)

### LINKED ISSUE #220: [debug] Audit aelfrice hook injection — UserPromptSubmit + PreToolUse on Grep/Glob
## Observation

aelfrice hooks are actively augmenting Claude Code conversations with belief context. Two distinct injection paths confirmed by direct observation in a live session:

1. **UserPromptSubmit hook** — every user message arrives with a leading `<aelfrice-memory>` block listing FTS5-retrieved beliefs relevant to the user's prompt. Visible as snippet ids like `0b130f4c…`, `0dc11de0…`, `084ae663…` at the top of incoming user turns.
2. **PreToolUse hook on Grep/Glob** — every `Grep` and `Glob` tool call gets a leading `<aelfrice-search query="...">` block with FTS5 hits matching the tool's pattern. Effectively `aelf search` runs once per pattern-match tool call before the actual tool fires.

## Why this is worth debugging

- **Possible duplication with `aelf rebuild`.** The UserPromptSubmit injection looks redundant against an explicit rebuild call from the agent. Need to confirm whether one is supposed to subsume the other.
- **PreToolUse on every Grep is aggressive.** A single investigation pass can fire 10+ greps; each one triggers an FTS5 query over the brain-graph DB. Latency budget and corpus-volume effects unmeasured.
- **Snippet relevance.** The injected blocks sometimes return memories from completely unrelated domains (e.g. "head of X government" trivia, the alpha-seek dual-matrix tables) on aelfrice-internal coding tasks. Either the relevance filter is wrong or the corpus is mixing domains.
- **No-op trigger detection.** The PreToolUse hook sometimes injects `no results (store has 20852 belief(s); your query is not in the indexed graph)` — i.e. fires regardless of whether it has anything useful to add. Cheap negative path is fine, but worth measuring.
- **Hook ordering and budget.** If both hooks fire on the same turn (UserPromptSubmit first, then PreToolUse on the agent's first tool call), the cumulative inject can be substantial. Token-budget impact on long sessions unmeasured.

## Investigation tasks

1. Locate the hook handler scripts (`~/.claude/settings.json` or project-local `.claude/settings.json` references — likely `aelf rebuild` and a Grep/Glob pre-hook script).
2. Document the current behavior: when each hook fires, what payload it injects, how the relevance filter is computed, what the latency cost is.
3. Audit the relevance filter — why are alpha-seek / random-trivia snippets surfacing on aelfrice coding tasks? Either domain isolation is missing or the FTS5 query is too loose.
4. Measure cumulative token-budget impact on a representative session.
5. Decide whether PreToolUse on every Grep/Glob is the right design vs. firing only on user-prompted searches.

## Why rook

Investigation + documentation + maybe a tightening of the relevance filter. Not a design rewrite. The mechanism exists and works; this issue is about characterizing and tuning it.

## Out of scope

- Disabling the hooks. The user has confirmed the injection is wanted; this issue debugs and tunes, doesn't rip out.
- Re-architecting the brain-graph schema.
- Changes to the public-side aelfrice CLI surface.

## References

- `/aelf:doctor` for hook diagnostics.
- `/aelf:setup` configures the hooks.
- Hook config likely in `~/.claude/settings.json`.

robotrocketscience:
Audit + characterisation of the two hook paths that inject aelfrice belief context into live sessions: UserPromptSubmit (per-prompt) and PreToolUse on Grep/Glob (per-tool-call). No code changes; the doc proposes three bounded tuning recommendations.

Replaces the previous PR (#268) which had an unrelated private project codename baked into one of its commits; even after redaction at tip, scoped history-scan still saw it in the PR's incoming range. This branch starts from main with one clean commit.

Closes #220. Refs #272.

### DIFF
--- a/docs/hook-injection-audit.md
+++ b/docs/hook-injection-audit.md
@@ -1,0 +1,470 @@
+# Hook injection audit
+
+**Issue:** [#220](https://github.com/robotrocketscience/aelfrice/issues/220)
+**Scope:** audit and characterisation only — no code changes.
+**Status:** investigation complete; recommendations bounded to tuning.
+**Date:** 2026-04-28
+
+---
+
+## Overview
+
+Two distinct hook paths inject aelfrice belief context into live
+sessions:
+
+1. **UserPromptSubmit — `<aelfrice-memory>` block.** Fires on every
+   user message. Runs a full four-layer retrieval against the raw
+   user prompt and injects a ranked belief block at the top of the
+   turn.
+
+2. **PreToolUse on Grep/Glob — `<aelfrice-search>` block.** Fires
+   before every `Grep` and `Glob` tool call. Lifts the search pattern
+   from `tool_input.pattern`, strips metacharacters, and runs a
+   reduced-budget retrieval against the extracted token set.
+
+Both hooks share the same underlying retrieval stack
+(`aelfrice.retrieval.retrieve`), the same DB path resolution
+(`aelfrice.cli.db_path`), and the same SQLite FTS5 index. They differ
+in trigger surface, query construction, and injected token budget.
+
+The audit also found a **query-construction mismatch** (§ Relevance
+filter audit) that affects both hooks and explains the cross-domain
+contamination reported in the issue.
+
+---
+
+## Hook 1: UserPromptSubmit — `<aelfrice-memory>`
+
+### Entry point
+
+| Item | Detail |
+|---|---|
+| Settings key | `UserPromptSubmit` → `command: /Users/thelorax/.local/bin/aelf-hook` |
+| Config file | `~/.claude/settings.json` |
+| Python entry | `aelfrice.hook:main` → `user_prompt_submit()` |
+| Source | `site-packages/aelfrice/hook.py` line 75 |
+
+The `~/.local/bin/aelf-hook` shim is a one-liner that calls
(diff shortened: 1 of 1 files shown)


robotrocketscience:
Review: nearly clean — one fixup needed.

**Line numbers drift** in the recommendations against current `src/`:

| Claim | Doc says | Actual (src/) |
|---|---|---|
| `_escape_fts5_query` in store.py | 219 | **294** |
| `db_path` in cli.py | 214 | **230** |
| `_l1_hits` in retrieval.py | 646 | **732** |
| `_extract_query` in hook_search_tool.py | 496 | 496 ✓ |

Looks like the audit was written against installed package line numbers (`site-packages/aelfrice/`) rather than `src/`. Since the doc is forward-pointing (recs #1, #3 cite specific line ranges), readers applying the recs will miss the target. Either re-anchor to `src/` or drop the `line N` and just point at the function name.

Otherwise: technical claims hold (verified `_escape_fts5_query` whitespace-tokenises and quote-wraps; the OR-keyword AND-trap is real), CI green, no discretion hits, the audit is the most thorough characterisation of the injection paths I've seen.

Push the line-number fix and this is mergeable. Dropping review label.

Issue #219, opened 2026-04-28
Re-ingest creates 5.3x row inflation; FTS5 retrieval injects near-duplicate beliefs


robotrocketscience:
## Symptom

The user's production aelfrice store contains 20,852 belief rows on only 3,913 distinct `content_hash` values — a 5.3× row inflation factor. Re-ingest events (re-running `aelf onboard`, transcript-ingest, etc.) create new belief rows with new `id`s but identical content rather than being consolidated.

The lab/private store shows the same pattern: 25,874 rows on 5,404 distinct hashes (4.8×).

## Distribution

Aelfrice public-product store, rows-per-`content_hash` distribution:

| rows per hash | n hashes |
|---|---|
| 1 | 1,964 |
| 2 | 154 |
| 3 | 19 |
| 4 | 119 |
| 5 | 276 |
| 8 | 37 |
| 10 | 1,307 |
| 20 | 9 |
| 140 | 22 |

Top duplicates (140 rows each) all have `n_origins=1`, `n_sessions=0`, with `created_at` spanning 2026-04-26 21:37 → 2026-04-28 01:37. Shape consistent with scanner re-runs.

## Why this matters

The `UserPromptSubmit` hook retrieval is BM25-budgeted. When the same content is stored as N rows, FTS5 returns N copies in the top-k, displacing diverse content. Observed live: a single conversation turn produced injection blocks containing 6 near-identical copies of the same code snippet and 10 near-identical copies of the same prose paragraph, saturating the budget.

BM25 IDF computation is also affected: term rarity is computed across documents (rows), so duplicate-content rows dilute the rarity of every term in them, distorting ranking globally.

## Schema state (as of v1.2)

- `beliefs.id` is PRIMARY KEY.
- `beliefs.content_hash` has **no UNIQUE constraint** and **no index**.
- The R5 finding documented elsewhere in the lab (which assumed re-ingest was silently dropped at a `content_hash UNIQUE` constraint) is incorrect against this schema.

## Reproduction

\`\`\`bash
sqlite3 ~/.git/aelfrice/memory.db \\
  "SELECT COUNT(*) AS n_rows, COUNT(DISTINCT content_hash) AS n_hashes FROM beliefs;"
\`\`\`

Run on any populated store. The ratio is the inflation factor.

## What is wanted (sketch — not a fix proposal)

Some combination of:

1. Storage-side dedup: insert path checks `content_hash`, skips or upserts on collision.
[...]

Issue #291, opened 2026-04-29
Rebuild redesign: query understanding (entity extraction, intent, term-weighting)


robotrocketscience:
## Scope (ratified 2026-05-02)

Replace the current rebuild query construction (entity/intent extract → BM25 → pack) with the **R1 + R3 stack** from the lab `feature-rebuild-query-understanding` campaign, behind a `query_strategy` setting. R2 (intent prior) is deferred until a real category taxonomy and belief-side category labels exist; tracked separately. LLM-in-the-loop is rejected by the campaign data — the win comes from cheap deterministic mechanisms, not a model call.

This is the umbrella issue for the public-repo implementation.

## Mechanism

1. **R1 — entity expansion.** Detect capitalised tokens in the raw query, lowercase, append to the BM25 term list (effective 2× qf in the integer-qf BM25 contract). Reuses `src/aelfrice/entity_extractor.py`.
2. **R3 — IDF clip.** Drop query terms below the per-store IDF low-quantile; duplicate (2× via integer qf) terms above the per-store IDF high-quantile. Quantiles are derived per-store from the live BM25 IDF distribution and cached on the store; the synthetic-tuned (1.5, 2.5) constants from R3.5 do **not** transfer (live IDF medians are 7.5–9.5; synthetic median 2.85). Reuses `src/aelfrice/bm25.py` IDF accessors.
3. **Stack order.** `tokenise → R1 entity-expand → R3 IDF-clip on combined term list → BM25`. Deterministic, zero-dep, sub-millisecond.

New surfaces on `BeliefStore`: `idf_table()` and per-store IDF-quantile cache. No schema change. Estimated ~190 LOC for R1 + R3 (R1 ≈ 30, R3 ≈ 100, glue ≈ 60).

## Setting

```toml
[rebuild]
query_strategy = "legacy-bm25"  # default for one minor release
# query_strategy = "stack-r1-r3"  # opt-in
```

`legacy-bm25` is the rollback escape and remains the default until the bench gates clear on real-corpus runs (#288 phase-1b harness). Default flips to `stack-r1-r3` after one clean minor release; `legacy-bm25` removed one minor release after the flip.

## Bench gates (real-corpus, via #288 harness)

- ΔP@10 ≥ **+0.05** absolute vs `legacy-bm25` strategy.
- No subset (style or category) regresses by more than 1σ of baseline.
- Recall@10: no regression.
- Per-rebuild p99 latency ≤ legacy + 5 ms.
[...]

Issue #286, opened 2026-04-29
Context rebuild redesign — scoping (parent)


robotrocketscience:
## Problem

The context rebuilder (UserPromptSubmit / PreCompact hook → `context_rebuilder.rebuild()`) is not working as intended. Real-session evidence captured in #281: 10 identical "Graceful degradation" beliefs surfaced in one rebuild, plus several off-topic beliefs (PreCompact hook installation, vsa-logic R6/R10), plus a verifiably **wrong** belief (`1bc8ab45a40351d9` claiming `insert_belief()` does content-hash dedup) presented with no hedging.

Three failure modes were observed in a single 2026-04-29 session: duplicates, off-topic results when no good match exists, and stale beliefs presented as authoritative. Each has a separate root cause.

## Root design assumption that needs revisiting

> "The rebuilder must always inject something — top-K-by-score regardless of whether top-K is actually relevant."

Everything else is downstream of that. There is no "I don't know — say nothing" path, no relevance floor, no typing of beliefs (stable fact vs transient note vs locked truth), no aging mechanism, and no eval harness measuring rebuild *quality* (v1.4 continuation-fidelity measures downstream behavior, not retrieval precision).

## Four open questions (each filed as a child issue)

1. **#289 Silent-vs-always contract + relevance floor.** Should `rebuild()` return empty when nothing scores above some threshold T? Changes the API contract.
2. **#290 Belief typing + aging policy.** Today's store mixes stable facts, snapshot-of-thought, and transient debugging state with no type axis. No ranker fix produces sane output without this.
3. **#291 Query understanding.** Bag-of-entities → BM25 grabs incidental words from recent turns. Either smarter heuristic or LLM step (the rebuilder has avoided LLM for determinism reasons — pick a lane).
4. **#288 Rebuild eval harness.** Without per-rebuild instrumentation + a fixed-corpus precision metric, no redesign can be claimed "better" than what's there.

## Related work (link, don't duplicate)

- #281 — symptom-level bug report (no output dedup, no relevance floor). The dedup fix is mechanical and lands in #283 / #219 territory; the relevance floor is question 1 above.
- #196 — v2.0 substrate decision (multi-axis vs single-axis uncertainty). Belief typing here is adjacent — typing is *what kind of claim*, not *how confident*.
- #197 — dedup module (v2.0 evaluation). Output-stage dedup vs storage-stage dedup; clarify scope.
[...]

PR #283, opened 2026-04-29
fix: re-ingest row inflation — dedup by content_hash (#219)

robotrocketscience:
## Summary

- Adds `insert_or_corroborate()` to `MemoryStore`: all ingest call sites now check for an existing belief by `content_hash` before inserting. Duplicate content arriving from a different source gets a corroboration row on the existing belief instead of a new row.
- `_maybe_consolidate_content_hash_duplicates()`: one-shot migration that collapses pre-existing duplicate `content_hash` groups on first open, summing alpha/beta, propagating origin and lock precedence, rewriting FKs, and doing a bulk DELETE. Runs in < 2 s on a 20 K-belief store with ~2 K duplicate groups.
- `_maybe_apply_content_hash_unique()`: table-swap migration that adds `UNIQUE(content_hash)` to `beliefs` (SQLite has no `ALTER TABLE ADD CONSTRAINT`). Uses `PRAGMA table_info` to preserve any extra columns from prior `ALTER TABLE` migrations.
- Fresh stores: `_SCHEMA` now declares `content_hash TEXT NOT NULL UNIQUE` so new DBs get the constraint from DDL.

Closes #219

## Commits (5 atomic)

1. `feat: insert_or_corroborate helper on MemoryStore`
2. `refactor: route ingest paths through insert_or_corroborate`
3. `feat: consolidate duplicate content_hash rows on first open`
4. `feat: UNIQUE(content_hash) on beliefs table`
5. `docs: re-ingest dedup contract`

## Test plan

- [ ] `uv run pytest tests/test_insert_or_corroborate.py -v` — 7 tests for the new helper (insert-new, duplicate-hash, corroboration row, count, alpha/beta unchanged, bad source_type, bad type on duplicate)
- [ ] `uv run pytest tests/test_content_hash_consolidation.py -v` — 9 tests for the migration (alpha/beta sum, dupe deleted, only canonical survives, FK rewrite, synthetic corroboration, idempotence, schema_meta marker, origin precedence, lock precedence)
- [ ] `uv run pytest tests/test_content_hash_unique.py -v` — 5 tests for UNIQUE constraint (fresh store rejects dupe raw INSERT, UNIQUE in sqlite_master DDL, existing store gets UNIQUE after migration, marker set, UNIQUE enforced post-migration)
- [ ] `uv run pytest --timeout=30 -q` — full suite: 1849 passed, 8 skipped

PR #293, opened 2026-04-29
fix: rebuild block dedupes by content_hash (#281 partial)

### LINKED ISSUE #281: Context rebuild surfaces duplicate + off-topic beliefs; ranking has no relevance floor or output-dedup
## Symptom

Real-session evidence from a UserPromptSubmit-driven rebuild on 2026-04-29 (operator working on PR review + #219 dedup fix). The injected `<aelfrice-memory>` block contained:

- **10 identical copies** of the "Graceful degradation: Missing deps, stale registries, unreachable DBs — warn and continue, never crash. The local graph is always available." line, ids `077d363880474bec`, `0a76f07b31d6f79d`, `40dfc953b2345cab`, `4e180692ba87f6c6`, `6465c70f6ddd61ab`, `9345a02605531c1b`, `aa0a7ff2e08dbdd7`, `c1506a37afc04c54`, `cab569d1c97d286d`, `fbac45ed3dbd9d47` — all the same content_hash.
- **4 identical copies** of "Continuation fidelity at fixed token budget. A regression band: v1.4.0 must hit ≥80%…" — ids `12e88bd9e3341f6d`, `deee347405220f2b`, `d6e1e46eab58f99e`, `fa8a2bde3282cb07`.
- Several beliefs about PreCompact hook installation, subagent transcript path layouts, and vsa-logic R6/R10 — none of which had any relation to the operator's open work (PR review queue + content_hash dedup branch).
- Zero beliefs surfaced for the actual active topics: PR review, `insert_belief`, `record_corroboration`, `content_hash`, history-scan CI, parallel-session protocol.

### 1. No output-level dedup

The duplicates are partly downstream of #219 (5.3× row inflation) — same content, different `belief_id` because of source-keyed hashing. But even with #219 fixed, the rebuilder packs whatever the BM25 layer returns; if duplicates exist (pre-migration, or from any new dedup gap), it has no defense. Suggest collapsing on `content_hash` before packing, regardless of belief id.

### 2. Ranking returns off-topic results with no relevance floor

The rebuilder appears to (a) extract entities/intents from recent turns, (b) BM25-query the store, (c) pack to budget. There is no apparent **relevance threshold**: when no good matches exist for the actual topic, it falls through to whatever scores best — which seems to be high-frequency boilerplate ("Graceful degradation" appears in many docs; "continuation fidelity" is repeated across research notes). The packed output looks confident and authoritative even when the score is garbage.

This is also evidence that **stale beliefs are surfacing as authoritative**: belief `1bc8ab45a40351d9` (a `<aelfrice-search>` PreToolUse-hook injection on a different query) claimed *"`store.insert_belief()` does content-hash dedup"* — verifiably false against current code (`src/aelfrice/store.py:795`, plain INSERT, no dedup). If trusted, this would have caused the operator to close #219 as a non-issue.

## Suggested investigation

Not prescribing a fix — calling for diagnosis first:

1. **Add a per-rebuild diagnostic log**: write `(query_entity, top-K candidates with BM25 scores, what was packed)` to a side file per rebuild invocation. Without this, we cannot see whether the ranker is doing what we think it's doing. One session's worth of these logs would tell us whether the problem is entity extraction, BM25 scoring, or packing-without-floor.
2. **Check the entity-extraction step** against this session's turns: which entities did it pull? Is it catching common substrate words ("subagent", "transcript", "continue") and missing the real topics ("PR review", "content_hash", "insert_belief")?
3. **Output-level `content_hash` dedup** should be cheap to add unconditionally and would mask the worst symptom while #219 lands. Implementation: dict-keyed by `content_hash`, last-write-wins, before pack-to-budget.
4. **Relevance floor**: candidate threshold below which beliefs are dropped rather than packed. Calibrate against a small captured-corpus eval (the v1.4 continuation-fidelity harness is the existing scaffold).
5. **Stale-belief audit**: belief `1bc8ab45a40351d9` is wrong about current code. How many such snapshot-of-transient-state beliefs are in the store, and how do we age them out? The locked/L0 mechanism handles "user said this is true forever" — there is no symmetric mechanism for "this was true yesterday and isn't anymore."

## Severity

Burning operator tokens on garbage context every UserPromptSubmit, AND occasionally surfacing wrong information as authoritative. The latter is a correctness bug; the former is a budget/UX bug.

## Out of scope here

- The actual fix for #219 (in flight).
- The promotion-trigger / wonder-consolidation work (#228, #229) — those are about *which beliefs get into the store*, not *which beliefs get retrieved per query*. This issue is about the retrieval/ranking stage.

robotrocketscience:
## Summary

Partial fix for #281: surface-level content_hash dedup in `rebuild_v14`. Different `belief_id`s sharing one `content_hash` (re-ingest before #219, multi-source ingest, any future dedup gap) collapse to one entry in the rebuild block. Locked beliefs win on collision.

The issue calls out two distinct failures: (1) no output-level dedup, (2) no relevance floor / off-topic ranking. This PR addresses (1) only — the cheap mask the issue body flags as suggestion #3. Diagnosis of (2) belongs in the redesign track (#286–#291) and is out of scope here.

## What changed

- `src/aelfrice/context_rebuilder.py`: track `seen_hashes` while packing the v1.4 rebuild output. Locked seeds it; session-scoped and L2.5/L1 hits skip on collision.
- `tests/test_context_rebuilder_hook.py`: two regression tests — 10 dup ids collapse to 1; locked wins over an L1 hit with the same content_hash.

## Test plan

- [x] `uv run pytest tests/ -q` → 1835 passed, 8 skipped on 3.13.
- [x] Targeted `pytest -k "281 or ac1"` → 4 passed.

Closes #281 partially. Leaves the relevance-floor / ranking-redesign half tracked under #286.

### DIFF
--- a/src/aelfrice/context_rebuilder.py
+++ b/src/aelfrice/context_rebuilder.py
@@ -287,24 +287,36 @@ def rebuild_v14(
     session_ids: set[str] = {b.id for b in session_hits}
 
     # Pack, accounting tokens. L0 always survives.
+    # Output-level content_hash dedup (#281): different belief_ids can
+    # share a content_hash (re-ingest before #219, multi-source ingest,
+    # or any future dedup gap). Without this, the rebuild block can
+    # surface 10+ identical lines. Locked wins (it's prepended whole);
+    # subsequent tiers skip any hash already packed.
     used: int = sum(_estimate_belief_tokens(b) for b in locked)
     out: list[Belief] = list(locked)
+    seen_hashes: set[str] = {b.content_hash for b in locked}
 
     for b in session_hits:
+        if b.content_hash in seen_hashes:
+            continue
         cost = _estimate_belief_tokens(b)
         if used + cost > token_budget:
             break
         out.append(b)
         used += cost
+        seen_hashes.add(b.content_hash)
 
     for b in non_locked_hits:
         if b.id in session_ids:
             continue  # already surfaced above
+        if b.content_hash in seen_hashes:
+            continue
         cost = _estimate_belief_tokens(b)
         if used + cost > token_budget:
             break
         out.append(b)
         used += cost
+        seen_hashes.add(b.content_hash)
 
     return _format_block(
         recent_turns, out, session_ids, token_budget=token_budget,
--- a/tests/test_context_rebuilder_hook.py
+++ b/tests/test_context_rebuilder_hook.py
@@ -249,6 +249,110 @@ def test_ac1_pure_rebuild_v14_orders_locked_then_session_then_l1(
     assert a < b
 
 
+# --- #281: output-level content_hash dedup --------------------------------
+#
+# Note on test setup: #283 added UNIQUE(content_hash) on beliefs, so two
+# rows with the same content_hash can no longer be inserted via the store
+# API. The dedup loop in rebuild_v14 is a defense-in-depth safety net for
+# (a) legacy stores opened before the consolidation migration runs, and
+# (b) the case where retrieve() and list_locked_beliefs() both surface
+# the same belief id (different rows for the same logical belief in the
+# returned lists). To exercise the dedup branch without violating the
+# UNIQUE constraint, we monkeypatch the retrieval seams to inject
+# duplicate Belief objects that don't actually live in the store.
+
+
+def test_issue_281_dedup_collapses_duplicates_in_retrieved_lists(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    """Same content_hash returned multiple times surfaces once."""
+    store = MemoryStore(str(tmp_path / "m.db"))
+    try:
+        dups = [
+            Belief(
+                id=f"dup{i:02d}",
+                content="graceful degradation: warn and continue",
+                content_hash="SHARED_HASH",
+                alpha=1.0, beta=1.0, type=BELIEF_FACTUAL,
+                lock_level=LOCK_NONE, locked_at=None,
+                demotion_pressure=0,
+                created_at="2026-04-26T00:00:00Z",
+                last_retrieved_at=None,
+            )
+            for i in range(10)
+        ]
+        monkeypatch.setattr(
+            store, "list_locked_beliefs", lambda: [],
+        )
+        monkeypatch.setattr(
+            "aelfrice.context_rebuilder.retrieve",
+            lambda *a, **kw: dups,
+        )
+        monkeypatch.setattr(
+            "aelfrice.context_rebuilder._session_scoped_hits",
+            lambda *a, **kw: [],
+        )
+        block = rebuild_v14(
+            [RecentTurn(role="user", text="graceful degradation")],
+            store,
+        )
+    finally:
+        store.close()
+    appearances = sum(
+        1 for i in range(10) if f'id="dup{i:02d}"' in block
+    )
+    assert appearances == 1, (
+        f"expected 1 surviving dup, got {appearances}"
+    )
+
+
+def test_issue_281_locked_takes_precedence_over_dup_l1_hits(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch,
+) -> None:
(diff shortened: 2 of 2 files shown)


robotrocketscience:
[claim:review:Gylf:2026-04-29T02:21:58Z]

robotrocketscience:
Sending back. PR #283 (merged earlier today) added a UNIQUE constraint on `beliefs.content_hash`, plus `insert_or_corroborate` everywhere. Two consequences:

1. **Tests fail at HEAD.** `tests/test_context_rebuilder_hook.py::test_issue_281_dedup_by_content_hash_across_belief_ids` and `test_issue_281_locked_takes_precedence_over_dup_l1_hits` both call `store.insert_belief` twice with the same `content_hash` — now `sqlite3.IntegrityError: UNIQUE constraint failed: beliefs.content_hash`. CI shows these as the only two failures on pytest 3.13. Tests must use `insert_or_corroborate` (or insert one belief and exercise the rebuild dedup against retrieval-side duplication, not store-side).
2. **Premise is partially obsolete.** Different `belief_id`s sharing one `content_hash` cannot coexist in `beliefs` post-#283. The remaining failure mode is "the same `belief_id` surfaces from multiple lanes (locked + L2.5 + L1)" — that's still worth deduping in the rebuild block, but the test fixtures and PR description need to be reframed around id-collision-across-lanes, not hash-collision-across-ids.

Recommend: rebase on main, drop the cross-id hash collision fixtures, keep the rebuild-side `seen_hashes` mask but exercise it via the same belief_id appearing in two retrieval lanes (or two beliefs with deliberately-equal `content_hash` only achievable via direct SQL inserts that bypass the store API — not preferred). Re-request review when green.

robotrocketscience:
[release:review:Gylf:2026-04-29T02:22:38Z]

robotrocketscience:
[claim:review:Setr:2026-04-29T02:28:08Z]

robotrocketscience:
Review: **substantive test failure** beyond the rebase, both new tests in `tests/test_context_rebuilder_hook.py` (`test_issue_281_dedup_by_content_hash_across_belief_ids` and `test_issue_281_locked_takes_precedence_over_dup_l1_hits`) inject two rows with `content_hash="H1"` via `insert_belief`. That worked before #283 landed; it doesn't anymore — `UNIQUE(content_hash)` rejects the second insert with `sqlite3.IntegrityError`.

The dedup logic itself in `rebuild_v14` looks correct; the test scaffolding is what breaks. Two ways to keep the regression coverage:

1. **Drop the UNIQUE constraint locally for the test fixture.** Construct an in-memory DB without the migration applied, or open a raw `sqlite3.Connection` against a path-based store and `CREATE TABLE beliefs_legacy ...` without UNIQUE. Faithful reproduction of the pre-#283 state, but invasive.
2. **Construct the duplicate state at the candidate-list level, not the store level.** `rebuild_v14` packs from a `hits` list. Pass two synthesized `Belief` objects with `content_hash="H1"` directly into the dedup code path (mock or refactor to expose an internal helper). Avoids the schema constraint entirely. This is what the tests actually need to verify — output-stage dedup, not store-stage state.

Option 2 is closer to the intent (output-stage mask). Option 1 is closer to integration coverage. Either is fine; both also need a rebase since the branch is BEHIND main.

Holding merge. Releasing review claim.

robotrocketscience:
[release:review:Setr:2026-04-29T02:28:53Z]

robotrocketscience:
Rebased on main (resolves BEHIND state) and reworked the tests so they don't violate the UNIQUE(content_hash) constraint that landed in #283. The dedup loop is now exercised by monkeypatching `list_locked_beliefs` and `retrieve` to inject duplicate Belief objects, which is the realistic post-#283 scenario (a single logical belief surfacing on multiple tiers, or a legacy store opened before consolidation runs). Full suite: 1858 passed, 8 skipped locally.

robotrocketscience:
[claim:review:Setr:2026-04-29T03:07:19Z]

robotrocketscience:
[release:review:Setr:2026-04-29T03:08:10Z]

Issue #206, opened 2026-04-28
[v1.6] Orphan finder + targeted Haiku reclassification (aelf doctor --classify-orphans)


robotrocketscience:
## Summary

Add an `aelf doctor --classify-orphans` pass that finds nodes with missing or low-confidence type labels (post-onboard or post-ingest) and runs a targeted batch of Haiku classification calls to fill them in. Strict superset of current onboard — only *adds* LLM calls, never removes them.

## Motivation

Current state:
- `aelf onboard` runs Haiku-batch classification at 99% accuracy on the corpora measured to date (simpler graphs).
- Per-turn `ingest` does not call Haiku; it relies on regex/AST classification, which leaves some nodes as `type=unknown` or low-confidence.
- Over time these accumulate as orphans — unlinked, untyped, unfindable.

A targeted re-classification pass on just the orphan set is cheap (small subset × Haiku $/token) and recovers retrieval quality without a full re-onboard.

## Definition of orphan

Candidate signals (any one qualifies):
- `type = unknown` or NULL after ingest
- Zero outgoing edges of type ≠ CONTAINS (genuinely disconnected from semantic graph)
- α + β below a threshold AND no human-validated edges incident
- Belief content matches no entity in `entity_index` (#143) once that ships

The exact set is a plan-phase decision; ship with one or two of these and iterate.

## Scope

- New CLI: `aelf doctor --classify-orphans [--max N] [--dry-run]`
- Reuses existing `onboard.py` Haiku-batch path; no new LLM machinery.
- Updates beliefs in place (or via `feedback_loop_synthesis` ingest_log row once #205 lands).
- Reports: orphans found, classified, skipped, before/after type distribution.

## Out of scope

- Replacing any Haiku calls with LLM-free propagation — that is the lab spike (separate). This issue keeps Haiku as ground truth.
- Auto-running on every ingest. Manual `aelf doctor` invocation only at v1.6.

## Dependencies

- **Soft:** entity_index (#143) and BFS multi-hop (#144) sharpen the orphan definition; not strict prerequisites.
- **Soft:** ingest_log (#205) gives a clean place to record the synthesis events.
- **Hard:** none. Could ship today against the v1.4 schema.

## Acceptance

- On a corpus with N synthetic orphans (zero-edge or `type=unknown` rows), `aelf doctor --classify-orphans` recovers ≥ 95% to a non-unknown type.
- No regression on the v1.4 onboard accuracy benchmark.
[...]

PR #300, opened 2026-04-29
fix: aelf doctor --gc-orphan-feedback for dangling feedback_history rows (#223)

### LINKED ISSUE #223: feedback_history has 100% orphan belief_ids; audit trail is structurally broken
## Symptom

\`feedback_history\` table has 171 rows. None of the 114 distinct \`belief_id\` values referenced are present in the current \`beliefs\` table:

\`\`\`sql
SELECT COUNT(DISTINCT fh.belief_id) FROM feedback_history fh
LEFT JOIN beliefs b ON fh.belief_id = b.id
WHERE b.id IS NULL;
-- result: 114 (i.e., all 114 distinct ids are orphaned)
\`\`\`

## Why this matters

\`feedback_history\` is the audit trail backing the Beta-Bernoulli posterior (\`apply_feedback\` writes here; \`aelf:feedback\` slash-command writes here). When 100% of those rows reference dropped beliefs:
- No model can join feedback to current beliefs to learn from past feedback events.
- Posterior calibration (the v1.0.0 central claim) cannot be retroactively verified or audited.
- Any future training pipeline that maps `(belief features, feedback) -> outcome` returns the empty join.

## Likely cause

This is probably a downstream consequence of #219 (re-ingest row inflation): re-ingest creates a new belief id for the same content_hash, the old belief is removed (or supplanted), but the feedback_history rows still point at the old id. The feedback survives the GC of its target.

If that diagnosis is right, the fix lives at re-ingest time: when consolidating duplicate-content rows into a canonical id (see #219), update feedback_history.belief_id to the canonical id in the same migration.

## Reproduction

\`\`\`bash
sqlite3 ~/.git/aelfrice/memory.db <<'SQL'
SELECT
  COUNT(*) AS feedback_rows,
  COUNT(DISTINCT fh.belief_id) AS distinct_targets,
  COUNT(DISTINCT b.id) AS targets_still_present
FROM feedback_history fh
LEFT JOIN beliefs b ON fh.belief_id = b.id;
SQL
\`\`\`

## Related

- #219 (probable upstream cause).
- Lab `experiments/phantom-prereqs/LOG.md § 2026-04-28 — T2 R0a`.

robotrocketscience:
## Summary

Closes #223. Pre-#283 re-ingest could leave `feedback_history` rows pointing at a `belief_id` that was later deleted (a fresh duplicate took its place with a new id). The UNIQUE `content_hash` constraint and `insert_or_corroborate` plug the source going forward, but the residue still exists in user databases (issue reporter saw 100% orphan rate).

What ships:

- `store.count_orphan_feedback_events()` / `delete_orphan_feedback_events()` — direct SQL helpers.
- `doctor.gc_orphan_feedback(store, *, dry_run=True)` returning `OrphanFeedbackReport`.
- `aelf doctor --gc-orphan-feedback` flag. Dry-run by default; `--apply` deletes.

Recovery is not attempted. `feedback_history` stores only `belief_id`, not `content_hash`, so the original target's content is unrecoverable. The pass deletes rather than re-links — the only safe op.

Mirrors the `--classify-orphans` pattern from #206.

## Test plan

- [x] `uv run pytest tests/ -q` → 1865 passed, 8 skipped on 3.13.
- [x] `pytest tests/test_doctor_gc_orphan_feedback.py -q` → 11 passed (store helpers, doctor function, formatter, CLI dry-run, CLI apply).
- [x] Discretion grep on full diff → clean.

Refs #223. Source mechanism plugged by #283 / #219.

### DIFF
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -77,8 +77,10 @@ from aelfrice.derivation import DerivationInput, derive
 from aelfrice.doctor import (
     classify_orphans as _classify_orphans,
     diagnose,
+    format_orphan_feedback_report as _format_orphan_feedback_report,
     format_orphan_report as _format_orphan_report,
     format_report,
+    gc_orphan_feedback as _gc_orphan_feedback,
 )
 from aelfrice.llm_classifier import (
     ENV_API_KEY as _LLM_ENV_API_KEY,
@@ -2054,6 +2056,8 @@ def _cmd_doctor(args: argparse.Namespace, out: object) -> int:
     """
     if getattr(args, "classify_orphans", False):
         return _cmd_doctor_classify_orphans(args, out)
+    if getattr(args, "gc_orphan_feedback", False):
+        return _cmd_doctor_gc_orphan_feedback(args, out)
     scope = getattr(args, "scope", None)
     exit_code = 0
     if scope in (None, "hooks"):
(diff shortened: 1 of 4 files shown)


robotrocketscience:
[claim:review:Setr:2026-04-29T03:00:38Z]

robotrocketscience:
[release:review:Setr:2026-04-29T03:00:57Z]

PR #349, opened 2026-05-02
fix(store): preserve corroborations across unique-applying migration

### LINKED ISSUE #336: #219 dedup migration silently drops synthetic corroboration rows on two production stores
## Symptom

The one-shot `_maybe_consolidate_content_hash_duplicates` migration (#219) silently no-op'd the synthetic-corroboration insert on two of three production stores. After dedup completes, those stores have:

| store | beliefs (pre-open) | beliefs (post-open) | rows deleted | corroboration rows recorded |
|---|---|---|---|---|
| `~/projects/aelfrice/.git/aelfrice/memory.db` | 6391 | 6391 | 0 | 130 (pre-existing live ingest, untouched) |
| `~/projects/work-stuff/.git/aelfrice/memory.db` | 5496 | 5496 | 0 | 0 |
| `~/projects/aelfrice-lab/.git/aelfrice/memory.db` | 25874 | 5404 | **20470** | **0** ← expected ~20470 `consolidation_migration` rows |

The aelfrice-lab store dedupped 20,470 duplicate beliefs, but **zero** synthetic corroboration rows were inserted. The migration succeeded (marker `content_hash_dedup_complete=2026-05-01T15:43:05.306823+00:00` written), so the duplicates are gone — but the count signal is gone too.

## Root cause (suspected)

The synthetic-insert is wrapped in a `try/except sqlite3.IntegrityError` (`store.py:920-1031`):

```python
try:
    self._conn.executemany(
        "INSERT INTO belief_corroborations (belief_id, ingested_at, source_type, ...) VALUES (?, ?, ?, NULL, NULL)",
        [(canon_id, ts_now, CORROBORATION_SOURCE_CONSOLIDATION_MIGRATION) for canon_id, _ in corroboration_rows],
    )
except sqlite3.IntegrityError:
    # Old-schema store: belief_corroborations.belief_id has INTEGER affinity ...
    pass
```

The comment explains this catches "legacy stores where `belief_corroborations` has INTEGER affinity on `belief_id` (old schema)". But the lab and work-stuff stores were freshly opened with `aelf 1.5.1` — the table was created in this same `__init__` from `_SCHEMA` with `belief_id TEXT NOT NULL REFERENCES beliefs(id) ON DELETE CASCADE`. There is no INTEGER-affinity table to catch.

Hypothesis: the actual `IntegrityError` is FK-related, not affinity-related. By the time the synthetic insert runs, the `UPDATE beliefs SET ...` has set canonical row state but the canonical row's `id` is still valid. However, **`belief_corroborations` has an `ON DELETE CASCADE` FK back to `beliefs(id)`** — and earlier in the same transaction, `UPDATE belief_corroborations SET belief_id = ? WHERE belief_id = ?` rewrites pre-existing FKs. If `PRAGMA foreign_keys=ON` is active and any of the rewrite steps produce an invalid intermediate state, the deferred FK check at COMMIT could fire — but that would abort the whole transaction, not just one statement.

A simpler hypothesis: `executemany` is silently raising on the *first* row whose canonical `belief_id` doesn't pass the FK because the canonical was itself among the to-be-deleted duplicates. (Group canonicalization picks oldest by `created_at`; if `created_at` ordering disagrees with the original `id` ordering, the wrong row could be marked canonical.)

Either way, the catch is too broad — it swallows a real signal-loss bug.

## Impact

- Dedup work itself is correct (canonical rows updated, FKs rewritten, duplicates deleted).
- The `consolidation_migration` source-type signal is lost on affected stores. Anything that consumes corroboration count as a freshness/multiplicity signal will undercount on these stores.
- Idempotency marker is set, so a re-open won't retry. The signal is permanently lost on that store unless the marker is unset and the migration re-run from a snapshot.

## Reproduction

```bash
# Open the lab store with v1.5.1:
cd ~/projects/aelfrice
uv run python -c "from aelfrice.store import MemoryStore; MemoryStore('/Users/thelorax/projects/aelfrice-lab/.git/aelfrice/memory.db')"

# Survey:
sqlite3 /Users/thelorax/projects/aelfrice-lab/.git/aelfrice/memory.db \
  "SELECT (SELECT COUNT(*) FROM beliefs), (SELECT COUNT(*) FROM belief_corroborations WHERE source_type='consolidation_migration');"
# → (5404, 0)
```

## Asks

- Tighten the `except sqlite3.IntegrityError: pass` to either log the count of skipped rows or distinguish FK errors from affinity errors.
- Decide whether to retry the synthetic insert outside the transaction once the duplicates are deleted (FK should pass at that point if canonical id is valid).
- Decide whether stores with a set `content_hash_dedup_complete` marker but `0` synthetic rows + nonzero pre-dedup duplicate count should be flagged for re-migration on next open.

Discovered while closing #190.

robotrocketscience:
Closes #336.

## Root cause

Two interacting bugs:

1. **Cascade-on-DROP wipe.** `_maybe_apply_content_hash_unique` swaps `beliefs` via `CREATE beliefs_new → INSERT SELECT → DROP TABLE beliefs → RENAME`. With `PRAGMA foreign_keys=ON` and `belief_corroborations.belief_id REFERENCES beliefs(id) ON DELETE CASCADE`, the `DROP TABLE beliefs` fires the cascade and deletes every row in `belief_corroborations` — including the synthetic `consolidation_migration` rows that `_maybe_consolidate_content_hash_duplicates` (which runs immediately before the unique-applying migration) had just inserted.
2. **Over-broad swallow.** The dedup pass wrapped its synthetic INSERT in `except sqlite3.IntegrityError: pass`, with a comment claiming this caught a legacy "INTEGER affinity belief_id" path. No such schema variant exists in this codebase — the table has always been created from `_SCHEMA` with TEXT affinity. The blanket pass would have masked any real failure had one fired, but on the actual production stores the synthetic INSERT succeeded; the rows were then wiped by (1).

The issue body suspected (2) and an FK-on-INSERT failure. The actual mechanism is (1) — verified by adding a diag print to the except clause and confirming it never fires while the post-state still shows zero `consolidation_migration` rows.

## Fix

Wrap the unique-applying migration's table swap in `PRAGMA foreign_keys=OFF` / `=ON`. This is the [SQLite-recommended pattern](https://www.sqlite.org/lang_altertable.html#otheralter) for table-mutation migrations. The toggle must execute outside any transaction, so the swap now uses an explicit `BEGIN`/`COMMIT` instead of `with self._conn`.

Also tighten the dedup pass's `except sqlite3.IntegrityError`: the catch is retained as a guard against unforeseen legacy-schema integrity failures, but it now `logging.warning`s with the skipped-row count instead of silent `pass`. Silent count-loss can't recur from this site even if a future schema change introduces a new integrity edge case.

## Regression coverage

Two new tests in `tests/test_content_hash_consolidation.py`:

- `test_synthetic_corroboration_survives_unique_migration` — seeds duplicates with the production-shape `belief_corroborations` schema (with the CASCADE FK that was missing from the existing helper), runs the migration, asserts the right number of `consolidation_migration` rows remain.
- `test_pre_existing_corroborations_survive_unique_migration` — same setup plus pre-seeded live `cli_remember` corroboration rows on the canonical belief; verifies they survive too. This proves the cascade-wipe path is fully closed for all corroboration types, not just the synthetic ones.

Why the existing `test_synthetic_corroboration_inserted` (already in the file) missed the bug: its `_seed_duplicates` helper creates `belief_corroborations` without the `REFERENCES beliefs(id) ON DELETE CASCADE` clause that production stores have. With no FK declared, the cascade never fires — the test's seed schema and the production schema diverged.

## Verification

- `pytest tests/test_content_hash_consolidation.py` — 11 passed (9 previous + 2 new).
- `pytest -q` — 1991 passed, 20 skipped (pre-existing skips, unrelated).
- Manual repro on a fresh tempdir store with 30 duplicates across 10 groups: pre-fix → 20 deleted, 0 corroborations; post-fix → 20 deleted, 20 corroborations.

## Out of scope

- **Recovering the lost rows on the lab store.** The duplicate beliefs that originally produced the 20,470 signal are gone; the count is unrecoverable without restoring from a pre-migration snapshot. Issue body's third "ask" (auto-flag-for-re-migration on stores with the marker set + zero synthetic rows + nonzero pre-dedup duplicate count) is deferred — the heuristic is fragile (a fresh store with no duplicates also has zero `consolidation_migration` rows and a set marker) and the recovery path requires manual intervention. Worth a separate issue if there's appetite for it.

## Summary by Sourcery

Prevent loss of corroboration rows during the content-hash uniqueness migration and improve observability of consolidation failures.

Bug Fixes:
- Ensure foreign key cascades no longer delete corroboration rows during the beliefs table swap that adds UNIQUE(content_hash).
- Preserve pre-existing corroboration records when running the content-hash deduplication and unique-constraint migrations.

Enhancements:
- Log consolidation-migration IntegrityError incidents with the number of skipped synthetic corroboration rows instead of silently swallowing them.

Tests:
- Add a regression seeding helper that mirrors the production belief_corroborations foreign key schema to reproduce the original cascade issue.
- Add tests verifying synthetic consolidation_migration corroborations and pre-existing live corroborations both survive the unique-applying migration.

## Release Notes

* **Bug Fixes**
  * Improved error reporting during content deduplication: detailed warnings are now displayed instead of silently ignoring errors, making troubleshooting easier
  * Strengthened database migration process to reliably preserve data during schema updates, preventing potential data loss scenarios
  * Added regression tests to ensure deduplication and migration reliability remain stable in future releases

### DIFF
--- a/src/aelfrice/store.py
+++ b/src/aelfrice/store.py
@@ -1005,10 +1005,14 @@ class MemoryStore:
             )
 
             # Synthetic corroboration rows: one per duplicate consumed.
-            # Best-effort: on legacy stores where belief_corroborations
-            # has INTEGER affinity on belief_id (old schema), the FK
-            # check may reject TEXT ids. Skip gracefully so the main
-            # dedup work (belief row removal) still completes.
+            # Best-effort: log and skip on IntegrityError rather than
+            # aborting the dedup work, but do not swallow silently —
+            # #336 (silent count-signal loss) was caused by an earlier
+            # blanket `except: pass` here combined with a CASCADE-on-
+            # DROP wipe in the unique-applying migration. The cascade
+            # bug is fixed in `_maybe_apply_content_hash_unique`; this
+            # except is retained only as a guard against unforeseen
+            # legacy-schema integrity failures.
             try:
                 self._conn.executemany(
                     """
@@ -1023,12 +1027,14 @@ class MemoryStore:
                         for canon_id, _dupe_id in corroboration_rows
                     ],
                 )
-            except sqlite3.IntegrityError:
-                # Old-schema store: belief_corroborations.belief_id has
-                # INTEGER affinity; TEXT belief ids fail FK check.
-                # Corroboration rows are signal-only; skip them without
-                # aborting the dedup work.
-                pass
+            except sqlite3.IntegrityError as err:
+                import logging
+                logging.getLogger(__name__).warning(
+                    "consolidation-migration: synthetic corroboration "
+                    "insert failed (%d rows skipped): %s",
+                    len(corroboration_rows),
+                    err,
+                )
 
             # Delete duplicate rows from beliefs and FTS.
             self._conn.execute(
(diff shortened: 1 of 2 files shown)


robotrocketscience:
[claim:review:Kulili:2026-05-02T18:38:32Z]

robotrocketscience:
[claim:review:Setr:2026-05-02T18:38:37Z]

robotrocketscience:
[release:review:Setr:2026-05-02T18:38:41Z]

robotrocketscience:
[release:review:Kulili:2026-05-02T18:39:14Z]

PR #426, opened 2026-05-05
docs(readme): v1.7 row → shipped (BM25F default-on; others opt-in) per #154

yoshi280:
Task 4 of the [#403/#154 v2.0 push workflow](https://github.com/robotrocketscience/aelfrice/issues/154). Updated to reflect the post-flip state.

Updates README.md's roadmap table to reflect the actual v1.7 ship state — including the BM25F default-on flip from PR #430.

## What's stale on main

The current row says v1.7 is **planned**, but components #149/#150/#152/#153/#216 are all merged on `main` and reachable via opt-in feature flags. With PR #428 (Porter stemming) + PR #430 (default-on flip) landing, BM25F (#148) becomes the default L1 lane.

## What ships in this PR

- v1.7 row → `shipped` with the actual flag state:
  - **`use_bm25f_anchors`** — default-on (per #428 + #430)
  - **`use_signed_laplacian`, `use_heat_kernel`, `use_hrr_structural`** — opt-in (placeholder lanes pending wiring)
- v2.0 row tightened — drops the "default-on flip is prereq" note since v1.7 is now shipped.

## Bench evidence reflected

Per [#154 comment 4380967901](https://github.com/robotrocketscience/aelfrice/issues/154#issuecomment-4380967901): post-stemming uplift was **+0.6650 NDCG@k** on the v0.1 retrieve_uplift fixture (30 rows, 6 categories) with no regressions. The bench-gate `test_retrieve_per_flag_no_regression` passes.

## Sequencing

This PR is logically downstream of #428 and #430. If they merge first, this lands cleanly as documentation. If this lands first, it'll briefly say "default-on" while main still has it default-off — a 1-day window at most. Coordinate via the merge order on the operator side.

## Test plan

- [x] README diff: 1 file changed, 2 lines net.
- [ ] Visual review: roadmap table renders cleanly on GitHub.

## Summary by CodeRabbit

* **Documentation**
  * Marked v1.7 as shipped.
  * BM25F anchor-text retrieval set as default-on with a cited NDCG@k uplift.
  * Remaining structural retrieval lanes described as opt-in via feature flags (pending activation wiring).
  * v2.0 remains unchanged.

Issue #413, opened 2026-05-04
PR merging / review / CI pipeline too strict


robotrocketscience:
### What happened

```shell
I added a .md file to /docs/ that touched nothing else. 

i dont expect any old text file PR to go through, but it shouldn't be held to the same standard as code, certainly doesnt need code rabbit or sourcely or other limited-usage tools run on it. no regression testing needed, etc. basically just needs the security checks and signature; idk use your best judgment here, but its annoying to add a single text file and get blocked by an authoritarian PR review cycle
```

### What you expected

easier review cycle for text only addition that touches nothing else

### Minimal reproduction

```shell
create a text file that says "hello world" somewhere and try to commit and push and make a PR, then see what happens when you try to merge it.
```

### aelfrice version

latest, but it doesnt matter

### Python version

idk

### Operating system

macos

### Which surface

aelf CLI

### Anything else

the templates are insufficient to capture this type of bug/issue

yoshi280:
[claim:issue:Kulili:2026-05-04T19:36:10Z]

yoshi280:
## Spec memo — differential PR-gate tiers (docs vs code)

Surveyed every workflow under `.github/workflows/` and the `.coderabbit.yaml` config to find what actually fires on a docs-only PR today, then a proposal with ratification questions.

### What runs on every PR today

| Workflow | Job(s) | Path filter? | Skippable for docs-only? |
| --- | --- | --- | --- |
| `ci.yml` | pytest matrix (3.12 + 3.13) | none | yes — no Python touched |
| `staging-gate.yml` | secrets-scan, pattern-scan (PII), history-scan, release-docs-check, commit-msg-prefix, pr-body-issue-link | none | **no** — these ARE the security checks the issue says to keep |
| `deadcode.yml` | deptry / vulture | none | yes |
| `codeql.yml` | static analysis (python, actions) | none | yes for `python`; `actions` only relevant if workflows changed |
| `typos.yml` | crate-ci/typos | none | **no** — typos in docs are exactly what this catches; cheap |
| `zizmor.yml` | workflow lint | already path-filtered to `.github/workflows/**` and `.github/actions/**` | already correct |
| `e2e.yml` | end-to-end | already opt-in via `e2e` label | already correct |
| `mutation.yml` | mutmut | weekly schedule + manual | already correct |
| `link-check.yml` | lychee | weekly schedule + manual | already correct |

Plus the bots:

[...]

PR #442, opened 2026-05-05
ci(workflows): migrate ci.yml to dorny/paths-filter detect job (#413)

yoshi280:
## Closes the docs-PR merge block

Branch protection requires `pytest (3.12)` + `pytest (3.13)` on every PR. The current `ci.yml` uses `paths-ignore` to skip the workflow on docs-only changes — but that makes those checks register as `expected` (i.e., "not satisfied yet, waiting"), which branch protection treats as a blocker. Result: docs-only PRs cannot merge.

Concretely: #426 (README v1.7 row → shipped) hit this today.

The TODO at the old `ci.yml:9-11` calls out exactly this fix: *"once branch protection requires this check, migrate to a dorny/paths-filter detection job so the workflow always runs and reports a status, with downstream jobs gated by `if:`."* This PR does that.

## What changes

- Workflow always runs on every PR (no `paths-ignore`).
- New `detect` job uses `dorny/paths-filter@v3.0.2` (SHA-pinned per repo convention) to classify whether the diff touches Python source / tests / `pyproject.toml` / `uv.lock` / the workflow itself.
- `pytest` matrix gates on `if: needs.detect.outputs.python == 'true'`. When the diff is docs-only, both `pytest (3.12)` and `pytest (3.13)` register as **skipped** — which branch protection accepts as satisfied. When the diff touches Python, both run as before.

## Why this works

Branch protection distinguishes:
- **`expected`** (workflow excluded by `paths-ignore` / never registered) → blocks merge.
- **`skipped`** (job registered but `if:` evaluated false) → satisfies the required check.

The migration shifts gating from "does the workflow run at all" to "does the work inside the workflow run." Registration is unconditional; the work is conditional.

## Refs

- Prior TODO: `.github/workflows/ci.yml:9-11` pre-migration.
- Surfacing PR: #426 (README v1.7 row update; cannot merge under current ci.yml).
- Original `paths-ignore` introduction: #413.

Closes the gap. After this lands, #426 just needs a close+reopen on its existing branch to retrigger CI under the new logic, then it can FF-merge.

## Summary by Sourcery

Migrate the CI workflow to always run on pull requests and use a detection job to conditionally run pytest based on changed files.

CI:
- Replace pull_request paths-ignore configuration with an unconditional trigger so required checks are always registered.
- Introduce a detect job using dorny/paths-filter to determine whether Python-related files changed.
- Gate the pytest job on the detect job output so tests are skipped, not absent, for docs-only or non-Python changes.

### DIFF
--- a/tests/test_slash_commands.py
+++ b/tests/test_slash_commands.py
@@ -45,6 +45,9 @@ EXPECTED_COMMANDS = (
     # v2.0 (#441) — explicit affirmation, sibling of unlock.
     "confirm",
     # v2.0 (#440) — hard-delete escape hatch with confirmation prompt.
+    # Registered here so test_no_extra_files_in_slash_commands_dir and
+    # test_slash_commands_match_visible_cli_subcommands both enforce the
+    # slash file exists and matches the CLI surface.
     "delete",
 )
 


yoshi280:
[claim:review:Toug:2026-05-06T21:09:00Z]

yoshi280:
[release:review:Toug:2026-05-06T21:10:56Z]

yoshi280:
Closed during a rebase attempt that hit a real merge conflict on `.github/workflows/ci.yml` against current main (edge-rerank ship overlap). Branch restored to original SHA 60387b3 and re-opened as #460. GitHub API would not reopen this PR. Sorry for the churn.

PR #460, opened 2026-05-06
ci(workflows): migrate ci.yml to dorny/paths-filter detect job (#413)

yoshi280:
Re-open of #442 (auto-closed during a rebase attempt that hit a real merge conflict in `.github/workflows/ci.yml`; the branch was force-pushed to main's HEAD, then restored to its original commit `60387b3`). The original PR could not be reopened via API.

Refers #413. The diff is the original single signed commit. Author should resolve the merge conflict in ci.yml against current main (edge-rerank ship modified overlapping region) before this can be merged.

## Summary by Sourcery

CI:
- Introduce a dorny/paths-filter-based detect job to determine when Python-related changes require running pytest, replacing the previous paths-ignore-based triggering logic.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,18 +3,16 @@ name: CI
 on:
   pull_request:
     branches: [main]
-    # #413 — docs-only PRs skip pytest. paths-ignore is a UNION filter:
-    # the workflow runs unless every changed file matches an ignored path.
-    # A mixed PR (e.g. src/foo.py + docs/x.md) still runs full CI.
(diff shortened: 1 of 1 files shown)


yoshi280:
Superseded by #446, which shipped the dorny/paths-filter migration to `ci.yml` in v1.7.0 with a different (more compact, inline) implementation. Confirmed by reading current main's `.github/workflows/ci.yml` lines 6-33: the workflow already runs on every PR with an in-job `dorny/paths-filter@v3` step, and pytest gates on the python-changed output. Same goal, already achieved.

Closing rather than rebasing to avoid the duplicate-implementation churn.

PR #513, opened 2026-05-09
feat: imperative /aelf:upgrade — slash rename + detector fix + banner reword

robotrocketscience:
## Summary

Three changes to make the upgrade UX imperative:

1. **`/aelf:upgrade-cmd` → `/aelf:upgrade`** — slash command renamed and rewritten as an imperative orchestrator. Calls `aelf upgrade-cmd --check`, runs the install-aware command in Bash (separate process — no mid-process interpreter replacement), then `aelf setup` (refresh slash bundle), then `aelf upgrade-cmd` (clear stale banner cache). The CLI verb stays `upgrade-cmd` (advisory).

2. **`detect_reachable_installs()` no longer false-positives under `uv run`** — the project venv hosting the running interpreter is now suppressed from `user_local_bin` reporting. Without this, running aelfrice from a project tree via `uv run` triggered a spurious "multiple aelfrice installs detected" warning when there was actually only one install on the user's persistent shell PATH.

3. **Update banner default points at `/aelf:upgrade`** — `format_update_banner(latest)` (no `command` kwarg) now reads `⬆ aelfrice X.Y.Z — run /aelf:upgrade`. Hosts that don't render slash commands can still pass `command=` to embed the literal shell line.

3 atomic commits + 1 gate. Independent of `feat/mcp-server-properly-built`.

## Behavior changes worth flagging

- **`format_update_banner(latest)` (no `command` kwarg)** now returns the slash form. Internal callers (`statusline.format_snippet`, `cli._maybe_emit_update_banner`) already use the no-kwarg form and get the new default. External callers that relied on the literal shell line MUST pass `command=upgrade_advice().command` explicitly.
- **`detect_reachable_installs()` excludes the running interpreter's venv** from `user_local_bin` reports. Callers using the detector to enumerate "every install on this system" no longer count "the one we're running from."
- **CLI surface unchanged.** `aelf upgrade-cmd` (canonical advisory print) and `aelf upgrade` (deprecated alias) both still work as before.

## Test plan

- [x] `tests/test_upgrade_ux.py`: 33 passing (includes 2 new detector-suppression tests + 3 banner tests reflecting the new default)
- [x] `tests/test_slash_commands.py`: 142 passing, 1 skipped (the new `_INVOKE_EXEMPT` entry for `/aelf:upgrade` since it orchestrates rather than passes through)
- [x] Discretion grep clean against `main`
- [x] `aelf upgrade-cmd --check` still works
- [x] `aelf upgrade` deprecated alias still emits warning per #427

[...]

PR #524, opened 2026-05-09
test: add run_doc_linker_uplift for #435 bench gate (#520)

robotrocketscience:
Closes #520. Parent gate: #435.

## What

Implements `run_doc_linker_uplift(rows)` + `DocLinkerUpliftResults` in `tests/retrieve_uplift_runner.py` so `tests/bench_gate/test_doc_linker.py` resolves `runner_mod.run_doc_linker_uplift` cleanly. Adds three unit tests in `tests/test_retrieve_uplift_runner.py` covering the empty-input path, a synthetic-row OFF/ON round-trip, and the no-anchors degenerate case (uplift must be exactly 0).

## Scoring contract (spec § A2)

For each row, two `retrieve_v2` calls on a fresh store, both with `with_doc_anchors=True`:

- **OFF arm:** seed beliefs/edges only, no anchors written.
- **ON arm:** seed beliefs/edges + each entry in `row['anchors']` written via `link_belief_to_document`.

NDCG@k scored against `expected_top_k`. Uplift = `mean_ndcg_on - mean_ndcg_off`. Strictly positive uplift is the ship trigger.

## Caveat — gate broken until rerank lands

The doc-anchor projection at v2.0.0 is read-only: `with_doc_anchors=True` surfaces `DocAnchor` rows alongside beliefs but does not influence ordering. On ranking-invariant corpora the gate will report `uplift ≈ 0` and fail the strict `> 0` assertion. That is the correct gate-broken signal — the gate ships when a downstream rerank consumes the projection.

## Verification

- `uv run pytest tests/test_retrieve_uplift_runner.py` — 14 passed (3 new + 11 existing).
- `uv run pytest tests/bench_gate/test_doc_linker.py` — skipped without `AELFRICE_CORPUS_ROOT` (autouse `bench_gated` marker), as expected.

## Out of scope

- Building the lab-side `tests/corpus/v2_0/doc_linker/` corpus.
- Downstream rerank that consumes `with_doc_anchors` for ordering.
- Bench-evidence sweep (closes #435 separately).

## Summary by Sourcery

Add a doc-linker uplift runner and results type used by the doc_linker bench gate and extend tests to cover its behavior, including empty input and edge cases.

New Features:
- Introduce DocLinkerUpliftResults dataclass to capture doc-linker uplift metrics in a standard shape.
- Add run_doc_linker_uplift driver that computes NDCG-based uplift for doc-linker benchmarks using anchors-off vs anchors-on retrieval runs.

Enhancements:
[...]

PR #525, opened 2026-05-09
fix(#522): upgrade-cmd --check emits run: line on update available

### LINKED ISSUE #522: aelf upgrade-cmd --check missing 'run:' line breaks /aelf:upgrade slash on pre-2.0.1 CLIs
## Summary

The `/aelf:upgrade` slash file (`~/.claude/commands/aelf/upgrade.md`, shipped v2.0.1 via #513) tells the agent in step 1:

> Run `aelf upgrade-cmd --check` to ask aelfrice itself what command would upgrade the active install. Parse the printed `run: <command>` line; if no update is available, print the "up to date" line verbatim and stop.

But empirically `aelf upgrade-cmd --check` on **v2.0.0** does *not* emit the `run: <command>` line when an upgrade is available. Only the no-flags `aelf upgrade-cmd` form does:

```
$ aelf --version
aelf 2.0.0
$ aelf upgrade-cmd --check
aelfrice 2.0.1 available (installed: 2.0.0)
verify: sha256:7f8a4311...
        https://pypi.org/project/aelfrice/2.0.1/

$ aelf upgrade-cmd
aelfrice 2.0.1 available (installed: 2.0.0)
verify: sha256:7f8a4311...
        https://pypi.org/project/aelfrice/2.0.1/
(installed via uv tool — use uv to upgrade)
run: uv tool upgrade aelfrice
```

The `--check` form is missing the install-context note and the `run:` line.

## Affected users

The slash file ships in v2.0.1's bundle, so the asymmetry only bites users who have the v2.0.1 slash bundle deployed against a v2.0.0 (or older) CLI. That happens when:

1. A user runs `aelf setup` from a v2.0.1 source tree before installing v2.0.1 from PyPI (dev install path).
2. A user pulls the slash bundle out-of-band (e.g., from a partial upgrade or a copy from another machine).

Clean PyPI users following the canonical upgrade flow (install 2.0.0 → run old `/aelf:upgrade-cmd` slash → manually `uv tool upgrade aelfrice` → `aelf setup`) are not affected because they don't have the new slash file until *after* the CLI is on 2.0.1.

## On v2.0.1

When the CLI is up-to-date, both forms print the same "up to date" message — symmetric. The asymmetry only shows when an upgrade is actually available, and only against pre-2.0.1 CLIs.

## Proposed fix

`aelf upgrade-cmd --check` should always emit the `run: <command>` line when an upgrade is available, mirroring the no-flags form. The slash file's parser is the contract; the CLI should match it.

Alternative (smaller blast radius): have the slash file fall through to `aelf upgrade-cmd` (no flags) if `--check` doesn't yield a `run:` line. That hardens the slash against any pre-2.0.1 CLI without changing CLI semantics.

Both are one-liner fixes. The first is the more correct contract; the second is the more defensible if anyone is consuming `--check` for non-slash purposes (e.g., scripts that want a quiet exit on no-update).

## Severity

**Minor.** The 2.0.1 release works for clean PyPI upgrade paths; this affects only the developer-install / out-of-band-bundle path. Filed for tracking, not as a v2.0.1 blocker.

## Related

- #513 introduced the `/aelf:upgrade` orchestrator and the parsing contract.
- v2.0.1 ships the orchestrator.

robotrocketscience:
Closes #522.

## Summary
The `/aelf:upgrade` slash file (#513) parses the `run: <command>` line printed by `aelf upgrade-cmd --check`. Pre-#522, `--check` short-circuited that line, so the slash silently produced nothing on pre-2.0.1 CLIs that had received the v2.0.1 slash bundle out-of-band.

Removes the `if not args.check:` guard around the install-context note + `run:` line in `_cmd_upgrade`. Both forms now emit identical output on update-available; the symmetric behavior is what the slash file's parser already assumed. The 'up to date' path is unchanged.

## Affected users
Per the issue body: dev-install paths and out-of-band slash bundle copies. Clean PyPI users not affected. Severity **minor**, not a release blocker.

## Test plan
- [x] 3 new regression tests in `tests/test_upgrade_ux.py`
- [x] Full local suite: 3067 passed, 51 skipped
- [ ] CI staging-gate

## Summary by Sourcery

Align `aelf upgrade-cmd --check` output with the no-flags form when an update is available so consumers can reliably parse the `run:` line.

Bug Fixes:
- Ensure `aelf upgrade-cmd --check` prints the install-context note and `run:` command when an update is available, matching the no-flags behavior.

Enhancements:
- Document the unified `--check` and no-flags upgrade output behavior directly in the `_cmd_upgrade` docstring.

Tests:
- Add regression tests covering `--check` vs no-flags output symmetry and `run:` line emission for both update-available and up-to-date scenarios.

### DIFF
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -2390,9 +2390,13 @@ def _cmd_upgrade(args: argparse.Namespace, out: object) -> int:
     mid-process is unreliable on Windows and can leave the user with a
     broken interpreter. We tell the user the exact line; they run it.
 
-    --check: only print "up to date" / "update available" status,
-    suppress the upgrade command line. Useful for scripts that want
-    a yes/no answer without copy-paste material.
+    --check: previously suppressed the install-context note and the
+    `run:` line. As of #522 both forms emit identical output when an
+    update is available — the `/aelf:upgrade` slash file (#513)
+    parses the `run:` line, so the two forms must agree. Both the
+    no-flags form and `--check` print the version banner, the verify
+    block, the install-context note, and `run: <command>`. The exit
+    code is unchanged: 0 either way.
     """
     advice = upgrade_advice()
     multi_warning = _format_multi_install_warning(
@@ -2425,17 +2429,20 @@ def _cmd_upgrade(args: argparse.Namespace, out: object) -> int:
                 f"        https://pypi.org/project/{_PKG}/{status.latest}/",
                 file=out,  # type: ignore[arg-type]
             )
-        if not args.check:
-            note = _UPGRADE_CONTEXT_NOTE.get(advice.context, "")
-            if note:
-                print(
-                    f"({note})",
-                    file=out,  # type: ignore[arg-type]
-                )
+        # #522: emit the install-context note + `run:` line under both
+        # forms so the /aelf:upgrade slash file (#513) can parse `run:`
+        # whichever form it called. The pre-#522 `--check` short-circuit
+        # broke the slash on pre-2.0.1 CLIs that had the new bundle.
+        note = _UPGRADE_CONTEXT_NOTE.get(advice.context, "")
+        if note:
             print(
-                f"run: {advice.command}",
+                f"({note})",
                 file=out,  # type: ignore[arg-type]
             )
+        print(
+            f"run: {advice.command}",
+            file=out,  # type: ignore[arg-type]
+        )
         return 0
     # No update or unknown -- be explicit.
     if status.latest and not status.update_available:
--- a/tests/test_upgrade_ux.py
+++ b/tests/test_upgrade_ux.py
@@ -545,8 +545,14 @@ def _run_upgrade_cmd_capturing(
 
     import aelfrice.cli as cli_mod
 
+    # Patch on `cli_mod`, not `lifecycle`: `cli.py` does
+    # `from aelfrice.lifecycle import upgrade_advice`, so the bound
+    # name in `cli_mod` is what `_cmd_upgrade` resolves. Patching
+    # `lifecycle` only worked locally on uv-tool installs where the
+    # real call happened to return the same value; CI runs in a venv
+    # and the venv branch fired instead.
     monkeypatch.setattr(
-        lifecycle,
+        cli_mod,
         "upgrade_advice",
         lambda: lifecycle.UpgradeAdvice(
             command="uv tool upgrade aelfrice", context="uv_tool"
--- a/tests/test_upgrade_ux.py
+++ b/tests/test_upgrade_ux.py
@@ -526,3 +526,89 @@ def test_reachable_installs_path_aelf_under_uv_root_not_double_counted(
     sites = lifecycle.detect_reachable_installs()
     assert len(sites) == 1
     assert sites[0].kind == "uv_tool"
+
+
+# ---------------------------------------------------------------------------
+# #522 — `aelf upgrade-cmd --check` must emit `run:` line on update available
+# ---------------------------------------------------------------------------
+
+
+def _run_upgrade_cmd_capturing(
+    monkeypatch: pytest.MonkeyPatch,
+    *,
+    check: bool,
+    update_available: bool,
+) -> str:
+    """Drive _cmd_upgrade with a stubbed update status, return stdout."""
+    import argparse
+    import io
+
+    import aelfrice.cli as cli_mod
+
+    monkeypatch.setattr(
+        lifecycle,
+        "upgrade_advice",
+        lambda: lifecycle.UpgradeAdvice(
+            command="uv tool upgrade aelfrice", context="uv_tool"
+        ),
+    )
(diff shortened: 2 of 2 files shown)


robotrocketscience:
[claim:review:Kulili:2026-05-09T16:54:20Z]

robotrocketscience:
Discretion grep clean (0 hits), commit signed (G), FF-mergeable, 17/29 required checks SUCCESS — but **pytest (3.13) fails on the new test that this PR adds**, blocking merge.

## Failure

`tests/test_upgrade_ux.py::test_upgrade_cmd_check_emits_run_line_on_update_available`:

```
>       assert "run: uv tool upgrade aelfrice" in out
E       AssertionError: assert 'run: uv tool upgrade aelfrice' in
E         'aelfrice 2.0.1 available (installed: 2.0.0)\n
E          verify: sha256:7f8a4311\n
E          ...\n
E          (running inside a virtual environment — use pip to upgrade)\n
E          run: pip install --upgrade aelfrice\n'
```

The implementation does emit a `run:` line — that part of the contract is satisfied — but the assertion locks in one specific upgrade command (`uv tool upgrade aelfrice`). CI runs inside a venv, so `aelf upgrade-cmd` correctly routes to `pip install --upgrade aelfrice`. The assertion is environment-dependent.

## Required changes (pick one)

1. **Loosen the assertion** to match any `run:` prefix:
   ```python
   assert "\nrun: " in out, f"expected a run: line in:\n{out}"
   ```
   This matches the bug-fix contract from #522 (the `/aelf:upgrade` slash file just needs a `run:` line to parse, not a specific command form).

2. **Monkeypatch the env detection** so the test is deterministic. Look at the venv-vs-system-vs-uv-tool branch in `src/aelfrice/cli.py` (or wherever `upgrade-cmd` chooses between `uv tool upgrade` / `pip install --upgrade` / etc) and stub out the path-detection helper to force one branch. Then assert against that branch's specific string.

(1) is the simpler fix and arguably the better contract — #522's spec is that *any* `run:` line is emitted, not a particular command. (2) costs more but also exercises the env-routing logic.

## Other notes

- Once the test passes, the PR is otherwise clean and FF-mergeable. I'm happy to merge after the fix.

Releasing the review claim so the author or another reviewer can re-claim once the fix lands.

robotrocketscience:
[release:review:Kulili:2026-05-09T16:55:44Z]

robotrocketscience:
Rebased onto main (post-#524 merge) and pushed a follow-up commit fixing the CI pytest 3.13 failure.

Root cause: `tests/test_upgrade_ux.py::_run_upgrade_cmd_capturing` patched `lifecycle.upgrade_advice` but `cli.py` does `from aelfrice.lifecycle import upgrade_advice` — the bound name in the cli module is what `_cmd_upgrade` resolves, so the patch on `lifecycle` was a no-op. Locally the user's machine has a uv-tool install of aelfrice, so the real `upgrade_advice()` happened to return uv_tool context and the test silently passed. CI runs `uv run pytest` inside a fresh `.venv` (no `uv tool install`), so `_is_uv_tool_install()` returns False, the venv branch fires, and the test sees `run: pip install --upgrade aelfrice` instead of the asserted `run: uv tool upgrade aelfrice`.

Fix: `monkeypatch.setattr(cli_mod, "upgrade_advice", ...)` instead of `monkeypatch.setattr(lifecycle, "upgrade_advice", ...)`.

Reproduced locally by simulating CI conditions:
```
PYTHONPATH=<sitecustomize forcing _is_uv_tool_install=False> uv run pytest tests/test_upgrade_ux.py -k 'check_emits_run_line or check_and_no_flags or check_silent'
```
- Without fix: 1 failed (matches CI error byte-for-byte), 2 passed
- With fix: 3 passed
- Full suite (no simulation): 3072 passed, 51 skipped

robotrocketscience:
[release:review:rogue1:2026-05-09T17:12:31Z]

robotrocketscience:
Merged via FF push to main as 6eef3ee. All required checks green including pytest 3.13 (regression fix verified).

robotrocketscience:
Status snapshot from a fresh fetch:

- `pytest (3.12)` and `pytest (3.13)` both pass on `6eef3ee` (started 17:09Z, completed green).
- All other checks pass.
- Branch is FF-mergeable: `git merge-base github/main github/fix/issue-522-check-emits-run` returns `a2600b5...` which equals current main HEAD.
- `attn:merge-conflict` label appears stale — likely set while #524 was in flight; the second commit (`fix(test): patch upgrade_advice on cli_mod`) plus #524's merge-then-no-textual-overlap with this PR's files (`src/aelfrice/cli.py`, `tests/test_upgrade_ux.py`) means there is nothing to rebase.

If your local view agrees, label can flip to `attn:review`.

robotrocketscience:
[claim:review:Gylf:2026-05-09T17:21:09Z]

robotrocketscience:
[release:review:Gylf:2026-05-09T17:21:46Z]

PR #559, opened 2026-05-10
feat(wonder): wonder_ingest + wonder_gc lifecycle (closes #229, #548)

robotrocketscience:
Closes #229. Sub-task **C1+C2** of umbrella #542.

Implements the wonder lifecycle that the TODO at `src/aelfrice/cli.py:884` flagged as deferred to v2.x: `wonder_ingest` persists `Phantom` candidates from `aelf wonder` as speculative beliefs, and `wonder_gc` soft-deletes them on a TTL when no evidence accrues.

## What lands

- **`src/aelfrice/wonder/lifecycle.py`** — new module:
  - `wonder_ingest(store, phantoms, session_id=None) -> WonderIngestResult` — writes each `Phantom` as a `type='speculative'` belief with `origin=ORIGIN_SPECULATIVE`, prior α=0.3 / β=1.0, `RELATES_TO` edges to each constituent, and a `wonder_ingest` corroboration row whose `source_path_hash` carries `"<generator>@<score:.4f>"` for audit. Idempotent via SHA-256 of sorted `constituent_belief_ids` as `content_hash` (not text — phantoms with identical text from different pairs are distinct candidates).
  - `wonder_gc(store, ttl_days=14, dry_run=False) -> WonderGCResult` — soft-deletes via `store.soft_delete_belief()`. Preserves any phantom with a `RESOLVES` edge or an α-update beyond the epsilon band.

- **`src/aelfrice/models.py`** — three enum entries + one field:
  - `BELIEF_SPECULATIVE = "speculative"` added to `BELIEF_TYPES`.
  - `EDGE_RESOLVES = "RESOLVES"` added to `EDGE_TYPES` and `EDGE_VALENCE` (0.0 — marker, not evidential).
  - `CORROBORATION_SOURCE_WONDER_INGEST = "wonder_ingest"` added to `CORROBORATION_SOURCE_TYPES`.
  - `Belief.valid_to: str | None` — soft-delete timestamp (NULL = active).

- **`src/aelfrice/store.py`** — additive schema + GC primitives:
  - `ALTER TABLE beliefs ADD COLUMN valid_to TEXT` migration.
  - Partial index `idx_beliefs_speculative_gc ON beliefs(origin, created_at) WHERE valid_to IS NULL` for GC scan performance.
  - `soft_delete_belief(belief_id, ts=None)` — idempotent via `WHERE valid_to IS NULL`.
[...]

Issue #546, opened 2026-05-10
[v2.1] A3 (optional): implicit retrieval-driven α-bump (#542 sub-task)


robotrocketscience:
Sub-task **A3** of umbrella #542 (wonder consolidation — substrate + lifecycle + dispatch).

**Status:** OPTIONAL for v2.1 — may slip to v2.2 without invalidating the umbrella.

## Scope

Implicit retrieval-driven α-bump. Post-retrieval hook schedules a deferred `apply_feedback("retrieval_exposure", ε=0.05)` after a 30-min grace window if no contradiction event fires within that window. New `audit_log` event type to keep this distinguishable from explicit user feedback.

## Why

`apply_feedback` is invoked rarely; the bake-off simulators assumed a feedback firehose that doesn't exist in production. Sub-surface promotion is starved of signal. Implicit retrieval feedback would tap a signal that the schema already records (`last_retrieved_at` populates on ~94% of beliefs) but does not currently couple to posterior updates.

## Why optional

This is the highest-correctness-risk piece of Track A. Failure modes flagged in upstream design: race conditions with explicit feedback, ε tuning without measurement, deferred sweeper scheduling correctness. Recommend gating on A1 + A2 shipping first so we have signal sources before adding a noise source.

## Acceptance (if shipped in v2.1)

- [ ] Post-retrieval hook implemented; deferred sweeper fires after grace window.
- [ ] Grace-window logic correctly suppresses α-bump when an explicit contradiction event lands inside `T_grace`.
- [ ] New `audit_log` event type lands; `apply_feedback` events are distinguishable from explicit feedback.
- [ ] Belief age vs (alpha - alpha_initial) correlation stays in [0.0, 0.5] on a 1-week synthetic workload (else the mechanism becomes a clock instead of a signal).
- [ ] No staging-gate / discretion-grep regressions.

## Blocks / depends on

- **Blocks:** none directly — strengthens future strategy revisits beyond v2.1.
- **Depends on:** A1, A2 (recommended sequencing, not strict).

## Provenance

Re-derive in the public tree. Design rationale informed by private lab notes; no paraphrased examples or fixtures from those notes may appear in the diff.

robotrocketscience:
## Content-stale finding (Faraday, 2026-05-10)

Surfacing for your disposition — not closing.

The deferred-feedback / implicit retrieval-driven α-bump mechanism this issue proposes **already shipped** under **#191 + #256**.

[...]

Issue #191, opened 2026-04-28
[phantom-prereqs T2] Implicit retrieval-driven feedback — sweeper + grace window


robotrocketscience:
## Summary

Track 2 of the phantom-prereqs campaign (#189). Adds a deferred-
feedback queue with a periodic sweeper that converts
"belief was retrieved AND not contradicted within a grace window"
into a small posterior increment on the surfaced belief. Distinct
from explicit user feedback at the audit-log level.

## Problem

`apply_feedback` is invoked rarely in normal use. Any downstream
consumer that wants to use feedback as a posterior-update signal
is starved.

`last_retrieved_at` is populated on the vast majority of beliefs
already — retrieval is happening at volume. The schema records
the event but we don't tap it as a posterior signal today.

### Pieces

1. **New table:** `deferred_feedback_queue(id, belief_id,
   enqueued_at, event_type, applied_at, status)`. `event_type`
   starts with `retrieval_exposure`; the table is built to
   accept future event types behind the same sweeper.

2. **Retrieval-surface enqueue:** post-`retrieve()` hook enqueues
   one `retrieval_exposure` row per surfaced belief, with
   `enqueued_at = now()`.

3. **Sweeper:** periodic job (CLI subcommand suggested:
   `aelf sweep-feedback`) processes queue rows whose
   `enqueued_at` is older than `T_grace` (default 30 minutes).
   For each row:
   - If a contradiction / correction event was recorded against
     the same `belief_id` between `enqueued_at` and
     `enqueued_at + T_grace`, **cancel** the row. Mark
     `status = 'cancelled'`.
   - Else, increment that belief's `alpha` by `epsilon`
     (default 0.05), record the application via existing
     `apply_feedback` machinery, and write a new `audit_log`
     event of type `retrieval_driven_feedback` distinct from
     explicit-user-feedback events.
   - Mark `status = 'applied'`.

4. **Idempotency:** sweeper re-runs over the same queue produce
   identical state. Once applied, a row is not re-applied.
   Once cancelled, a row is not later applied.

### Defaults

- `T_grace`: 30 minutes.
- `epsilon`: 0.05.
- Sweeper schedule: not wired into a daemon at v1; runs as a CLI
  subcommand. Document the recommended scheduling (cron, hook,
  manual). Production scheduling can move to a daemon in a
  follow-up if needed.

## Why queen

The sweeper has correctness questions that the spec doesn't fully
close, and which I expect to surface during implementation:

[...]

PR #562, opened 2026-05-10
test(implicit_feedback): age-vs-alpha-drift Pearson correlation guard (#555)

### LINKED ISSUE #555: [v2.1] follow-up: belief-age vs alpha-drift correlation audit
Carve-out from #546 (closed as duplicate of #191). The implicit retrieval-driven α-bump substrate (deferred-feedback sweeper, `EVENT_RETRIEVAL_EXPOSURE`, post-retrieval enqueue, grace-window suppression) shipped under #191. Unit tests at `tests/test_implicit_feedback.py` cover per-event behavior (+ε once per row, cancel on contradict).

The one acceptance bullet from #546 that does **not** appear to be covered:

> Belief age vs (alpha − alpha_initial) correlation stays in [0.0, 0.5] on a 1-week synthetic workload (else the mechanism becomes a clock instead of a signal).

This is the regression guard that distinguishes "implicit feedback as signal" from "implicit feedback as clock." Without this audit, a future change to `epsilon` / `T_grace` / sweeper-cadence could quietly turn the mechanism into pure age-tracking.

## Scope

- New test or audit script (probably `tests/test_implicit_feedback_age_correlation.py` or `scripts/audit_alpha_age_correlation.py`) that:
  - Generates a synthetic 1-week workload — N beliefs created at varying timestamps, retrieved at varying frequencies; some with explicit feedback, some without.
  - Runs the deferred sweeper end-to-end (or simulates `T_grace` elapse).
  - Computes Pearson correlation between belief age and `(alpha − alpha_initial)`.
  - Asserts correlation ∈ [0.0, 0.5].
- Document the synthetic workload shape so the audit is reproducible.

## Out of scope

- Tuning `epsilon` or `T_grace` — that's a tuning issue, not a guard.
- Production telemetry — this is a synthetic-workload guard, not a live monitor.

## Provenance

Re-derive in the public tree.

robotrocketscience:
Closes #555 — follow-up from #546 (which closed as duplicate of shipped #191).

## What lands

New file: `tests/test_implicit_feedback_age_correlation.py`. Single test that:

1. Builds a synthetic 1-week workload: 200 beliefs spread uniformly across the 7-day window. Per-belief retrieval count ~ Poisson(λ=2), with each retrieval placed at a uniform random time within the belief's lifetime — so retrieval frequency is uncorrelated with age **by construction**.
2. ~15% of beliefs receive an explicit feedback event that cancels their implicit-sweep rows (per the grace-window logic).
3. Drives the deferred sweeper end-to-end: `T_sweep = T_end + 2·T_grace` so every grace window has elapsed.
4. Computes `scipy.stats.pearsonr(age_days, alpha − alpha_initial)`.
5. Asserts the sample r falls in the documented bound.
6. Reproducible: all RNG seeded explicitly (default seed 42).

Measured value at seed 42: **r = -0.035** (well within the bound).

## Spec deviation — please review

**Issue body specifies bound `[0.0, 0.5]`. This test ships `[-0.2, 0.5]`.**

Rationale: the issue body also specifies "retrieval frequency must be uncorrelated with age." With a truly decorrelated workload, the *true* correlation is 0 and the *sample* r at N=200 hovers around 0 with σ ≈ 0.07 (standard error of Pearson r at this sample size). Clamping the lower bound at exactly 0.0 would produce spurious failures on roughly half of seeds (any time noise pushes r slightly negative).

The lower bound `-0.2` (about 3σ below zero) catches a real "systematic anti-correlation" regression — sweeper bias that disproportionately suppresses old beliefs — without firing on sampling noise. **The upper bound `0.5` is preserved exactly as specified** — that's the actual guard for the "implicit feedback becomes a clock" regression the issue is concerned about.

If the operator prefers the strict `[0.0, 0.5]` bound, the test should add multi-seed averaging (e.g. mean r across 30 seeds) so the noise floor doesn't trip the lower edge — that's a more involved change. Flag here so the reviewer can choose.

The test docstring includes the same explanation so a future maintainer can reason about the bound.

## Out of scope

Per the issue body:
- Tuning `epsilon` / `T_grace` (tuning concern, not regression guard).
- Production telemetry (this is a synthetic-workload guard).

## Notes

- Worktree was based on a commit that pre-dated PR #559 landing on main; rebased cleanly onto current `github/main` (`e45fde3`). Single commit, signed.
- Existing `tests/test_implicit_feedback.py` (29 tests) continues to pass — this PR adds coverage, doesn't replace.

## Summary by Sourcery

Tests:
- Introduce a synthetic one-week workload test that asserts Pearson correlation between belief age and alpha drift stays within a safe range, ensuring the sweeper does not effectively act as a clock or penalise older beliefs.

### DIFF
--- a/tests/test_implicit_feedback_age_correlation.py
+++ b/tests/test_implicit_feedback_age_correlation.py
@@ -1,0 +1,217 @@
+"""Regression guard: belief age vs alpha-drift Pearson correlation (#555).
+
+Verifies that the deferred-feedback sweeper does not silently become a clock
+(i.e., older beliefs do not accumulate disproportionate alpha bumps purely
+because they are older, independent of retrieval frequency).
+
+Synthetic workload -- 1 week, N=200 beliefs
+--------------------------------------------
+- N=200 beliefs created uniformly across a 7-day window (one per ~50 min).
+- Retrieval-event counts drawn from Poisson(lam=2) per belief, capped at 10.
+  Each retrieval is placed at a random time within the belief's lifetime.
+  The Poisson draw uses the seeded RNG and is independent of creation order,
+  so retrieval frequency is decorrelated from age by construction.
+- ~15 % of beliefs also receive one explicit positive-feedback event inside
+  the grace window; the sweeper cancels those rows (no alpha change).
+- All retrieval exposures are enqueued with enqueued_at = retrieval_time.
+  The sweeper runs at T_end + 2 * T_grace so every grace window has elapsed.
+- epsilon=0.05, T_grace=1800 s (defaults).
+
+What the correlation bound means
+----------------------------------
+Pearson r in [-0.2, 0.5]:
+  - True correlation is ~0 when retrieval frequency is decorrelated from age.
+    Sampling noise with N=200 gives sigma ~0.07, so -0.2 is ~3 sigma below
+    zero -- a very conservative lower bound that will only fire if the sweeper
+    is systematically penalising older beliefs.
+  - r > 0.5 means the sweeper is a clock: age drives alpha bumps more than
+    retrieval frequency explains. A large increase to epsilon or a dramatic
+    decrease in T_grace relative to the window can push r above 0.5.
+  - The lower bound is intentionally -0.2 (not 0.0): with a decorrelated
+    workload, r will hover near 0 and occasionally go slightly negative due
+    to sampling variation. Clamping at 0.0 would produce spurious failures.
+
+RNG seed: 42.  Change only when adjusting workload shape; document why.
+"""
+from __future__ import annotations
+
+import math
+import random
+from datetime import datetime, timedelta, timezone
+
+from scipy.stats import pearsonr
+
+from aelfrice.deferred_feedback import (
+    DEFAULT_EPSILON,
+    DEFAULT_T_GRACE_SECONDS,
+    enqueue_retrieval_exposures,
+    sweep_deferred_feedback,
+)
+from aelfrice.models import BELIEF_FACTUAL, LOCK_NONE, Belief
+from aelfrice.store import MemoryStore
+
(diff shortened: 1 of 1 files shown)


robotrocketscience:
[claim:review:Maxwell:2026-05-10T05:21:54Z]

robotrocketscience:
Closing this PR per operator review. The substantive issue isn't the bound numbers — it's the choice of correlation measure.

Pearson r only detects **linear** correlation. A non-linear age dependence — e.g. a sweeper bug that disproportionately suppresses both very-young and very-old beliefs (U-shape), or a mid-window plateau — would produce r ≈ 0 and slip past this guard entirely. The "implicit feedback becomes a clock" failure mode the issue is trying to catch is not guaranteed to be linear in age, so a Pearson-only guard is the wrong tool for the job.

The test (and the issue body) need to switch to non-linear correlation measures:

1. **Chatterjee's ξ coefficient** (Chatterjee 2021, *A New Coefficient of Correlation*, JASA). ξ ∈ [0, 1]: 0 iff Y is independent of X, 1 iff Y is a measurable function of X. Detects arbitrary functional dependence including non-monotonic. Available via the `xicor` Python package or implementable from scratch (~10 lines on ranks).
2. **One additional non-linear measure** — likely **distance correlation** (Székely-Rizzo 2007) which is 0 iff independent and detects non-linear/non-monotonic dependence. The `dcor` package implements it. Spearman ρ alone isn't sufficient — it catches monotonic but not U-shaped dependence.

Issue #555 is being reopened with a revised acceptance bullet specifying the correlation measures. The bound numbers will follow from that.

Closing this PR rather than asking for changes because the rewrite is large enough that a fresh PR off updated spec is cleaner.

robotrocketscience:
[release:review:Maxwell:2026-05-10T05:27:01Z]

PR #571, opened 2026-05-10
fix(extraction): preserve underscored identifiers in markdown strip

### LINKED ISSUE #572: extract_sentences mangles snake_case identifiers in multi-token sentences
## Summary

`aelfrice.extraction.extract_sentences` strips Markdown `_italic_` and `__bold__` formatting with unanchored regexes. When a sentence contains 2+ underscored tokens (snake_case modules, file paths, error codes), the first underscore of one token pairs with an underscore in the next, eating the outer underscores and concatenating the spans.

Stored belief content silently desyncs from the literal source text. Query-side `extract_entities` preserves the original form, so the L2.5 entity-index hash lookup never matches and multi-hop retrieval through identifier chains fails.

## Reproduction

```python
from aelfrice.extraction import extract_sentences

extract_sentences("The auth_service.py module imports user_session.py.")
# ['The authservice.py module imports usersession.py.']

extract_sentences("Error E_AUTH_001 raised by auth_service.py.")
# ['Error EAUTH_001 raised by authservice.py.']
```

End-to-end via `ingest_jsonl`: belief content stored as `'The authservice.py module exports SessionPool with EAUTH_001.'` instead of the literal input.

## Mechanism

`extraction.py:46`:

```python
cleaned = re.sub(r"_([^_]+)_", r"\1", cleaned)  # intent: strip _italic_
```

The regex has no word-boundary anchor outside the underscores. `[^_]+` is greedy. With multiple underscored tokens the first `_` of token A pairs with the next `_` (which belongs to token B), eating both outer underscores and the inter-token text.

The same flaw applies to the `__bold__` regex on line 43.

## Impact

Every caller of `extract_sentences` is affected: `ingest_turn` → `extract_sentences` (used by the v1.4 PreCompact hook, transcript ingest, the `aelf remember` CLI, and commit-ingest). Triple extraction reads from belief `content` post-ingest, so triple subjects/objects derived from underscored identifiers are mangled too.

The L2.5 entity-index hash lookup compares query-side entities (preserved literal forms) against store-side entities derived from belief content (mangled forms). On any corpus containing code references with 2+ underscored tokens per sentence, the index returns no hits — multi-hop retrieval through identifier chains silently degrades to L1-only.

## Proposed fix

Anchor with negative lookbehind/lookahead requiring non-alphanumeric context outside the underscores. PR #571 has the patch + 5 regression tests.

## Why this didn't surface earlier

The shipped retrieval benchmark (`aelfrice/benchmark.py`) uses a 16-belief / 16-query toy corpus with no multi-hop coverage — every query has surface-form keywords that appear in its target belief. The L2.5 entity-index can't deliver a measurable lift on that workload, so a regression that disables L2.5 silently produces unchanged numbers.

Worth filing as a follow-up: extend the benchmark with a small multi-hop section that would catch this kind of regression.

robotrocketscience:
Closes #572.

## Summary

`aelfrice.extraction.extract_sentences` strips Markdown `_italic_` and `__bold__` formatting with unanchored regexes. When a sentence contains 2+ underscored tokens (snake_case modules, file paths, error codes), the first underscore of one token pairs with an underscore in the next, eating the outer underscores and concatenating the spans. See #572 for full mechanism + impact analysis.

## Fix

Anchor with negative lookbehind/lookahead requiring non-alphanumeric context outside the underscores. `\b` does not work because `_` is in `\w`; the boundary must specifically exclude alphanumeric on both sides.

```python
cleaned = re.sub(
    r"(?<![A-Za-z0-9])_{2}([^_]*)_{2}(?![A-Za-z0-9])", r"\1", cleaned,
)
cleaned = re.sub(
    r"(?<![A-Za-z0-9])_([^_]+)_(?![A-Za-z0-9])", r"\1", cleaned,
)
```

The asterisk variants (`*italic*`, `**bold**`) are unaffected by the same surface issue (asterisks rarely appear inside identifiers); left untouched.

## Test plan

- [x] 5 new regression cases in `tests/test_extraction.py`:
  - `test_underscore_identifiers_preserved`
  - `test_multiple_underscored_tokens_preserved`
  - `test_underscore_italic_still_stripped`
  - `test_underscore_bold_still_stripped`
  - `test_mixed_italic_and_identifier`
- [x] Existing `test_strips_markdown_headers_and_emphasis` (uses `**` for bold) still passes.
- [x] Full suite: **3140 passed, 53 skipped, 0 failed** (68s).

### DIFF
--- a/src/aelfrice/extraction.py
+++ b/src/aelfrice/extraction.py
@@ -40,10 +40,19 @@ def extract_sentences(text: str) -> list[str]:
     cleaned = re.sub(r"^#{1,6}\s+", "", cleaned, flags=re.MULTILINE)
     # Bold: **text** or __text__
     cleaned = re.sub(r"\*{2}([^*]*)\*{2}", r"\1", cleaned)
-    cleaned = re.sub(r"_{2}([^_]*)_{2}", r"\1", cleaned)
+    # Underscore-bold/italic must be at a true word boundary so
+    # snake_case identifiers and file paths like ``auth_service.py``
+    # are not mangled. Negative lookbehind/lookahead require non-
+    # alphanumeric context outside the underscore (``\b`` does not
+    # work because ``_`` is in ``\w``).
+    cleaned = re.sub(
+        r"(?<![A-Za-z0-9])_{2}([^_]*)_{2}(?![A-Za-z0-9])", r"\1", cleaned,
+    )
     # Italic: *text* or _text_ (single, not double)
     cleaned = re.sub(r"\*([^*]+)\*", r"\1", cleaned)
-    cleaned = re.sub(r"_([^_]+)_", r"\1", cleaned)
+    cleaned = re.sub(
+        r"(?<![A-Za-z0-9])_([^_]+)_(?![A-Za-z0-9])", r"\1", cleaned,
+    )
     # Markdown table rows: lines containing | characters
     cleaned = re.sub(r"^\s*\|.*\|\s*$", " ", cleaned, flags=re.MULTILINE)
     # List markers: leading -, *, +, or numbered list items
--- a/tests/test_extraction.py
+++ b/tests/test_extraction.py
@@ -57,3 +57,54 @@ def test_strips_table_rows_and_list_markers() -> None:
     out = extract_sentences(text)
     assert all("|" not in s for s in out)
     assert any("Intro paragraph" in s for s in out)
+
+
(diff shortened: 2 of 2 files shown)


robotrocketscience:
[claim:review:leibniz:2026-05-10T06:36:24Z]

robotrocketscience:
[release:review:leibniz:2026-05-10T06:39:33Z]

Issue #557, opened 2026-05-10
Claude not using aelf to remember checkpoints, only knows to use it to remember user preferences.


dmayman:
### What problem does this solve

After doing any chunk of work, I want to be able to clear context and keep moving. I'm not getting enough out of aelf automatically to do this. I need to babysit it and make sure it's remembering things. When asking Claude, it had a lot of suggestion for the workflow. 

Here's a screenshot of my expectations from Claude and its description of the limitations of aelfrice.
<img width="2968" height="1318" alt="Image" src="https://github.com/user-attachments/assets/0ad29a08-fce0-441b-9b1c-2b3be7a154d3" />

### How are you handling it today

I'm telling Claude manually to record memories to aelf at checkpoints after doing any big chunks of work.

### Proposed solution

I like Claude's suggestion in the screenshot of recording memories that are not user preferences or beliefs, but just things that happened so that Claude can reference them and the reason behind them, wherever you're at in your project. I know this gets captured in the code and commits but I assume aelfrice has a more granular way of recalling.

robotrocketscience:
[claim:issue:Setr:2026-05-10T05:02:05Z]

robotrocketscience:
## Diagnostic

The complaint reduces to three layered causes. Two are actionable docs/UX bugs; the third is a closed design decision that the README shouldn't have left ambiguous.

### 1. The default-on auto-capture pipeline just shipped — and the README doesn't mention it

v2.1.0 (released today, 2026-05-09) made auto-capture default-on via #529:

- `feat(setup): default-on transcript-ingest, commit-ingest, session-start (#529)` — commit `3c27f45`
- `docs(install): rewrite "Optional hooks" → default-on table (#529)` — commit `24134b6`

Bare `aelf setup` now wires four hook events:
- `UserPromptSubmit` / `Stop` / `PreCompact` / `PostCompact` — log every turn to per-project JSONL at `<git>/aelfrice/transcripts/turns.jsonl`, then ingest into the belief graph at compaction (`src/aelfrice/transcript_logger.py`).
- `PostToolUse:Bash` on successful `git commit` — ingest commit metadata as beliefs (`src/aelfrice/hook_commit_ingest.py`).
- `SessionStart` — inject locked beliefs at session boot.

So for fresh v2.1 installs, session activity *is* captured passively. The reporter's Claude almost certainly read the README and inferred otherwise — `README.md` lines 66–75 (`## What it remembers`) only list manual inputs:

[...]

PR #590, opened 2026-05-10
feat(hook): session-end Stop hook prompts to lock session corrections (#582)

### LINKED ISSUE #582: feat(hook): session-end Stop hook prompts to `aelf:lock` corrections from this session
## Background

User pattern observed across many sessions (memory `6576745a6ef24ca1`):

> 1. Agent attempts approach A. It fails.
> 2. User corrects: "Don't do A, do B instead."
> 3. Agent does B. It works.
> 4. Context fills up or session resets.
> 5. Agent loses memory of steps 1-3.
> 6. Agent attempts approach A again (its prior default).
> 7. User corrects again: "We already tried that. Do B."
> 8. Agent does B again. Works again.
> 9. Context fills up again.
> 10. Repeat from step 5.

This is a load-bearing failure mode — corrections aren't surviving context resets. The system has the mechanism to fix it (`/aelf:lock`), but the mechanism isn't being invoked.

Gap analysis from `aelf wonder --axes`: this domain is **agent_only_dominant**, meaning beliefs about session-management workflow are *paraphrases of user questions* (`agent_inferred`), not user-asserted answers. The user has stated the questions but never *locked* the answers as `lock_level=user`. So the next session inherits the questions, not the conclusions.

## Proposed change

A `Stop` hook (one of the Claude Code hook surfaces, fires when session ends or compacts) that:

1. Queries beliefs created in the current session with `origin=agent_inferred` and `correction:applied` (or similar marker emitted when `aelf correct` fires).
2. If any exist, prepends a stderr block to the session-end summary:
   > Found N corrections in this session that aren't locked. Run `/aelf:lock <id>` to make them survive into the next session, or `/aelf:promote <id>` to upgrade origin without locking.
3. Lists each candidate with its content + suggested lock command pre-filled.

The hook is informational, not blocking. The user (or assistant) reads the prompt and can choose to lock items.

Optional v2: an `auto-lock` config (`AELF_AUTOLOCK_CORRECTIONS=1` env var or TOML key) that auto-locks corrections without prompting. **Default off** — locking is a user-meaning-bearing action and shouldn't happen silently.

## Acceptance criteria

- [ ] Stop hook installed and registered in `~/.claude/settings.json` via `aelf:setup`.
- [ ] Hook fires at session end / PreCompact, queries `feedback_events` table for this session's correction events.
- [ ] If N > 0, emits stderr block listing candidates with pre-filled `/aelf:lock` commands.
- [ ] Block is suppressed (no spam) when N == 0.
- [ ] `AELF_AUTOLOCK_CORRECTIONS=1` env var bypasses prompt, auto-locks; logs each auto-lock to stderr for transparency.
- [ ] `aelf:doctor` checks for the hook's installation, just like the existing UserPromptSubmit hook check.

## Why this is high-leverage

Converting "agent_only_dominant" to "user_validated" is the single intervention that breaks the loop above. Every locked correction is one fewer correction the next session needs.

Pairs naturally with #578: #578 surfaces locked beliefs at session start; this issue ensures more corrections become locked.

## Open design questions

- **What counts as a "correction in this session"?** Two candidate signals:
  - (a) `feedback_events` rows with `kind=correct` and `ts` within session window.
  - (b) Beliefs whose content was modified during the session via `aelf correct` CLI / MCP tool.
  - Recommendation: both, deduplicated.
- **How to identify session boundaries?** The store has `session_id` on beliefs; the Stop hook can read `$AELF_SESSION_ID` if set, fall back to "last 4h" otherwise.
- **Should the hook also prompt to lock high-feedback positive beliefs?** Out of scope for v1; let users start with corrections-only and expand later.

## Related

- Companion to #578 (session-start enrichment of locked beliefs). Together, they form a closed loop: end-of-session captures corrections → start-of-next-session injects them.
- Closes the gap that wonder's gap-analysis surfaced (agent_only_dominant on the parallel-session topic).
- Builds on existing `/aelf:lock` and `/aelf:promote` skills; no new commands needed for v1.

robotrocketscience:
Closes #582.

## What lands

A new `Stop` hook (`aelf-stop-hook`, default-on) that fires once per assistant-turn end, walks the store for unlocked correction-class beliefs created in the current `session_id`, and either emits a stderr listing with pre-filled `aelf lock` commands (default) or auto-locks them when `AELF_AUTOLOCK_CORRECTIONS=1` is set in the environment.

## Spec deviation — read this first

The issue's body assumed a `aelf correct` CLI command and a `feedback_history.kind=correct` marker that **don't exist on `main`** — neither was ever shipped. I confirmed: the only `aelf` subcommands related to belief mutation are `lock`, `confirm`, `delete`, `promote`, `unlock`. The `feedback_history` schema has no `kind` column.

So the issue's two proposed signals — (a) `feedback_events with kind=correct` and (b) `beliefs modified via aelf correct CLI` — are unimplementable as written.

This PR ships the **implementable v0** (operator selected; recorded in #582 reply): query unlocked beliefs in the current session whose `type=BELIEF_CORRECTION` OR `origin in {agent_inferred, agent_remembered}`. That's the closest signal the schema actually exposes for "correction-class belief touched in this session." It does **not** capture beliefs whose existing content was modified — only newly-created ones from the session. A follow-up to ship `aelf correct` + `feedback_history.kind` would let this hook also surface modifications.

## Atomic commits

7 SSH-signed commits, in dependency order:

1. `0f5f2c0` `feat(hook): stop() handler emits lock-prompt for session corrections (#582)` — pure logic in `hook.py`: `stop()` + `_belief_is_lock_candidate` + `_collect_lock_candidates` + `_format_stop_prompt` + `_autolock_enabled` + `_autolock_candidates` + `main_stop()`.
2. `27f9507` `build: register aelf-stop-hook console script (#582)` — pyproject entry point.
3. `39a8ec5` `feat(setup): install/uninstall_stop_hook helpers (#582)` — settings.json wiring under `hooks.Stop`, mirrors `install_session_start_hook` exactly.
4. `4fbfcf6` `feat(cli): wire stop hook into 'aelf setup' / 'aelf unsetup' (#582)` — default-on installation, `--no-stop-hook` opt-out flag on both subcommands.
5. `d61be0a` `feat(doctor): nag when default-on Stop hook is missing (#582)` — extends the v2.1 #557 auto-capture nag to the fourth basename. Updates the doctor parity guardrail test.
6. `098faeb` `test(hook): coverage for stop hook + setup install + doctor nag (#582)` — 41 new tests across two files; updates 2 existing tests for the new fourth basename.
7. `302f7a0` `docs(changelog): unreleased entry for #582 stop hook + lock prompt`.
8. `d7ca88d` `style(hook): use 'harness' phrasing in #582 docstrings (#582)` — tiny phrasing fix to clear the pre-push discretion grep.

## Acceptance

- [x] Stop hook installed and registered in settings.json via `aelf setup` (default-on, idempotent).
- [x] Hook fires at session end, queries store for this session's correction candidates.
- [x] If N > 0, emits stderr block with pre-filled `aelf lock --statement '...'` commands per candidate.
- [x] Block is suppressed when N == 0.
- [x] `AELF_AUTOLOCK_CORRECTIONS=1` env var bypasses prompt, auto-locks candidates, logs each auto-lock to stderr.
- [x] `aelf doctor` checks for the hook's installation (extends the existing #557 nag).
- [x] Coexists with the transcript-ingest Stop entry under the same `hooks.Stop` event key — both are separate entries; basename-uninstall of one does not displace the other (test `test_uninstall_stop_hook_basename_does_not_remove_transcript_ingest`).

## Verification

- `uv run pytest -x -q` → **3220 passed, 53 skipped**. (3179 baseline + 41 new + 2 updated existing.)
- Discretion grep on diff vs main: clean (one remaining "Claude Code" mention in `style(hook)` rephrasing was the only addition; that was the discretion-trip I rephrased away).
- Pre-push hook clean.
- All 8 commits SSH-signed.

## Out of scope (deferred — explicit)

- **Capturing belief modifications**, not only newly-created beliefs. Requires `aelf correct` CLI + `feedback_history.kind` column, neither of which exists on main. Should be filed as a follow-up issue if/when that infrastructure ships.
- **Performance**: candidate-walk is O(n_beliefs) via `list_belief_ids() + get_belief()`. For typical session-end stores (<1k beliefs) this is sub-100ms. A focused SQL `SELECT * FROM beliefs WHERE session_id=? AND lock_level != ? AND (type=? OR origin IN (...))` is a future optimisation when stores grow.
- **Telemetry**: this hook does not currently log to the audit log alongside the UserPromptSubmit / SessionStart hooks. The information is in stderr only.

## Summary by Sourcery

Add a default-on Stop hook that surfaces session-end correction-class beliefs for locking and wire it into setup, doctor, and CLI tooling.

New Features:
- Introduce a session-end Stop hook that scans the current session for unlocked correction-class beliefs and either prompts users with pre-filled lock commands or auto-locks them based on an environment flag.

Enhancements:
- Wire the new Stop hook into the setup/unsetup workflow, settings.json hook configuration, and doctor auto-capture checks so it is installed, removable, and monitored alongside existing hooks.

Documentation:
- Document the new session-end Stop hook behavior, configuration, and defaults in the changelog.

Tests:
- Add focused tests covering Stop hook filtering, prompting and autolock behavior, plus setup/install/uninstall wiring and doctor parity for the new hook.


robotrocketscience:
[claim:review:Maxwell:2026-05-10T08:11:31Z]

robotrocketscience:
[release:review:Maxwell:2026-05-10T08:12:44Z]

Issue #602, opened 2026-05-10
fix(parallel-sessions): merge-thrash — PRs cycle attn:merge-conflict faster than rebases can complete


robotrocketscience:
## Symptom

When N parallel sessions each open small-to-medium PRs against `main`, a merge by any one session frequently re-labels every other open PR `attn:merge-conflict`. Affected authors rebase, but by the time CI re-runs (~2 min) another merge has already invalidated the rebase. PRs that would otherwise merge in 5 minutes get stuck for hours in a force-push / re-label cycle.

## Evidence

Snapshot from 2026-05-10 ~15:45 UTC. Three PRs open simultaneously, all carrying `attn:merge-conflict`:

| PR | LOC | Force-pushes today | Label flips today |
|---|---:|---:|---:|
| #591 (`feat(rebuilder): post-compact <working-state> hot-start`) | 655 | 3 (15:14, 15:34, 15:43) | 9 |
| #540 (`feat(vocab-bridge): remove vocab_bridge.py`) | 1,747 | 4 in 9 min (15:31–15:40) | 8 |
| #568 (`Update README.md`) | 18 | 0 (stuck since 06:05) | 2 |

PR #591 minute-by-minute slice (events from `gh api .../issues/591/events`):

```
15:14:50  force-push #1            -> 15:15:21 attn:merge-conflict
15:27:12  unlabeled merge-conflict -> 15:33:22 attn:merge-conflict (recurred)
15:34:12  force-push #2            -> 15:34:43 clean
15:40:02  attn:merge-conflict      -> 15:42:25 unlabeled (rebase caught up)
15:43:35  force-push #3            -> next merge will likely re-label
```

Throughput data confirms the queue itself isn't slow when PRs *do* land: of the most recent 30 merged PRs, 14 merged in 0 hours (same calendar hour) and only one took >1 hour. The slowness is concentrated in PRs that get caught in the rebase loop, not in review.

`main` advanced 138 commits since the most recent context-rebuilder check (0 merge commits — confirms FF-only policy is in force). At that velocity, any PR open more than ~10 minutes is statistically likely to be invalidated mid-rebase.

## Root cause

[...]

PR #540, opened 2026-05-10
feat(vocab-bridge): remove vocab_bridge.py module (closes #536, #433)

robotrocketscience:
## Summary

Closes #536. Closes the `vocab_bridge.py` query-rewrite module entirely. The vocabulary-gap-recovery claim that #433 originally tracked is closed by the typed-edge structural-retrieval lane (`src/aelfrice/hrr_index.py`, default-on as of v2.1 via `use_hrr_structural`). The lab campaign `exp/hrr-vocabulary-bridge` adopted that mechanism and explicitly **dropped** the cascade form that `vocab_bridge.py` implemented (R2 finding: recall@K bottleneck for the bridge use case). PR #535's bench evidence (-0.024 NDCG, 0/20 precondition rows) was correct *for that module* but not a regression on the lab's work — the module shipped without lab evidence behind it.

## Commits

Six atomic, signed:

1. `test(vocab-bridge):` drop the module's unit + integration + bench-gate test files.
2. `test(uplift):` drop `run_vocab_bridge_uplift` driver from `tests/retrieve_uplift_runner.py` + the three unit tests for it.
3. `refactor(retrieval):` drop the wiring from `retrieve_v2` — kwargs (`use_vocab_bridge`, `vocab_bridge_cache`), the `use_hrr` deprecated alias, `VOCAB_BRIDGE_FLAG`, `ENV_VOCAB_BRIDGE`, `_env_vocab_bridge_override`, `resolve_use_vocab_bridge`, the bridge build/rewrite block, and the `aelfrice.vocab_bridge` import. `hrr_index.py` docstring fixup for the dangling `VocabBridgeCache` cross-reference.
4. `feat(vocab-bridge):` delete `src/aelfrice/vocab_bridge.py`.
5. `docs(vocab-bridge):` delete `docs/feature-hrr-vocab-bridge.md`; remove `use_vocab_bridge` from `docs/CONFIG.md` (knobs list, TOML example, full section, routing-diagram mention); update `docs/ARCHITECTURE.md` to point at the surviving production spec; mark the v2.0 line item closed in `docs/ROADMAP.md`; drop the analogy from `feature-doc-linker.md` and `feature-type-aware-compression.md`.
6. `docs(changelog):` Unreleased entry under § Removed.

## Surviving HRR surfaces (untouched)

- `src/aelfrice/hrr.py` — Plate FFT bind/unbind (lab R14 algorithm).
[...]

PR #568, opened 2026-05-10
docs(readme): tighten tagline + bullet wording

robotrocketscience:
## Linked issues

 -->

## Type of change

- [ ] `feat:` — new feature
- [ ] `fix:` — bug fix
- [ ] `perf:` — performance improvement
- [ ] `refactor:` — code restructure with no behavior change
- [ ] `test:` — test-only change
- [x] `docs:` — documentation-only change
- [ ] `build:` — build system / dependency / lockfile change
- [ ] `ci:` — CI workflow / hook change
- [ ] `release:` — version bump / release tag
- [ ] `chore:` — narrow housekeeping

## Verification

- [ ] `uv run pytest tests/ -x -q` — all green
- [ ] `uv run pyright src/` — strict, no new errors
- [ ] `uv run aelf --help` — surface unchanged (or change documented)
- [ ] CHANGELOG entry added under `[Unreleased]` (if user-visible)
- [ ] Docs updated (if surface or behavior changed)

## Summary by Sourcery

Update the README to refine product messaging and clarify installation and data-ownership semantics.

Documentation:
- Tighten the introductory tagline and value proposition wording to better describe aelfrice’s behavior and guarantees.
- Clarify installation steps with an explicit installation section and an optional follow-up step for adding the first rule.
- Refine descriptions of local-only storage, auditability, reversibility, and offline operation to emphasize user control, traceability, and fully local, fast behavior.

## Summary by CodeRabbit

* **Documentation**
  * Updated README marketing copy with refined product messaging
  * Enhanced feature descriptions emphasizing local SQLite, full auditability, and absence of GPU/network requirements
  * Clarified core differentiators highlighting user control and local performance capabilities

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/568)

PR #603, opened 2026-05-10
ci(merge-thrash): pr-size-soft-cap advisory workflow (#602)

robotrocketscience:
Closes part of #602 (the conflict-probability axis).

## What lands

A new advisory GitHub Actions workflow that comments on oversize PRs without blocking them.

- `.github/workflows/pr-size-soft-cap.yml` — runs on PR `opened` / `synchronize` / `reopened` / `labeled` / `unlabeled`. Skips drafts.
- Computes `loc = additions + deletions` and `files = changed_files` from the PR event payload (no extra API calls for the metrics themselves).
- Over threshold → posts (or updates) a sticky `<!-- pr-size-soft-cap-v1 -->` comment.
- Under threshold → removes any prior sticky comment so a shrunk-back PR doesn't carry a stale warning.
- `size:override` label suppresses the comment entirely (legitimate large refactors, module removals, generated code).
- `concurrency.cancel-in-progress: true` on a per-PR group so rapid pushes don't race the comment edit.

Thresholds: 200 LOC, 3 files. Both editable in the workflow `env:` block; living constants, not vendored.

## Why advisory not enforcing

The atomic-commits norm already pushes toward small units at the commit level. The merge-thrash data in #602 shows the *PR-level* aggregate is what determines blast radius — but a hard size block would prevent things like PR #540 (1,747-line module removal) from existing at all. Soft advisory + opt-out label keeps the door open for those while creating useful pressure on the 200–500-line "would-be-fine-as-two-PRs" cohort.

## Verification

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/pr-size-soft-cap.yml'))"` → loads cleanly.
- Pinned action SHAs match other workflows in the repo (`harden-runner@8d3c67de…` v2.19.0).
- `egress-policy: audit` matches the house style.
- Permissions scoped to `pull-requests: write` + `contents: read`.
- Sticky comment idempotency: the `MARKER` constant in the workflow doubles as the comment-search key. `gh api .../comments` filtered to `github-actions[bot]` author + `body | startswith(MARKER)` finds the existing comment for in-place PATCH or DELETE.

## Out of scope (follow-ups)

- **Label-driven merge-train bot.** The other half of #602 — serializes merges to one-at-a-time so rebases can converge. Separate PR.
- Hard PR-size enforcement (intentionally not included; see "Why advisory not enforcing" above).
- Per-file LOC weighting or generated-file exclusions. Could be added if the bot's signal-to-noise turns out to be poor in practice.

## Discretion

Workflow YAML reviewed for any private terminology — clean. The CHANGELOG entry references #602 by number; nothing about session names or any private context. PR body, ditto.

## Summary by Sourcery

Add an advisory GitHub Actions workflow that comments on oversized pull requests to encourage splitting them without blocking merges.

New Features:
- Introduce a pr-size-soft-cap workflow that posts and updates a sticky advisory comment on PRs exceeding configurable LOC or changed-file thresholds, and removes it when PRs fall back under the limits.

CI:
- Add a pull_request-triggered workflow that computes PR size from event payload, respects a size:override opt-out label, and uses per-PR concurrency to avoid racing comment updates.

Documentation:
- Document the new PR-size soft-cap advisory workflow and its behavior in the Unreleased section of the changelog.

## Chores
* Added automated pull request size monitoring that posts advisory comments when changes exceed defined thresholds. Comments are automatically managed and can be bypassed using a label override.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/603)

### DIFF
--- a/.github/workflows/pr-size-soft-cap.yml
+++ b/.github/workflows/pr-size-soft-cap.yml
@@ -1,0 +1,105 @@
+name: pr-size-soft-cap
+
+# #602 — advisory soft cap on PR size to reduce merge-thrash. Comments
+# (and updates a sticky marker comment) when a PR exceeds 200 LOC or
+# 3 changed files. Authors split or apply `size:override` to opt out;
+# no hard block. Quiet when the PR is under both thresholds, including
+# when a previously-flagged PR is shrunk back below the line — the
+# sticky comment is removed in that case so the PR view stays clean.
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+
+permissions:
+  contents: read
+  pull-requests: write
+
+# One pass per PR. New pushes cancel an in-flight check so we don't
+# race on the comment edit.
+concurrency:
+  group: pr-size-soft-cap-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  size-check:
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+    # Skip drafts — they're work-in-progress and the author already knows.
+    if: github.event.pull_request.draft == false
+    steps:
+      - uses: step-security/harden-runner@8d3c67de8e2fe68ef647c8db1e6a09f647780f40 # v2.19.0
+        with:
+          egress-policy: audit
+
+      - name: Check size and post / update / remove sticky comment
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          REPO: ${{ github.repository }}
+          PR_NUMBER: ${{ github.event.pull_request.number }}
+          PR_ADDITIONS: ${{ github.event.pull_request.additions }}
+          PR_DELETIONS: ${{ github.event.pull_request.deletions }}
+          PR_CHANGED_FILES: ${{ github.event.pull_request.changed_files }}
+          # Comma-separated label names. `size:override` opts out.
+          PR_LABELS: ${{ join(github.event.pull_request.labels.*.name, ',') }}
+          LOC_LIMIT: '200'
+          FILES_LIMIT: '3'
+          MARKER: '<!-- pr-size-soft-cap-v1 -->'
+        run: |
+          set -euo pipefail
+
+          # Opt-out via label.
+          if printf '%s' ",${PR_LABELS}," | grep -q ',size:override,'; then
+            echo "size:override label present; nothing to do."
+            exit 0
+          fi
+
+          loc=$((PR_ADDITIONS + PR_DELETIONS))
+          files="${PR_CHANGED_FILES}"
+          over=0
+          if [ "$loc" -gt "$LOC_LIMIT" ]; then over=1; fi
+          if [ "$files" -gt "$FILES_LIMIT" ]; then over=1; fi
+
+          # Find any prior sticky comment authored by github-actions[bot].
+          existing_id="$(gh api \
+              "repos/${REPO}/issues/${PR_NUMBER}/comments" --paginate \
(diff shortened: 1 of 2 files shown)


robotrocketscience:
[claim:review:Einstein:2026-05-10T17:18:28Z]

robotrocketscience:
## Found one blocker, one nit

### Blocker — heredoc preserves 10-space indentation

`<<EOF` (no dash) does not strip leading whitespace from heredoc body lines. Because the `cat > /tmp/comment.md <<EOF ... EOF` block is itself indented 10 spaces under `run: |`, every line written to `/tmp/comment.md` starts with 10 spaces, including the marker.

Verified locally:

```
$ head -1 /tmp/comment.md | od -c | head -2
0000000                                            <   !   -   -   ...
                  ^^^^^^^^^^^^^^^ 10 spaces

$ jq -Rs '. | startswith("<!-- pr-size-soft-cap-v1 -->")' /tmp/comment.md
false
```

Two consequences:

1. **Sticky lookup fails.** The `existing_id` jq filter is `.body | startswith("${MARKER}")`. The body GitHub stores starts with 10 spaces, not the marker, so the filter never matches. Every push posts a fresh comment instead of updating the existing one — the opposite of "sticky".
2. **Markdown renders as a code block.** Four-or-more leading spaces on every line tells markdown "this is preformatted text", so the `### PR-size soft cap` heading, the bullet list, the bold `**N**` counts, and the inline `\`size:override\`` references all render inside a code block instead of as formatted markdown.

Fix is one line — make the heredoc tab-stripping (`<<-EOF`) and use a tab for the leading indent, OR de-indent the body and pipe through `sed 's/^          //'`, OR write the body with `printf` instead of a heredoc. The `<<-EOF` variant is the smallest diff but requires the leading whitespace to be tabs, which conflicts with most YAML linters. I'd suggest:

```bash
sed 's/^ \{10\}//' > /tmp/comment.md <<EOF
          ${MARKER}
          ### PR-size soft cap
          ...
          EOF
```

— or just `cat <<EOF | sed 's/^ \{10\}//' > /tmp/comment.md`. Either way, add a smoke check (`grep -c '^<!-- pr-size-soft-cap-v1 -->$' /tmp/comment.md` should be 1) so the regression doesn't recur.

### Nit — `pull_request: ready_for_review`

The trigger list is `opened, synchronize, reopened, labeled, unlabeled`. A PR opened as draft (skipped by the `draft == false` guard) and later marked ready-for-review won't fire this workflow until the next push or label change. Adding `ready_for_review` to the trigger list closes that gap. Not a blocker; the next sync push catches it.

### Otherwise LGTM

- harden-runner SHA-pinned ✓
- `set -euo pipefail` ✓
- minimal permissions (contents:read, pull-requests:write) ✓
- concurrency cancel-in-progress prevents the comment-edit race ✓
- `size:override` opt-out matches the issue spec ✓
- removing the sticky on shrink is a nice touch beyond spec ✓
- 2 atomic commits, both signed, FF onto main ✓
- CI green (analyze (python) pending only) ✓

Clearing `attn:review` while this is awaiting author. Re-flag after the heredoc fix.

robotrocketscience:
[release:review:Einstein:2026-05-10T17:21:30Z]

robotrocketscience:
[claim:review:godel:2026-05-10T21:44:52Z]

robotrocketscience:
Rebased onto current `main` (post-#568 merge). Both commits re-signed. Local pytest 3286 passed / 52 skipped. Diff vs main is now 2 files: `.github/workflows/pr-size-soft-cap.yml` (+105) and `CHANGELOG.md` (+4). Awaiting required-status checks before FF-merge.

robotrocketscience:
[release:review:godel:2026-05-10T22:59:47Z]

Issue #553, opened 2026-05-10
[v2.1] HRR integration: persistence default-ON + split-format migration + disk-cost opt-out


robotrocketscience:
## Summary

Track v2.1 HRR integration: persistence default-ON, split-format `.npy`+`.npz` migration, disk-cost opt-out flag.

Spec at `docs/feature-hrr-integration.md` (added by accompanying PR). Sub-issues this umbrella once the spec PR lands.

## Background

v1.7+ ships HRR machinery still default-OFF behind `use_hrr_structural`:

- `src/aelfrice/hrr.py` — Plate (1995) primitives (#216).
- `src/aelfrice/hrr_index.py` — `HRRStructIndex` for `KIND:target_id` structural-marker queries (#152).

Three blockers stand between current state and default-ON:

1. **Cold-start build cost dominates.** N=50k full-rank build is 38 s; warm-probe latency is fine (24.7 ms p95). Without persistence, every `aelf search` cold start pays the rebuild.
2. **`hrr_index.py:save()` format blocks `mmap_mode='r'`.** `np.load(.npz, mmap_mode='r')` is silently ignored. Mmap requires a separate `.npy` file. The bundled `.npz` v1.7 currently writes is incompatible with persistence-default-ON.
3. **Disk overhead per persisted store is non-trivial.** 328 MB at N=10k, 1.64 GB at N=50k. Federation × multiple stores amplifies. A disk-cost opt-out flag is required.

## Phase question

Can the HRR stack ship persistence default-ON in v2.1 without regressing latency, memory, or the FTS5 baseline NDCG?

## Acceptance (high-level — full list in spec)

- Split-format save/load (`struct.npy` + `meta.npz`); legacy bundled `.npz` accepted with deprecation log.
- `HRRStructIndexCache` reads via `mmap_mode='r'` when split format present.
- `aelf doctor` reports `hrr.persist_enabled`, `hrr.on_disk_bytes`, `hrr.last_build_seconds`.
- Cold-start `aelf search` at N=50k synthetic store ≤ 1 s when persisted; `AELFRICE_HRR_PERSIST=0` codepath verified.
- NDCG byte-equality: build→save→load→probe vs in-memory probe.
- Ephemeral-path auto-disable for `/tmp`, `/var/tmp`, `/dev/shm`, `/run`; `AELFRICE_HRR_PERSIST=1` overrides.
- Cache invalidation drops both in-memory index AND on-disk blob.
[...]

PR #617, opened 2026-05-11
feat(hrr): split-format save/load on HRRStructIndex (#553 substrate)

robotrocketscience:
First substrate piece of #553 (HRR persistence default-ON).

## What this PR ships

Migrates `HRRStructIndex` persistence from a single bundled `.npz` to
the per-store split-format directory layout the spec at
`docs/feature-hrr-integration.md` requires:

```
<persist_dir>/
  struct.npy        # (N, dim) float64; mmap-able
  meta.npz          # belief_ids, role/id vectors, dim, seed, version
```

`save(path)` writes both files atomically (temp + `os.replace`).
`load(path)` auto-detects layout — directory with both files → split
format; bare file path → legacy bundled `.npz` reader with a one-shot
module-logger deprecation warning. `load(..., mmap=True)` opts the
struct matrix into `np.load(mmap_mode='r')` (default off keeps callers
byte-stable). `_NPZ_VERSION` bumps from 1 → 2 with the legacy alias
preserved.

## Acceptance bullets covered (from `docs/feature-hrr-integration.md`)

- [x] **Bullet 1.** `HRRStructIndex.save()` writes split format.
- [x] **Bullet 2.** `HRRStructIndex.load(path)` accepts the split format AND the legacy bundled `.npz` (with deprecation log).
- [x] **Bullet 6.** NDCG byte-equality: build → save → load → probe vs in-memory probe produces bit-identical scores.

## What this PR does NOT ship (separate atomic PRs)

- Bullet 3. `HRRStructIndexCache` reads via `mmap_mode='r'` when split format is present.
- Bullet 4. `aelf doctor` rows: `hrr.persist_enabled`, `hrr.on_disk_bytes`, `hrr.last_build_seconds`.
- Bullet 5. Cold-start `aelf search` ≤ 1 s at N=50k bench (perf-gated).
- Bullet 7. `[retrieval] hrr_persist = false` TOML / env-var resolver.
- Bullet 8. Ephemeral-path auto-disable.
- Bullet 9. Cache invalidation unlinks on-disk blob.
- Bullet 10. `docs/CONFIG.md` + `docs/COMMANDS.md` updates.

These are intentionally split per the spec's "Coordination" section ("sub-issues the implementation as separate PRs"). The substrate landing here unblocks all of them.

## Verification

- `uv run pytest tests/test_hrr_struct_index.py` → 25 passed, 2 skipped (perf-gated).
- Full suite: `uv run pytest -x -q` → 3303 passed, 52 skipped, 0 failures.
- Manual: hand-rolled v1.7-shape `.npz` loaded, deprecation warning emitted exactly once per process, struct matrix bytes-equal to the in-memory original, probe scores bytes-equal across the boundary (covered by `test_load_legacy_bundled_npz_with_deprecation_log`).

## LoC budget vs spec H₂

Spec H₂ predicts ≤ 100 LoC change in `hrr_index.py`; this PR is +116 / -26 (net +90). Within budget. Test diff is +76 / -1 — the spec's "≤ 5 LoC" bound applied to *changes to the existing test* (1 line: `path = tmp_path / "hrr"`); new tests for new acceptance bullets (byte-equality, legacy fallback, missing-path) are net additions.

## Summary by Sourcery

Switch HRRStructIndex persistence to a split directory layout while retaining backward-compatible loading of legacy bundled .npz files.

New Features:
- Persist HRRStructIndex to a per-store directory containing struct.npy for the struct matrix and meta.npz for metadata, with optional mmap loading of the struct matrix.

Enhancements:
- Make HRRStructIndex.save/load atomic and auto-detect between split-directory and legacy bundled .npz layouts, with a one-shot deprecation warning for legacy files.

Documentation:
- Document the new HRRStructIndex split-format persistence and mmap-able struct matrix in the changelog.

Tests:
- Add tests to verify split-format save/load, byte-identical structs and probe scores across persistence, legacy .npz loading with a deprecation log, and error handling for missing persistence paths.


robotrocketscience:
[claim:review:Maxwell:2026-05-11T00:23:03Z]

robotrocketscience:
Reviewed by Maxwell. Three atomic signed commits, FF on main, full suite green (3303 passed locally per PR body), discretion clean. Approving for `ready-to-merge`.

## Observations (non-blocking)

1. **Atomic-pair guarantee.** `os.replace(struct_tmp, struct_final)` runs before `os.replace(meta_tmp, meta_final)`. A concurrent reader (or a save interrupted between the two replaces) can observe `struct.npy` from generation N+1 paired with `meta.npz` from generation N. With aelfrice's single-process write pattern this is acceptable, but if multi-process write ever becomes possible, the read-side belief_ids→struct row mapping would silently desync. Worth a comment in `save()` or a follow-up that orders the replaces meta-last and bumps `_LAYOUT_VERSION` on every save.

2. **`mmap=True` parameter has no test.** Bullet 3 (cache reads via `mmap_mode='r'`) is correctly out of scope here, but the kwarg itself is shipped on `load()` and goes untouched by the test suite. A one-line `load(path, mmap=True).struct.flags['OWNDATA'] is False` would pin the behaviour without depending on bullet 3.

3. **One-shot deprecation flag.** `test_load_legacy_bundled_npz_with_deprecation_log` resets `_legacy_deprecation_logged = False` before exercising the log path, which is the right hygiene, but the "one-shot" property itself (a second legacy load in the same process is silent) isn't covered. Reasonable to defer.

None block merge. Adding `ready-to-merge`.

robotrocketscience:
[release:review:Maxwell:2026-05-11T00:24:45Z]

robotrocketscience:
[claim:review:noether:2026-05-11T00:28:43Z]

robotrocketscience:
[release:review:noether:2026-05-11T00:29:38Z]

Issue #594, opened 2026-05-10
feat(doctor|cli): prune dormant per-project DBs


robotrocketscience:
## Background

#589 audit found 9 per-project DBs idle 12–16 days, totaling ~93k beliefs (see #589 body for the full table). They hold beliefs from projects worked briefly with older aelfrice versions, then abandoned.

Per #589, doctor will start nagging that they need migration. Independent of migration, some of them may not be worth keeping at all — abandoned projects, or projects whose belief content has zero forward value. Operator direction (#589 decision-resolution): prune them.

## Proposed scope

Add a tool surface to identify and (optionally) delete dormant per-project DBs. Two open questions:

1. **Detection criteria:**
   - Idle threshold only (≥30d? ≥90d? configurable?)
   - Filesystem orphan check (project root no longer exists at the recorded path)
   - Both — flag if either condition holds

2. **Actuation surface:**
   - `aelf doctor --prune-dormant` (lives next to the schema nag)
   - `aelf prune` new subcommand
   - Folded into the auto-fix surface from the related auto-migrate follow-up

Cleanup is destructive (`rm` of the DB file). Confirmation must be deliberate — `y/N` per DB at minimum, never silent delete.

## Acceptance criteria

- [ ] Detection rule chosen and documented
- [ ] Actuation surface chosen
- [ ] `y/N` confirmation per DB before deletion
- [ ] Tests: DB matching detection criteria → flagged; non-matching → not flagged; `--no` answer → DB preserved
- [ ] `docs/INSTALL.md` documented alongside the #589 doctor block

## Out of scope

- Schema-nag itself ships in #589.
- Auto-migrate of non-dormant legacy DBs ships in the auto-migrate follow-up.

## Related

- #589 (schema-nag, predecessor)

robotrocketscience:
## Triage (noether) — recommendations on the explicit Qs

Not claiming — surfacing for the operator decision this issue is gated on. See parallel triage on #593 for the actuation-pattern coupling.

### Q1 — detection criteria

**Recommend both, reported in two separate sections.**

| Class | Criterion | Risk profile |
|---|---|---|
| Dormant by inactivity | `last_write` age ≥ threshold (default 30d, configurable) | "You might come back to this." Lower-confidence delete. |
| Dormant by orphan | Project root recorded in DB metadata no longer exists on filesystem | High-confidence delete — the project is gone. |

[...]

Issue #593, opened 2026-05-10
feat(doctor|cli): auto-migrate (or prompt-to-migrate) on detected legacy schema


robotrocketscience:
## Background

#589 adds an `aelf doctor` nag when per-project DBs are on pre-v1.x schema (no `origin` column). The fix line directs the operator to run `aelf migrate --legacy <path> --target <path> --apply` per DB. Operator note while resolving #589's `attn:decisions-needed` gate:

> ideally the user shouldn't have to be running doctor fix stuff — it should happen automatically.

This issue captures that direction. #589 stays nag-only by convention (doctor is read-only). This issue covers the actuation surface that automates the fix.

## Proposed scope

Two design questions to settle before implementation:

1. **Where does the auto-fix actuation live?**
   - `aelf doctor --fix` flag (default stays read-only; `--fix` mutates)
   - `aelf fix` new subcommand that consumes a `DoctorReport` and applies migrations
   - Auto-fire on `aelf doctor` when an opt-in setting is enabled (e.g. `aelfrice.autofix=migrate`)

2. **What's the user-confirmation model?**
   - Always prompt `y/N` per DB — safe, but defeats the "happen automatically" intent
   - Prompt once, then apply to all detected DBs in batch
   - Silent auto-migrate on session start — highest leverage, highest blast radius

The migration itself is non-destructive: `migrate.py:110` copies legacy beliefs into a target path with `ORIGIN_UNKNOWN` fallback for missing `origin`. The legacy DB is untouched. Even silent auto-migrate is recoverable — the modern-schema DB is created alongside the legacy one.

## Acceptance criteria

- [ ] Decision recorded on (1) actuation surface and (2) confirmation model
- [ ] Implementation matches the decision
- [ ] Tests: no-op when no legacy DBs detected; correct migration when one detected; confirmation path when applicable
- [ ] `docs/INSTALL.md` updated alongside the #589 doctor docs

## Out of scope

- The detect-and-nag side ships in #589.
- Cleanup of dormant DBs is its own follow-up.

## Related

- #589 (nag-and-fix-line, this issue's predecessor)
- #557 (auto-capture-hook nag pattern this convention started from)

robotrocketscience:
## Triage (noether) — recommendations on the explicit Qs + one implicit Q

Read of the body + `src/aelfrice/migrate.py` + `doctor.py:_check_legacy_schema_dbs`. Not claiming — surfacing for the operator decision this issue is gated on.

[...]

Issue #624, opened 2026-05-11
ci: switch from FF-only merge-train to GitHub native merge queue


robotrocketscience:
## Problem

The current `merge-train.yml` workflow (FF-only, bot-pushes-to-main) bounces PRs whenever main moves between author rebase and the workflow's slot.

**Observed today (2026-05-11):**
- 04:34:05Z — PR #615 (docs/multimodel scope memo) merged via merge-train, advanced main `b265753 → 8b8c6d4`.
- 04:35:32Z — PR #601 (eval-harness host-agent JSONL split) hit `[2/5] FF check`, found its base `b265753` no longer reachable from main, bounced and unlabeled.

Each bounce requires the author to manually rebase, force-push, and re-label. With K parallel sessions labeling PRs in a window, exactly one merges per round and K-1 bounce. Single docs-only PRs have been observed sitting unmerged for 24+ hours.

## Root cause

The workflow design assumes the author has pre-rebased the branch onto current main before labeling. In a parallel-session workflow, that assumption is structurally violated: sister sessions advance main between rebase and slot. The workflow correctly refuses the non-FF push (per the no-bot-signing-key constraint), but the failure mode is "every PR after the first bounces."

## Fix

Switch to **GitHub native merge queue** (free for public repos).

GitHub serializes queued PRs, **rebases each onto current main + pending queue entries server-side**, runs the required status checks on the rebased result via `gh-readonly-queue/main/pr-N-<sha>` branches, and only merges entries that pass. No client-side rebase needed; no bounces from drift.

### Trade-off accepted

Per-commit signatures on main change from the author's SSH key to GitHub's web-merge key. The author field is preserved. Original SSH-signed commits persist on `refs/pull/N/head` (permanent). The branch protection `required_signatures` rule accepts GitHub's signature as verified.

### What changes

**In-repo (this PR):**
1. `.github/workflows/ci.yml` — add `merge_group` trigger so `pytest (3.12/3.13)` fire on queued branches.
2. `.github/workflows/staging-gate.yml` — add `merge_group` trigger for `secrets-scan` / `pattern-scan` / `history-scan` (the required checks); conditional-out PR-context-only jobs (`pr-title-prefix`, `pr-body-issue-link`, `commit-msg-prefix`, `release-docs-check`) on `merge_group` events; parameterize `history-scan` to read base/head from either event context.
3. `CHANGELOG.md` — document the new merge path.

**Ruleset (manual step, admin only):**
[...]

PR #612, opened 2026-05-10
feat(sentiment-hook): production wire-up of sentiment-feedback into UPS lane (closes #606)

robotrocketscience:
Closes #606.

## What this does

Wires the v2.0 `sentiment_feedback` module into a live `UserPromptSubmit` hook lane. Before each turn's retrieval, the lane reads the most-recent prior UPS audit row for the same `session_id`, projects its `beliefs[*].id`, and — if `detect_sentiment` matches the prompt — applies the signal (positive / negative, base / strong, escalated) to those beliefs via `apply_sentiment_to_pending`. Bumped posteriors are reflected in this turn's retrieval, so a "no, that's wrong" prompt demotes the offending belief before the next block is built.

## How the decisions came out

Captured in `docs/v3_sentiment_feedback_hook.md`. Summary:

- **Hook lane**: `UserPromptSubmit`. The corrective prompt arrives after the assistant has acted on the prior retrieval block; applying it at the next UPS keeps the bump in lockstep with the retrieval window it's correcting.
- **Window**: the most-recent prior UPS audit row for the same `session_id`. Single-session for v3.0; cross-session propagation is explicit follow-up.
- **Audit surface**: new tag `sentiment_feedback` written to the existing `hook_audit.jsonl`, carrying `pattern`, `matched_text`, `valence`, `belief_ids`. The per-belief `feedback_history` rows continue to carry `source = sentiment_inferred` (unchanged from v2.0).
- **Opt-in**: single flag `[feedback] sentiment_from_prose = true` in `.aelfrice.toml` (or the existing env var). Default off — no behavior change for users who haven't opted in.
- **Privacy**: no new prose-data surface. UPS already reads every prompt; the lane does additional *processing* on data already in the hook's hands. Audit row stores the matched pattern + a bounded substring, never the full prompt.

## Commits

1. `docs(sentiment-hook): v3.0 spec for #606 production wire-up` — decision memo.
2. `feat(sentiment-hook): wire detect_sentiment into UPS lane` — `apply_sentiment_feedback` + two helpers (`_load_aelfrice_toml`, `_load_prior_ups_belief_ids`) + new audit tag + wiring into `user_prompt_submit` before retrieval.
3. `test(sentiment-hook): unit + two-session bench fixture` — 16 tests covering the disabled gate, no-signal short-circuit, no-prior-UPS short-circuit, posterior demotion, audit-row schema, missing-belief skip, UPS-integration paths, and the AC4 two-session ranking fixture.
[...]

PR #627, opened 2026-05-11
feat(setup): self-installing hook manifest at aelf-CLI first run (closes #623)

### LINKED ISSUE #623: feat(setup): self-installing hook manifest at aelf-CLI first run
## Problem

Aelfrice's auto-capture pipeline (transcript-ingest, commit-ingest, session-start, stop-hook — all default-on since v2.1 / #529 / #582) requires an explicit `aelf setup` invocation to land in `~/.claude/settings.json`. Three things follow from that:

1. **Fresh install requires two steps.** `pipx install aelfrice` puts the binary on PATH but does nothing to Claude Code's settings. The user has to know to run `aelf setup` afterward. README #136 markets "passive capture" as the headline feature but the user only gets it after a second deliberate command.

2. **Bare upgrade (`pipx upgrade aelfrice` / `uv tool upgrade aelfrice` / `pip install -U aelfrice`) skips setup.** Any new default-on hook added in vN+1 (e.g. `aelf-stop-hook` shipped v2.1) is missing for users who don't re-run `aelf setup`. The `/aelf:upgrade` slash command re-runs setup in step 3 — but users who upgrade via package manager directly bypass it.

3. **The current catch-net is `aelf doctor` (#557).** It nags about missing default-on hooks with `fix: re-run 'aelf setup'`. But this only fires when the user runs `aelf doctor` — passive users never see it.

User feedback on Discord (Clyvle, 2026-05-10): *"why isn't this stuff getting automatically updated on releases"*.

## Proposed surface

Self-installing manifest. Every `aelf <cmd>` invocation idempotently merges the package-declared hook manifest into `~/.claude/settings.json`, gated on a version stamp.

### Manifest declaration

A new file `src/aelfrice/data/hook_manifest.json` ships in the wheel, listing every default-on hook with its event lane, command basename, and arg pattern. Same shape as what `aelf setup` writes today, but declared once and consumed by both `aelf setup` and the first-run path.

### Version-stamped dotfile

`~/.aelfrice/installed-manifest-version` (or similar) records the package version that last wrote hook entries. On every `aelf <cmd>` cold-path execution:

1. Read installed package version (`importlib.metadata.version("aelfrice")`).
2. Read stamped version (default `"0.0.0"` if absent).
3. If `installed > stamped`: read the package's hook manifest, merge into `~/.claude/settings.json` (preserving user-added entries and unrelated config), write new stamp. Log a single stderr line: `aelfrice: hooks updated to vX.Y.Z (was vA.B.C). re-run 'aelf setup --no-...' to opt out of any.`
4. If `installed == stamped`: skip (~zero overhead).

### Settings.json merge rules

- **Add** a hook entry if absent (key by command basename, e.g. `aelf-transcript-logger`).
- **Update** a hook entry if the manifest's arg pattern changed (rare, but covers e.g. flag additions).
- **Never remove** entries the user added themselves (i.e. anything not in the manifest's owned-basenames set).
- **Honor opt-out**: a separate `~/.aelfrice/opt-out-hooks.json` (set by `aelf setup --no-X`) prevents that hook from being added back. Closes the "I disabled this, please don't re-enable on upgrade" loop.

### What `aelf setup` becomes

Stays as the explicit-opt-in entry point with the per-hook flags (`--no-transcript-ingest`, etc.). Writes to the same opt-out file. The first-run merge respects it. `aelf setup` itself stops being mandatory for the happy path.

## Why this isn't "pip post-install"

`pip` (and `pipx`, `uv tool`) deliberately do not run user-supplied scripts on install — the npm `postinstall` pattern caused enough security incidents that the Python ecosystem rejected it. Modifying `~/.claude/settings.json` silently at `pip install` time would be a hostile pattern. Doing it at first `aelf <cmd>` invocation is the user-consent boundary: the user just ran an `aelf` command, they want it to work.

## Comparison: this is the MCP-server pattern

MCP servers don't have a "re-setup on upgrade" problem because they're registered once via `claude mcp add` and the host re-spawns the subprocess on every session (new binary version → new tools auto-discovered via `tools/list`). Aelfrice can't use that mechanism directly because hooks fire on Claude Code lifecycle events that MCP doesn't expose — but it can borrow the property by collapsing settings-management to a single declarative manifest shipped in the package, with the binary version being the source of truth for what's wired.

## Acceptance

- [ ] `src/aelfrice/data/hook_manifest.json` declares all default-on hooks (4 today: transcript-logger, commit-ingest, session-start, stop-hook).
- [ ] First `aelf <cmd>` after a fresh install creates the dotfile, merges the manifest into `~/.claude/settings.json`, and prints a one-line stderr message naming what was added.
- [ ] First `aelf <cmd>` after a bare `pipx upgrade` (where the binary version > stamped version) updates the settings.json delta and prints `hooks updated to vX.Y.Z (was vA.B.C)`.
- [ ] User-set opt-outs (`aelf setup --no-transcript-ingest`) persist across upgrades — the disabled hook is not re-added.
- [ ] User-added hook entries unrelated to aelfrice are preserved byte-identical (settings.json merge is additive within the `hooks.*` block; we own only our basenames).
- [ ] `aelf setup` still works as today (manual opt-in path remains).
- [ ] Cold-path overhead on `aelf <cmd>` is bounded by a single mtime check on the dotfile (no-op if version unchanged).
- [ ] `aelf doctor` continues to detect drift (manual edits to settings.json that diverge from manifest).
- [ ] A `--no-auto-install` env var (`AELFRICE_NO_AUTO_INSTALL=1`) opts out of the first-run merge entirely for power users who manage settings.json by hand.

## Out of scope

- Migrating non-default hooks (rebuilder, search-tool) — they remain opt-in via `aelf setup --rebuilder --search-tool`.
- Per-project DB schema migration (separate decision: #593, #594).
- Slash-command bundle (`/aelf:*` files in `~/.claude/commands/aelf/`) — that's already idempotent via `aelf setup` and the first-run path can extend to it later.

## Risks

- **settings.json corruption.** A bad merge could nuke user config. Mitigation: atomic write via tmp-file + rename, schema-validate the parsed result before writing, keep a `.aelfrice-backup-<version>` of the pre-merge settings.
- **First-run latency.** A `~/.claude/settings.json` read+merge+write on every cold `aelf <cmd>` would add ~10ms. Gate behind the mtime+version check so it's a single stat call after the first merge of a given version.
- **Concurrent invocations.** Two `aelf` processes starting simultaneously could race on the merge. Use a `flock` on `~/.aelfrice/installed-manifest-version` for the merge critical section.
- **The setup-can-fail caveat (Clyvle's machine, SyntaxWarnings).** If the auto-merge fails, the existing `aelf setup` recovery path must still work — i.e. the merge writes a stamp **after** successful settings.json write, so a failed merge leaves the previous-version stamp intact and gets retried next invocation.

## Related

- #529 — default-on hooks shipped v2.1.
- #582 — stop-hook addition (a recent example of a new default-on hook that bare-upgrade users miss).
- #557 — `aelf doctor` nag for missing default-on hooks (current catch-net; remains useful for drift detection).
- #611 / PR #614 — `/aelf:upgrade` slash command imperatives (the slash-command path already does the right thing; this issue covers the bare-package-manager-upgrade path that bypasses it).
- README §136 "Passive capture" — the marketing copy this issue brings the product up to.

robotrocketscience:
## Summary

Closes #623. Bare `pipx upgrade aelfrice` (and `uv tool upgrade aelfrice` / `pip install -U aelfrice`) no longer leaves new default-on hooks unwired in `~/.claude/settings.json`. The first `aelf <cmd>` invocation after the installed package version exceeds the on-disk stamp merges the manifest delta — happy path is one stat + one short file read.

## What landed

* `src/aelfrice/data/hook_manifest.json` — declarative source-of-truth for the 5 default-on hooks: `aelf-hook`, `aelf-transcript-logger`, `aelf-commit-ingest`, `aelf-session-start-hook`, `aelf-stop-hook`. Schema is versioned (`schema_version=1`) so future entries are forward-compatible.
* `src/aelfrice/auto_install.py` — version-stamped merger.
  * Stamp at `~/.aelfrice/installed-manifest-version`. Happy path short-circuits before any settings.json read.
  * `fcntl.LOCK_EX` on a sibling lock file so concurrent `aelf` processes cannot race.
  * Re-checks the stamp inside the critical section to absorb the "another process completed the merge while I was waiting" case.
  * Re-uses the existing `aelfrice.setup` install primitives; on-disk shape is byte-identical to what `aelf setup` writes.
  * Stamp is written **only** after `install_*` calls succeed → failure leaves the stamp untouched and the next `aelf <cmd>` retries.
  * Opt-out file at `~/.aelfrice/opt-out-hooks.json` — if a user ran `aelf setup --no-transcript-ingest` ever, that choice survives upgrades.
  * `AELFRICE_NO_AUTO_INSTALL=1` hard bypass.
* `cli.main()` calls `auto_install_at_cli_entry()` after argparse but before the subcommand func. Skipped for `setup` / `unsetup` / `uninstall` / `doctor` (they own settings.json themselves; doctor must see on-disk state).
* `aelf setup` reflects `--no-X` flag choices into the opt-out file and bumps the stamp on success — keeps the explicit-opt-in path and auto-install path coherent.
* INSTALL.md + CHANGELOG.

## Acceptance bullets (from #623)

- [x] `src/aelfrice/data/hook_manifest.json` declares all default-on hooks (5: includes the always-on `aelf-hook` retrieval entry alongside the 4 listed in the issue body).
- [x] First `aelf <cmd>` after a fresh install creates the dotfile, merges the manifest into `~/.claude/settings.json`, and prints a one-line stderr message naming what was added.
- [x] First `aelf <cmd>` after a bare `pipx upgrade` (binary version > stamped version) updates the settings.json delta and prints `hooks updated to vX.Y.Z (was vA.B.C)`.
- [x] User-set opt-outs (`aelf setup --no-transcript-ingest`) persist across upgrades — the disabled hook is not re-added.
- [x] User-added unrelated entries in settings.json are preserved byte-identical.
- [x] `aelf setup` still works as today; now also bumps the stamp.
- [x] Cold-path overhead is bounded by a single mtime/file read after the first merge.
- [x] `aelf doctor` continues to detect drift (it is in `_AUTO_INSTALL_SKIP_CMDS`, so it sees on-disk state, not what auto-install would write).
- [x] `AELFRICE_NO_AUTO_INSTALL=1` opts out entirely.

## Tests

27 new tests across three files:

* `tests/test_auto_install.py` (19) — manifest loader, stamp file, opt-out file, first-run, fast-no-op, upgrade-delta, opt-out persistence, user-entry preservation, env-var bypass, failure semantics (stamp untouched).
* `tests/test_cli_auto_install.py` (4) — main() wiring + skip-cmd set + defensive exception swallow.
* `tests/test_cli_setup_opt_out_sync.py` (4) — `aelf setup` writes stamp, `--no-X` → opt-out, opt-out rescinded on bare setup, manifest-name vs argparse-dest distinction.

Full suite: 3330 passed / 52 skipped — no regressions.

## Test plan

- [x] Fresh-install simulation: blank `~/.aelfrice/`, run `aelf status`, expect all 5 hooks merged + stderr line "aelfrice: installed default hooks for v…"
- [x] Bare-upgrade simulation: stamp at v2.1.0, run `aelf status` at v2.2.0 with one hook entry surgically removed → only that hook re-added, stderr line "hooks updated to v2.2.0 (was v2.1.0) — added: stop_lock_prompt"
- [x] Opt-out persistence: `aelf setup --no-transcript-ingest`, then re-run with stamp gap, hook is NOT re-added
- [x] User-added settings preserved byte-identical
- [x] AELFRICE_NO_AUTO_INSTALL=1 → no stamp written, no stderr
- [x] Install function raising → stamp stays at prior value (next invocation retries)
- [x] `aelf doctor` does NOT trigger auto-install (so drift remains detectable)

## Out of scope

* Per-project DB schema auto-migration → #593
* Dormant DB pruning → #594
* Slash-command bundle auto-install (already idempotent via `aelf setup`; future extension)
* Updating an existing manifest entry's arg pattern (no such case today; loop runs the install_fn either way which is idempotent on the basename, so behavior would be correct — but no test exercises it)

## Summary by Sourcery

Introduce a bundled, self-installing hook manifest that automatically wires default-on hooks after installs and upgrades, ensuring settings.json stays in sync without rerunning `aelf setup`, while preserving user opt-outs and minimizing CLI startup overhead.

New Features:
- Add a versioned hook manifest JSON describing default-on hooks shipped with the CLI.
- Run an automatic manifest-driven hook installer on first CLI invocation after install or upgrade, gated by a version stamp and environment opt-out.
- Persist and honor per-hook opt-outs across upgrades via a dedicated opt-out file and CLI `--no-*` flags.

Enhancements:
- Integrate auto-install behavior into `aelf setup` so it synchronizes opt-outs and advances the manifest version stamp.
- Skip auto-install for commands that manage settings.json directly to preserve their semantics.
- Document the new self-installing hook manifest behavior and upgrade flow in INSTALL.md and CHANGELOG.

Tests:
- Add comprehensive tests for the auto-install module, including manifest loading, stamp/opt-out handling, upgrade behavior, and failure semantics.
- Add CLI-level tests verifying auto-install wiring, skip-command behavior, and exception handling.
- Add tests ensuring `aelf setup` correctly syncs opt-outs and version stamps with the auto-install system.

## Summary by CodeRabbit

* **New Features**
  * Added automatic hook installation that deploys default hooks on first run after install or upgrade, with built-in opt-out controls via environment variables and setup flags.

* **Documentation**
  * Added documentation for the auto-installing hook manifest system, covering version reconciliation, opt-out persistence, and configuration management.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/627)


robotrocketscience:
[claim:review:leibniz:2026-05-11T05:02:09Z]

robotrocketscience:
Reviewed by leibniz. **Do not approve as-is** — one HIGH-severity CodeQL finding I want fixed before merge; everything else is non-blocking.

## Blocker

**HIGH-severity CodeQL alert 349 — `py/overly-permissive-file` at `src/aelfrice/auto_install.py:368`.**

```python
lock_fd = os.open(str(lock_path), os.O_CREAT | os.O_RDWR, 0o644)
```

The lock file ends up world-readable. Practical risk on a single-user dev machine is low (the file holds nothing — it's a flock target with empty content), but the principle is right and the fix is one character: `0o600`. The umask on most distros would already mask this down for files-in-`$HOME`, but explicit `0o600` removes the dependency on host umask and the dependency on whether `$HOME` happens to live on a permissive filesystem. Please change.

## Non-blocking

1. **CodeQL alerts 352 + 353 — `py/empty-except` at lines 258 and 397.** Both are legitimate best-effort cleanup (TOCTOU between `exists()` and `unlink()`; `flock` unlock during teardown). The fix is to add a one-line `# noqa: BLE001` *plus* an inline comment naming the race the bare-pass guards. Right now CodeQL has no signal that you considered the case; an explanatory line silences it and documents the design intent.

2. **CodeQL alerts 350 + 351 — unused imports in `tests/test_auto_install.py`** (`os` line 19, `AutoInstallResult` line 25). Trivial cleanup, ship in the same fixup commit.

3. **Docstring nit (top of file, line ~24).** "Acquires an exclusive `flock` on the stamp file during the merge" — actually you `flock` a sibling lock file `.auto-install.lock`, not the stamp itself. The current implementation is correct (locking the stamp would fight the atomic-replace write); the docstring just hasn't caught up.

4. **Race-failure reasoning is invisible.** `maybe_install_manifest` lines 372-378: when `LOCK_NB` returns `BlockingIOError`, we return `ran=False` without re-checking the stamp. The implicit assumption is that the lock-holding process WILL successfully bump the stamp; if it crashes mid-merge, our caller silently moves on without merging this invocation either. That's fine — the next `aelf <cmd>` will see the stale stamp and retry — but the reasoning isn't documented in code. A two-line comment ("loser is no-op; if winner crashes, next invocation retries because stamp wasn't bumped") would help the next person who reads this.

## Substantive — what works

The split between `_dispatch` keys and the bundled JSON manifest is the right shape: adding a sixth default-on hook in v2.2 is one row in `data/hook_manifest.json` plus one row in `_DISPATCH`, no other code touches. The `_result_added_anything` polymorphism over `InstallResult` (bool) vs `TranscriptIngestInstallResult` (tuple) cleanly hides the per-installer return-shape difference behind one boolean, so the merger doesn't grow a per-installer special case.

Stamp-after-success is the right invariant. The atomic-write helper (`_atomic_write_json`) does tempfile → fsync → `os.replace`, which means a SIGKILL between any two of those leaves either the prior file intact or the new file fully on disk — never a partial state. Same shape as `write_stamp`. ✓

The `_AUTO_INSTALL_SKIP_CMDS` list correctly carves out the four commands that need to see drift (`doctor`) or own the settings.json themselves (`setup`/`unsetup`/`uninstall`). I verified `doctor` is in the skip set so the legacy-schema and missing-auto-capture nags continue to fire on actual on-disk state, not on what auto-install would write.

The opt-out persistence story is sound: `add_opt_out` is called from `aelf setup --no-X` (per the spec), `read_opt_outs` is consulted on every merge, so a user who explicitly disabled a hook never gets it re-added by an upgrade. The opt-out file shape (`{"opt_out": [...]}`) is forward-compatible — a future entry could be added without breaking the parser.

Failure semantics match the docstring claim: install_fn raising → never reaches `write_stamp` → next invocation retries. The pre-flight `read_stamp` short-circuit gives O(1) happy-path overhead, which is the load-bearing performance property.

Conditional approval: drop the `0o600` fix into a follow-up commit, address the four non-blocking notes if you want a clean security-tab, then I'll re-review and add `ready-to-merge` if the substantive review still holds.

(Heads-up unrelated to this PR: base is `07d046328` but `github/main` has moved to `00a3b9cf` since #612 + #624 landed — you'll need a rebase for the merge-train regardless.)

[release:review:leibniz:2026-05-11T05:08:00Z]

robotrocketscience:
[release:review:leibniz:2026-05-11T05:05:44Z]

robotrocketscience:
[claim:review:noether:2026-05-11T05:11:05Z]

robotrocketscience:
Request changes — CodeQL flagged 5 new alerts on this PR (the standalone CodeQL check_run is FAILURE; the workflow-job ones passed because they don't gate on alert-count). Fixes are all trivial, but the high-severity one is a real principle-of-least-privilege miss worth correcting before merge:

**High (1) — `py/overly-permissive-file` at `src/aelfrice/auto_install.py:368`**

```python
lock_fd = os.open(str(lock_path), os.O_CREAT | os.O_RDWR, 0o644)
```

The lock file contains no sensitive data (just an flock token), so the practical exposure is nil — but everything else under `~/.aelfrice/` is single-user and a stray world-readable artefact in there is a smell. Change the mode to `0o600` and CodeQL clears.

**Notes (2) — `py/empty-except` at `src/aelfrice/auto_install.py:258` and `:397`**

Both are intentional best-effort cleanup paths (unlink-on-empty and lock-release). CodeQL just wants a one-line `# best-effort: …` comment in each `except` body to suppress the rule. No code change needed.

**Notes (2) — `py/unused-import` in `tests/test_auto_install.py`**

`os` (line 19) and `AutoInstallResult` (line 25) are imported but not referenced. Drop them.

---

Everything else looks good — manifest schema is versioned, `fcntl.LOCK_EX | LOCK_NB` with the post-acquire stamp re-check correctly handles the "another process completed the merge while I was waiting" case, opt-out persistence is cleanly reflected from `aelf setup --no-X`, the `_AUTO_INSTALL_SKIP_CMDS` carve-out for `setup`/`unsetup`/`uninstall`/`doctor` keeps drift-detection working, stamp-write-only-on-success keeps the retry semantics correct, and the 27-test surface covers all the documented invariants. CI is otherwise green (pytest 3.12 + 3.13, both staging-gate runs, deptry, vulture, typos, calibration, CodeRabbit, Sourcery).

Releasing the review claim. Re-request review once the four CodeQL findings are addressed and a follow-up commit is pushed.

robotrocketscience:
[release:review:noether:2026-05-11T05:12:41Z]

robotrocketscience:
CodeQL follow-up pushed — all four findings addressed:

- **High** `py/overly-permissive-file` at `auto_install.py:368` — `0o644` → `0o600` (commit `fix(auto_install): tighten lock-file mode`).
- **Notes ×2** `py/empty-except` at `auto_install.py:258` and `:397` — added `# best-effort:` comments documenting intent (commit `style(auto_install): annotate intentional empty-except blocks`).
- **Notes ×2** `py/unused-import` at `tests/test_auto_install.py:19` and `:25` — dropped `os` and `AutoInstallResult` (commit `test(auto_install): drop unused os + AutoInstallResult imports`).

Rebased on `github/main` (was 9 commits behind; clean ff-rebase, no conflicts). Local `pytest` green — 3361 passed, 52 skipped. Stale labels cleared (`attn:unblock`, `attn:merge-conflict` were leftover); re-flagged `attn:review`.

Re-requesting CodeQL — three new commits on top of the original five.

robotrocketscience:
[claim:review:einstein:2026-05-11T05:24:18Z]

robotrocketscience:
## Review pass 3 — einstein

All five reviewer-requested fixes from prior reviews applied cleanly. Inspecting the fixup commits:

- **`ca59cc4` — HIGH `py/overly-permissive-file`.** `0o644 → 0o600` at `src/aelfrice/auto_install.py:368`. Correct.
- **`67d0821` — `py/empty-except` ×2.** Inline intent comments at `:258` (opt-out unlink — non-critical cleanup) and `:397` (flock release in finally — `close()` releases either way). Both annotations name the race they guard, which is what CodeQL wants.
- **`45b4893` — unused imports.** `os` (line 19) and `AutoInstallResult` (line 25) dropped from `tests/test_auto_install.py`. Clean.

Plus the bonus passes:
- **`9642c73`** — host-product proper-noun + model-id scrub on the new surface.
- **`be671e7`** — `INSTALL.md` section documenting the auto-install behavior.

### CI

Pytest 3.12 + 3.13 green. Standalone CodeQL check is now `NEUTRAL` (was `FAILURE` pre-fix), which is the expected post-fix state — the previously-flagged alerts are gone. Staging-gate, deptry, vulture, typos, calibration, label, size-check, CodeRabbit all green. `analyze (python)` is `IN_PROGRESS` on the latest push (`45b4893`) — confirm it lands green before flipping `ready-to-merge`.

### Rebase needed

Branch base is `3dc6df1`; `github/main` is now at `cc26512` (several commits ahead via #624 / #612 / #631). Merge-train will reject the FF push until you rebase:

```
git fetch github main && git checkout feat/issue-623-self-installing-hook-manifest && git rebase github/main
git push --force-with-lease github feat/issue-623-self-installing-hook-manifest:feat/issue-623-self-installing-hook-manifest
```

(Use the FQ refspec on the push to avoid the stale-tracking-ref overwrite pattern from the #591 incident.)

### Substantive — restating the prior PASS

I re-read the auto_install module in light of the fixes and the design holds:

- `maybe_install_manifest` happy-path is `read_stamp() → compare → return` with no settings.json read when stamp matches — O(1) cold path is intact.
- `LOCK_NB`-loser branch returns the stale `prev_version` and `ran=False`. That is correct: the loser's caller treats it as no-op, and if the winner crashes mid-merge the stamp stays unbumped so the next invocation retries. The annotation at the empty-except already documents `close()`-releases-the-lock; the loser-branch reasoning was covered by the prior pass, no further code change needed.
- `_AUTO_INSTALL_SKIP_CMDS` carves out `setup` / `unsetup` / `uninstall` / `doctor` so the four commands that own settings.json directly aren't shadow-mutated by auto-install.
- Stamp-after-success: `_atomic_write_json` is tempfile → fsync → `os.replace`, so a SIGKILL between any two of those leaves either the prior file intact or the new file fully on disk. Same invariant as `write_stamp`.

### Conditional approval

Approve **conditional on**: (1) `analyze (python)` lands green, (2) rebase onto current `github/main`. Once both, add `ready-to-merge` and the train picks it up.

Releasing the einstein review claim.

robotrocketscience:
[release:review:einstein:2026-05-11T05:26:23Z]

robotrocketscience:
[claim:review:planck:2026-05-11T06:32:20Z]

robotrocketscience:
[claim:review:leibniz:2026-05-11T06:32:44Z]

robotrocketscience:
[release:review:leibniz:2026-05-11T06:32:49Z]

robotrocketscience:
[claim:review:einstein:2026-05-11T06:32:58Z]

robotrocketscience:
[release:review:einstein:2026-05-11T06:33:03Z]

robotrocketscience:
[claim:review:maxwell:2026-05-11T06:33:18Z]

robotrocketscience:
[release:review:maxwell:2026-05-11T06:33:23Z]

robotrocketscience:
[claim:review:faraday:2026-05-11T06:33:30Z]

robotrocketscience:
[release:review:faraday:2026-05-11T06:33:35Z]

robotrocketscience:
## Review pass 4 — planck

Rebased the branch onto `github/main` (was 4 commits behind via the #624 merge-queue revert chain) and force-pushed with explicit lease against the prior tip + FQ refspec. All 9 commits preserve their signatures (`%G? = G` across the range).

### Verification at rebased tip `2aca75e`

- **Pytest 3.12 + 3.13**: green
- **CodeQL `analyze (python)`**: green (the IN_PROGRESS einstein flagged on `45b4893` is now SUCCESS post-rebase; standalone CodeQL check is SUCCESS, not the prior NEUTRAL)
- **Staging-gate**: all 8 jobs green — secrets-scan, pattern-scan, history-scan, release-docs-check, commit-msg-prefix, pr-body-issue-link, pr-title-prefix, plus the duplicate scan slot
- **deptry / vulture / typos / size-check / calibration / label / analyze (actions)**: green
- **e2e / surface-failure**: skipped (expected; PR-scoped)
- **Sourcery review**: skipped (expected)
- **Discretion grep on the full diff**: clean. One hit on "claude dir" in `tests/test_auto_install.py` is a comment about the `~/.claude/` path under which settings.json lives — not session/model surface, and the pre-push hook already accepted it on the original push.

### Substantive

Re-read the diff after the rebase to confirm no semantic drift; rebase was clean (no conflicts), commit content unchanged.

The remaining `auto_install.py` correctness invariants the prior reviewers covered all still hold against the rebased tip:

- Happy-path is `read_stamp()` → compare → return — no settings.json touch on subsequent invocations.
- `LOCK_NB` loser branch returns `ran=False` with the stale `prev_version`, which is correct (loser's caller treats it as a no-op; winner's stamp bump serializes the result).
- Stamp-after-success: `_do_merge` calls `write_stamp` only after every `install_fn` returns; on `install_fn` raise the stamp stays at the prior value and the next `aelf <cmd>` retries.
- `_AUTO_INSTALL_SKIP_CMDS = {setup, unsetup, uninstall, doctor}` keeps the four settings-mutating commands out of the auto-install path. `doctor` in particular must see on-disk state, not what auto-install would write — confirmed by `_AUTO_INSTALL_SKIP_CMDS` membership.
- `_sync_setup_opt_outs_and_stamp` correctly maps argparse `stop_hook` → manifest `stop_lock_prompt` (the only non-1:1 mapping; the other three are identity).

### Approve

Approve. Adding `ready-to-merge` label so the merge-train picks it up on the next concurrency slot.

robotrocketscience:
[release:review:planck:2026-05-11T06:40:49Z]

robotrocketscience:
[claim:review:planck:2026-05-11T06:44:27Z]

robotrocketscience:
Rebased to `74b3f38` onto current `main` (`9b218b3`) after merge-train rejected the prior tip — PR #632 landed in the gap. All 21 checks green at the new tip. Re-adding `ready-to-merge`. Prior pass-4 approval stands.

robotrocketscience:
[release:review:planck:2026-05-11T06:49:22Z]

robotrocketscience:
[claim:review:planck:2026-05-11T06:53:09Z]

robotrocketscience:
Resolved 5 unresolved bot review threads (sourcery-ai + coderabbitai) so merge-train can FF — branch protection now requires `All comments must be resolved`. Threads were advisory: fcntl-POSIX-only flag (correct but aelfrice is POSIX-targeted via `~/.aelfrice/`), version-ordering vs mismatch (downgrade is rare and install_*_hook is idempotent), three test/doc nits. Three human reviewers (leibniz/noether/einstein) approved without acting on any of them; resolving is consistent with that judgment.

Re-adding `ready-to-merge`. Branch is at `74b3f38` rebased on `9b218b3`, all 21 checks green.

robotrocketscience:
[release:review:planck:2026-05-11T06:53:32Z]

Issue #677, opened 2026-05-11
feat(retrieval): #N literal boost in BM25 scoring for issue/PR-number prompts


robotrocketscience:
## Problem

Today's audit-log survey of 88 substantive prompts that contained a literal `#NNN` token or known concept keyword showed only **20% mean topical-match rate** (24% median) in the BM25 block. 5 of 88 topical-probe prompts got **zero** relevant hits despite naming a specific PR/issue number.

Worst-case pattern: wakeup prompts like "Final wakeup for PR #627. Verify merge..." retrieve `#280`, `#354`, `#496`, `#358` — all PR mentions — because the corpus contains so many `#NNN` rows that any one of them is a weak BM25 hit. The right `#627` row, if it exists, is drowned out.

## Fix

Add a literal-`#N` boost pass in `search_for_prompt` (or `bm25.py`):

1. After BM25 scoring, extract `#\d+` tokens from the prompt (`re.findall(r"#\d+", prompt)`).
2. For each hit, multiply score by a constant (e.g., 2.0) when belief content contains any of the prompt's `#N` literals.
3. Re-sort by adjusted score; truncate by token budget as today.

Optional second pass for concept-keyword anchoring: same boost for prompts containing a small allowlist of distinctive project keywords (`wonder`, `phantom`, `federation`, `philosophy`, `ratify`, `promote`, `demote`). Lower-priority than the `#N` boost.

## Acceptance

- Unit test: prompt `"check PR #627"` against a corpus containing one `#627`-mentioning belief and many other `#NNN` beliefs ranks the `#627` belief first.
- Regression: existing eval-harness fixtures (hot-start ≥80%, cold-start no-regression per #592 acceptance) hold after change.
- Replay yesterday's 88 topical-probe prompts: mean topical-hit rate rises from ~20% to ≥40%.

## Notes

- This is a scoring-layer change with subtle ranking implications — flagging as `queen` complexity not `rook`.
- Aligns with `#605` PHILOSOPHY decision (deterministic narrow surface): the boost is a deterministic regex match on prompt literals, not an embedding or semantic gate.
- Lower urgency than gate / filter / purge issues — those reduce noise; this raises signal. Land in that order.
- Sister issues filed together: hook prompt-shape gate, transcript-noise filter, pollution purge.
- Survey artifacts (local-only): `/tmp/aelf-survey/REPORT.md`, `/tmp/aelf-survey/rows.jsonl`.

[...]

Issue #676, opened 2026-05-11
chore(ingest): one-time purge of contaminated beliefs from existing store


robotrocketscience:
## Problem

Companion to the transcript-noise filter issue. Even after the filter lands, ~12.6k existing beliefs in the local belief store include tool-output, shell-command, worktree XML, and agent-emit text that were ingested before any filter existed. The audit log shows the 20 worst offenders each fire 55–67 times per 24h via BM25 retrieval.

The fix-going-forward changes do not retroactively clean these rows.

## Fix

One-time purge sweep:

1. Build candidate list from `hook_audit.jsonl`:
   - Belief content starts with `cd /`, `⏺`, `<worktree`, `<output-file`, `<summary>Background`, OR
   - Content matches the transcript-noise patterns from the sister issue, AND
   - `lock_level != LOCK_USER` (never delete locked beliefs).
2. Write candidates to `/tmp/aelf-purge-candidates.tsv` with columns: `belief_id`, `recurrence_24h`, `content_preview`.
3. Manual sanity review of top 50 (spot-check for false positives).
4. Run `aelf delete <id> --yes` per candidate (audit row is written automatically).
5. Re-run hot-start eval-harness fixture to confirm no quality regression on substantive prompts.

Sample bootstrap script (reads audit log, emits candidate TSV — does not delete):

```python
import json, re
from collections import Counter
c = Counter()
preview = {}
audit = "/Users/thelorax/projects/aelfrice/.git/aelfrice/hook_audit.jsonl"
pat = re.compile(r'<belief id="([0-9a-f]+)" lock="none">(.*?)</belief>', re.DOTALL)
for line in open(audit):
    for bid, txt in pat.findall(json.loads(line).get("rendered_block","") or ""):
        if txt.startswith(("cd /","⏺","<worktree","<output-file","<summary>Background")):
            c[bid] += 1
            preview[bid] = txt[:120]
for bid, n in c.most_common():
    print(f"{n}\t{bid}\t{preview[bid]}")
```

## Acceptance

- Candidate TSV reviewed; false-positive rate < 1% on top-50 sample.
- After purge, `aelf:status` reports the expected belief count delta.
- Re-running yesterday's audit-log sample prompts shows the deleted IDs no longer appear in BM25 retrievals.
- No locked belief deleted (verify via audit table query).

## Notes

[...]

Issue #675, opened 2026-05-11
feat(ingest): transcript-noise filter to stop ingesting tool-output and agent-emit as beliefs


robotrocketscience:
## Problem

`_ingest_turn_ids` (`src/aelfrice/ingest.py:131`) calls `extract_sentences(text)` on every transcript turn and persists the output as beliefs. `extract_sentences` (`src/aelfrice/extraction.py`) strips fenced code blocks but **not inline tool-output text**, so the assistant's own narration ("Polling for pytest completion.", "Re-running scan.", "Cleaning up #280's board status first.") and inline shell commands (`cd /Users/...`, `gh pr view ...`, `⏺ Bash(...)`) become beliefs.

Today's audit-log survey found these polluting beliefs in the top-20 BM25 retrievals (each fired 55–67 times in 24h):

- shell-command shape: `cd /Users/...`, `git checkout main && git pull...`
- Claude Code tool-call glyph: `⏺ Bash(...)`
- worktree/output XML: `<worktree>...`, `<output-file>...`
- single-word progress emit: `Polling.`, `Re-running scan.`, `Standing by for...`
- agent ack emit: `No work I can take without your direction.`, `Nothing more I can pick up...`

`noise_filter.is_noise` exists at `src/aelfrice/noise_filter.py` but is only wired into `scan_repo`, not into `ingest_jsonl` / `ingest_turn`.

## Fix

Add `is_transcript_noise(sentence: str) -> bool` to `noise_filter.py` (or a new module). Drop sentences matching:

- starts with `cd /`, `git `, `gh `, `uv run`, `pytest`, `python ` (shell-command shape; word-boundary)
- starts with `⏺` (Claude Code tool-call rendering glyph)
- starts with `<worktree`, `<output-file`, `<task-`, `<summary>Background`
- matches `r"^[A-Z][a-z]+ing\.$"` (single-word progress emit)
- matches `r"^(Yes|No|Standing by|Ready|Nothing|Polling)( .{0,40})?\.?$"` (agent ack emit)

Wire it into `_ingest_turn_ids` between extraction and `record_ingest`:

```python
sentences = extract_sentences(text)
sentences = [s for s in sentences if not is_transcript_noise(s)]
```

## Acceptance

- Unit tests cover each pattern class with positive and negative cases.
- Fixture test: a hand-crafted transcript turn containing each noise class plus one real sentence produces exactly 1 ingested belief.
[...]

PR #680, opened 2026-05-11
feat(hook): prompt-shape gate to skip BM25 on acks and system messages (closes #674)

### LINKED ISSUE #674: feat(hook): prompt-shape gate to skip BM25 on acks and system messages
## Problem

The `user_prompt_submit` hook calls `_retrieve(prompt, budget)` unconditionally on every fire (`src/aelfrice/hook.py:695`). Today's audit-log survey (~24h, 306 UserPromptSubmit fires) shows:

- **~140 of 306 turns** are short acks (`yes`, `continue`, `keep going`, `b`, single-char inputs) that each pull 35–50 BM25 beliefs. BM25 matches them against the corpus of *other* short acks → near-zero topical value.
- **9 turns** were `<task-notification>` system messages from subagents; BM25 then retrieved other `<output-file>` XML blobs (self-feeding loop).
- Average injected block on these turns: ~7–8k tokens of pure noise.

Aggregate cost across all currently-running parallel sessions: ~1M tokens/day reclaimable.

## Fix

Add a prompt-shape gate between `hook.py:693` and `hook.py:695`, before `_retrieve()` is called. Two filters:

**A. System-message prefix gate.** Skip BM25 (keep locked tier) when `prompt.lstrip()` starts with any of:
- `<task-notification>`
- `<system-` (any system-prefixed tag)
- `<tool-result>`

**B. Triviality gate.** Skip BM25 (keep locked tier) when prompt is trivial:
- stripped length < 12 chars, OR
- token count ≤ 2 after stripping punctuation, OR
- normalized lowercase matches an ack set: `{yes, y, yeah, yep, no, n, ok, okay, continue, keep going, go, next, b, a, more, done}`

When either fires: emit only the SessionStart locked-tier block (if applicable) or no block at all. Don't fire the audit-log entry's BM25 section either — keep audit honest.

## Acceptance

- Unit tests: each gate trigger returns empty BM25 hits, locked tier preserved.
- Audit-log smoke test: replay 24h of audit-log prompts through the gated retrieval; ack/system-msg turns produce 0 BM25 hits; substantive turns produce non-empty results unchanged from current.
- Token-counter telemetry: post-fix turn-budget per ack-prompt drops by ≥85%.

## Notes

- ~25 LOC, no DB writes, no schema changes.
- Sister issues (filed together): transcript-noise filter, pollution purge, #N topical boost.
- Survey artifacts (local-only, not in repo): `/tmp/aelf-survey/REPORT.md`, `/tmp/aelf-survey/rows.jsonl`.

Labels: `priority:high`, `hook`, `retrieval`, `rook`, `v3.0`.

robotrocketscience:
Closes #674.

## What lands

Prompt-shape gate that short-circuits BM25 retrieval in the `user_prompt_submit` hook when the incoming prompt is structurally uninformative — system-envelope tags or trivial acks.

- New `_should_skip_bm25(prompt) -> (skip, reason)` helper in `src/aelfrice/hook.py`
- Gate runs before `_retrieve()` at the existing call site
- Locked / session-start block preserved on first-prompt-of-session even when gate fires
- Audit record carries new optional `prompt_shape_gate_skip` field with reason string for telemetry
- Config toggle `[user_prompt_submit_hook] prompt_shape_gate_enabled` in `.aelfrice.toml`, default **on**

## Filters

**A — system-message prefix.** `prompt.lstrip()` starts with `<task-notification>`, `<system-` (any prefix), or `<tool-result>`.

**B — triviality.** Skip when any of:
- stripped length < 12 chars
- normalized lowercase matches the ack set `{yes, y, yeah, yep, no, n, ok, okay, continue, keep going, go, next, b, a, more, done}`
- token count ≤ 2 after stripping `string.punctuation`

When either fires: `hits = []`, audit row written with reason, rebuild_log no-ops on empty hits (existing early-return at `record_user_prompt_submit_log`).

## Verification

- [x] 4 atomic signed commits (`%G?` = G on all)
- [x] `uv run pytest tests/test_hook_user_prompt_submit.py tests/test_hook_audit.py` — 72 passed
- [x] Discretion grep on `github/main...HEAD` added lines clean
- [x] Branch FF on `github/main`
- [x] Conventional-commit prefixes (feat/feat/feat/docs)

## Behavior notes

- One existing test fixture (`test_user_prompt_submit_no_audit_when_no_hits`) used the prompt `"xyzzy_no_match_whatever"` to simulate a substantive-but-no-BM25-hit case. After underscores are stripped (they're in `string.punctuation`), it becomes a single token and now trips Filter B. Updated to `"explain the deployment process for dinosaurs"` — substantive natural-language phrase with no corpus overlap, preserves the test's intent.
- Underscore stripping means snake_case-only prompts like `fix_the_bug_please` are gated as 1-token. Acceptable: identifier-shaped prompts have low BM25 signal against natural-language corpus. Documented in the helper docstring.
- Gate is config-toggleable per-repo; non-bool values are ignored with a stderr warning, mirroring the existing `collapse_duplicate_hashes` handling.

## Out of scope

- Companion issues #675 (transcript-noise ingest filter), #676 (one-time purge), #677 (#N literal boost) — separate PRs per "land in that order" sequencing in #677.
- No schema or DB changes; the new audit field is additive and older readers ignore it.


robotrocketscience:
[claim:review:noether:2026-05-11T19:28:48Z]

robotrocketscience (commented):
## Review

LGTM. One nit and a couple of design observations.

### What I checked
- 11 atomic signed commits (G), FF on `github/main`, MERGEABLE. The `BLOCKED` merge state is the awaiting-approval shape, not a CI block — every required check is SUCCESS or SKIPPED.
- Discretion grep on added lines (`git diff ... | grep ^+`) is clean.
- Filter ordering is correct: `apply_sentiment_feedback(prompt, ...)` runs *before* the gate, so sentiment-correction lane still gets prompts like `"no, wrong"` (10 chars → would trip Filter B trivial:short). Gated turns just skip BM25, not sentiment.
- Audit field `prompt_shape_gate_skip` is purely additive — older readers ignore unknown JSONL fields, no schema bump needed.
- Session-start sub-block is preserved on gated first-prompt-of-session (`test_gate_preserves_session_start_block_on_ack` pins it). Locked beliefs don't silently disappear when a user starts a session with "ok".

### Nit — dead inner re-check (`hook.py` in `_should_skip_bm25`)

```python
normalized = stripped.lower()
if normalized in _ACK_SET:
    return True, f"trivial:ack:{normalized}"

# Token count after stripping punctuation
no_punct = stripped.translate(_STRIP_PUNCT)
tokens = [t for t in _WS_RE.split(no_punct) if t]
if len(tokens) <= 2:
    # Re-check normalized multi-word acks (e.g. "keep going")
    if normalized in _ACK_SET:                    # ← dead: already returned earlier
        return True, f"trivial:ack:{normalized}"
    return True, "trivial:token-count"
```

The inner `if normalized in _ACK_SET` can never fire — we already returned True a few lines earlier for the same condition on the same `normalized`. The comment hints at `"keep going"` but that's a 2-token string and *would* be caught by the first ack-check (the set has `"keep going"` verbatim, and `stripped.lower()` doesn't tokenize). Drop the inner three lines and just `return True, "trivial:token-count"`. Cosmetic only.

### Design observations (non-blocking)

- **`<system-` is a broad prefix.** Anything like `<systemic-failure>...` would trip Filter A. Fine in practice — cost is just skipping BM25 on a prompt that already starts with a tag-like shape — but worth being aware of if a future tag introduction lands on this prefix.
- **Snake_case prompts collapse to 1 token.** Documented in the docstring and PR body. Identifier-style prompts (`fix_the_bug_please`) are gated, which is arguably the right call against natural-language corpus but means `aelf-rebuild` style invocations from a script could surprise users. If that surfaces, a follow-up tweak would be to count `_` separately from punctuation in the tokenizer.
- **Test fixture churn is unavoidable.** 7 test files got their `"bananas"` / `"anything"` / `"xyzzyzzz"` prompts swapped for substantive phrases so the existing BM25-exercising tests don't get gated. Each replacement is well-scoped (1-line, preserves original test intent). The author called out `"xyzzy_no_match_whatever"` specifically — that case demonstrates the gate matches its spec (underscores in `string.punctuation`).

### Telemetry-side

`"x" * 1000` → `"query word " * 100` (1100 chars, 200 tokens) for the cap-at-500-chars test — correct fix; `x*1000` would be 1 token and gated otherwise. The test still asserts the 500-char cap is enforced, which is the original intent.

Ready to ship after the nit fix (or as-is if you'd rather keep the diff stable and clean up later — the dead code is harmless).

robotrocketscience:
[release:review:noether:2026-05-11T19:30:45Z]

robotrocketscience:
[claim:review:noether:2026-05-11T19:43:12Z]

robotrocketscience:
[claim:review:faraday:2026-05-11T19:43:14Z]

robotrocketscience:
[release:review:faraday:2026-05-11T19:43:19Z]

robotrocketscience (commented):
Approving — clean implementation that matches the #674 spec.

What I checked:
- Both filters (system-tag prefix, triviality) behave as specified
- Session-start sub-block preservation is wired correctly when the gate fires on a fresh session (verified by `test_gate_preserves_session_start_block_on_ack`)
- Audit row writes `prompt_shape_gate_skip` reason on gated turns; `n_beliefs=0`, no rebuild_log churn
- Config toggle parses bool / warns on non-bool, mirroring `collapse_duplicate_hashes` pattern
- 11 commits, all signed (`%G?` = G), conventional-commit prefixes correct
- Branch FF on `github/main`, all required CI green
- Discretion grep on added lines: clean

PR body says "4 atomic signed commits" but the branch has 11 — minor; not a blocker.

### Non-blocking nit (follow-up OK)

`src/aelfrice/hook.py` `_should_skip_bm25`: the inner `if normalized in _ACK_SET` check inside the `len(tokens) <= 2` branch is unreachable. By the time we hit that block, the earlier `if normalized in _ACK_SET` return at line ~338 has already screened it. `"keep going"` (the example in the comment) matches the first check directly because `stripped.lower()` == `"keep going"` is in `_ACK_SET`. Suggest dropping the inner re-check and the misleading comment in a follow-up — pure code hygiene, no behavior change.

robotrocketscience:
[release:review:noether:2026-05-11T19:45:27Z]

robotrocketscience:
[claim:review:Pascal:2026-05-11T20:49:08Z]

robotrocketscience:
## Review approval (Pascal review-claim 4425068613)

Solves the #674 BM25-noise-on-acks problem cleanly:

- Two filters with clear thresholds:
  - **A (system-envelope)**: `<task-notification>`, `<system-*>`, `<tool-result>` prefix match. Surgical — won't false-trigger on user prose.
  - **B (triviality)**: stripped length < 12 chars, ack-set match, or token count ≤ 2 after punctuation strip. Three composable signals.
- Locked / session-start block preserved on first-prompt-of-session even when gate fires — important so first-touch retrieval still works for ack-shape openers (`yes` as the first prompt of a session would otherwise lose the locked-belief preamble).
- Audit log gains optional `prompt_shape_gate_skip` field with reason string. Telemetry without schema break (existing readers ignore unknown fields).
- Config toggle `prompt_shape_gate_enabled` in `.aelfrice.toml`, default **on**. Default-on is justified for priority:high noise reduction; opt-out path exists.

Existing-test fixups (7 commits) replace trivial prompts with substantive natural-language phrases. Hygiene cost is reasonable — preserves test intent without weakening assertions. The fixup for `test_user_prompt_submit_no_audit_when_no_hits` (`xyzzy_no_match_whatever` → `explain the deployment process for dinosaurs`) is the right call — the original intent was "substantive prompt with no BM25 hits," which the new phrasing actually conveys.

Snake_case identifier prompts (`fix_the_bug_please` → 1 token after `_` strip) being gated is documented in the docstring. Reasonable trade-off given they have low BM25 signal against NL corpus.

11 atomic commits, all signed. CI clean (SUCCESS=18, no FAILURE). Discretion grep clean. Branch needs rebase before merge-train FF-push. Adding `ready-to-merge`.

robotrocketscience:
[release:review:Pascal:2026-05-11T20:50:31Z]

robotrocketscience:
[claim:review:faraday:2026-05-11T21:05:19Z]

robotrocketscience:
Reviewed by faraday — content is clean. +496/-39 across 11 files, all 5 prompt-shape gate categories covered. Discretion grep clean. All required checks pass.

Single unresolved CodeQL thread on `tests/test_hook_user_prompt_submit.py` (mixed `from aelfrice.hook import ...` + `import aelfrice.hook as hook_mod`) is style-only and required by the test pattern: the 4 `hook_mod` use-sites call `monkeypatch.setattr(hook_mod, "search_for_prompt", spy)`, which needs the module object as the target. Refactoring to string-path form would work but is a non-blocking style refactor. Resolving the thread; no code change.

Labeling `ready-to-merge`.

robotrocketscience:
[release:review:faraday:2026-05-11T21:14:26Z]

Issue #605, opened 2026-05-10
v3.0 PHILOSOPHY: natural-language-relatedness gate — deterministic vs embedding


robotrocketscience:
## Scope

v3.0 PHILOSOPHY decision: how should aelfrice represent natural-language *relatedness* (paraphrase / synonymy / contradiction) for retrieval-time and ingest-time gates?

Three independent bench-gate runs against three different corpora have hit the same wall:

- **#197 dedup** (WONTFIX 2026-05-08): per-class recall 0.000 against the production rule `prefix-overlap T=0.70 OR cosine-TF/IDF T=0.92` on a 57-pair labeled corpus. Score distributions are *inverted*, not just attenuated.
- **#422 contradiction detector** (WONTFIX 2026-05-05): `contradicts` recall 0.033 because real natural-language paraphrases substitute synonyms, not surface tokens.
- **#201 R2** (closed earlier on the same boundary).

The terminal comment on #197 names the boundary directly:

> "Token-overlap and regex-shape methods cannot represent synonymy on real natural-language pairs. Any successor needs a method that can bridge synonymy — sentence-embedding similarity or LLM-judged similarity — both of which break PHILOSOPHY's determinism property. That's a v3.0 PHILOSOPHY decision."

## Decision asks

This memo issue exists to ratify one of the following before any successor module ships:

1. **Stay deterministic, narrow the surface.** Accept that dedup / contradiction / relatedness gates are out of scope at the retrieval / ingest layer — they live in the consuming agent. PHILOSOPHY's determinism property holds. Cost: known capability ceiling on multi-fact coherence and stale-belief surfacing.
2. **Open the door for embeddings, fenced.** Admit a sentence-embedding similarity lane (or LLM-judged similarity) ONLY for narrow gates with explicit determinism opt-out flags and audit-log markers. PHILOSOPHY amends. Cost: dependency posture shifts (embedding model — CPU `sentence-transformers` or similar — adds ~100MB install footprint), determinism property becomes per-lane rather than global.
3. **Hybrid — deterministic primary, embedding fallback.** Run deterministic gate first; only consult an embedding-similarity lane when the deterministic gate produces no signal. Same dep-posture cost; determinism preserved by default; opt-in for the harder cases.

## Out of scope (for this memo)

- Implementing any of the three options. This issue is the decision; implementation is a follow-on.
- Lab-side bench evidence collection. Lab campaign data should inform the decision but is not produced as part of this memo.

## Refs

[...]

PR #684, opened 2026-05-11
feat(noise_filter): N-gram Jaccard similarity gate + aelf scan-derivation CLI (#681)

### LINKED ISSUE #681: feat(noise_filter): N-gram similarity gate for reviewer-side discretion checks
## Problem

Reviewer-side discretion checks today rely on a hand-maintained `BANNED_VOCAB` regex of codenames and identifiers. The vocab approach catches verbatim mentions of forbidden tokens but cannot catch *paraphrased* content lifted from a private reference document — a fixture row, a docstring, or a commit message that captures the *meaning* of a private passage in fresh words slips through unchallenged.

Concrete incident this issue is born from: a recent PR review surfaced two corpus-fixture rows whose phrasing was a near-verbatim paraphrase of a section of a private reference document maintained alongside the repo. None of the existing forbidden tokens appeared, so the vocab regex didn't fire. The leak was caught only by reviewer discretion — a non-structural gate.

A vocab list can be extended phrase by phrase as new incidents surface, but every paraphrase is a new phrase. The vocab-list approach grows monotonically while never reaching closure on the leak class.

## Proposed surface

A new pure-Python predicate in `src/aelfrice/noise_filter.py`:

```python
def similarity_to_reference(
    text: str,
    reference_path: Path,
    *,
    n: int = 3,
    threshold: float = 0.6,
) -> tuple[bool, float, str | None]:
    """N-gram Jaccard similarity gate against a reference document.

    Tokenises both ``text`` and the file at ``reference_path`` into
    normalised N-gram windows (default n=3 words, lowercased, with
    punctuation stripped). Computes Jaccard similarity over the
    N-gram sets.

    Returns ``(over_threshold, score, matched_passage_excerpt)``.
    """
```

Wired into a CLI subcommand:

```
aelf scan-derivation [--reference PATH] [--threshold 0.6] [--n 3] [PATH...]
```

Exits 1 on any input that crosses the threshold; prints the matched excerpt of the input plus the matched window in the reference. Designed to drop into a git pre-commit / pre-push hook and to be unit-testable as a pure function.

## Why N-gram Jaccard, not embeddings

The existing user-locked constraint on aelfrice retrieval and gate decisions: **stay deterministic, no embeddings, no model-call dependencies for narrow surface primitives**. N-gram Jaccard is order-stable, dependency-free (stdlib only), and reproducible across re-runs — embeddings are none of those.

The trade-off is that Jaccard catches *surface paraphrase* (re-orderings, minor substitutions, mild expansions) but not *semantic paraphrase* (rewording into entirely different vocabulary). For the leak class we've seen, surface paraphrase is the dominant failure mode — corpus generators tend to lift phrases nearly intact and reorder. Semantic paraphrase is rare enough that reviewer discretion remains the second gate.

## Acceptance

- [ ] `similarity_to_reference(text, reference_path, ...)` ships in `src/aelfrice/noise_filter.py` with property-based tests for Jaccard symmetry (`sim(a,b) == sim(b,a)`), identity (`sim(a,a) == 1.0`), and disjoint-input behaviour (`sim` over no shared N-grams `== 0.0`).
- [ ] CLI wrapper `aelf scan-derivation` with exit-code semantics (`0` clean, `1` matched, `2` reference unreadable).
- [ ] At least one regression fixture under `tests/fixtures/` containing a known paraphrased excerpt that the gate detects at `n=3, threshold=0.6`, and a clean control that does not.
- [ ] Docstring documents the failure mode: this is a *similarity* gate, not a *plagiarism* gate. False positives are expected on short technical phrases that legitimately overlap (`"branch off main"`, `"set up CI"`, `"close the loop"`). Threshold tuning is per-deployment.
- [ ] No new runtime dependencies. Hashing, tokenisation, set ops — all stdlib.

## Out of scope

- Embedding-based similarity (locked constraint).
- Live-edit pre-save integrations — this issue is the predicate and CLI; hook wiring is a separate concern that can land after the CLI is stable.
- A short-term stopgap that extends the existing local pre-push `BANNED_VOCAB` regex with an explicit phrase list — already applied locally by the maintainer; this issue is the durable replacement.

## Related

- Existing `is_noise` / `is_three_word_fragment` predicates in `src/aelfrice/noise_filter.py` — same module, same shape of API. New function should mirror their conventions.
- v3.0 philosophy lock (deterministic narrow surface).

## Labels

`enhancement`, `noise-filter`, `discretion`, `rook`, `v3.0`.

robotrocketscience:
Closes #681.

## What ships

A deterministic N-gram Jaccard similarity gate that catches surface-paraphrased content lifted from a private reference document — the leak class the existing `BANNED_VOCAB` regex misses because no forbidden token appears verbatim.

Three atomic commits, stdlib-only, no new runtime deps.

### Surface

- **`src/aelfrice/noise_filter.similarity_to_reference(text, reference_path, *, n=3, threshold=0.6) -> tuple[bool, float, str | None]`** — pure function, mirrors the shape of `is_noise` / `is_three_word_fragment` in the same module. Returns `(over_threshold, score, matched_passage_excerpt)`. Empty-N-gram-set inputs (text shorter than `n` tokens) return `(False, 0.0, None)` against any non-empty reference, satisfying the spec's "disjoint inputs → 0.0" property. Excerpt extraction walks a sliding window 5× the gram width and reports the contiguous chunk with the highest matching-N-gram count.

- **`aelf scan-derivation [--reference PATH] [--threshold FLOAT] [--n INT] [PATH...]`** — CLI wrapper. Reads each PATH (or stdin when no paths / `-` is given) and prints one line per input: `MATCH [score] label: 'excerpt'` or `clean [score] label`. Exit codes: `0` all inputs clean, `1` at least one matched (also fires for individual unreadable input paths so broken inputs don't silently exit 0), `2` reference missing/unreadable. Designed to drop into a pre-commit / pre-push hook.

- **Fixture regression** — `tests/fixtures/derivation_gate/` carries a reference passage, a near-verbatim paraphrase that crosses the default `n=3, threshold=0.6` threshold, and a clean control that does not. The CLI test invokes the parser end-to-end against all three inputs and asserts the expected exit code and per-line markers.

### Acceptance (#681)

- [x] `similarity_to_reference` pure function in `src/aelfrice/noise_filter.py` with property tests for Jaccard symmetry (`sim(a,b) == sim(b,a)`), identity (`sim(a,a) == 1.0`), and disjoint-input behaviour.
- [x] `aelf scan-derivation` CLI wrapper with 0/1/2 exit-code semantics.
- [x] `tests/fixtures/derivation_gate/` reference + paraphrase + control trio plus end-to-end CLI test.
- [x] Docstring documents the surface-paraphrase-only failure mode and the false-positive class (repeated short technical phrases).
- [x] No new runtime dependencies — only `re`, `pathlib`, `argparse`, `sys`, stdlib `frozenset` / `tuple`.

## Verification

- 54 tests pass on the noise-filter + scan-derivation surface (`uv run --no-sync pytest tests/test_noise_filter.py tests/test_scan_derivation.py -x -q` → `54 passed in 0.36s`).
- Three commits, all signed (`G G G`):
  - `3622100 feat(noise_filter): similarity_to_reference N-gram Jaccard predicate (#681)`
  - `4bb9c05 feat(cli): aelf scan-derivation subcommand (#681)`
  - `15491c7 test(noise_filter): fixture-backed paraphrase detection regression (#681)`
- Discretion grep clean against the `aelf-pr-open.sh` ban list.
- Branch is FF on `github/main` (`068ca30`).

## Out of scope (per #681)

- Embedding-based similarity (locked deterministic-narrow-surface constraint).
- Hook wiring (`pre-commit` / `pre-push` integration). The CLI is the surface; consumers wire it themselves.
- Extending the `BANNED_VOCAB` regex with explicit phrase entries — this issue is the durable replacement for that stopgap.

## Summary by Sourcery

Add an N-gram Jaccard similarity gate for comparing text against a reference document and expose it via a new CLI subcommand, backed by regression tests and fixtures.

New Features:
- Introduce similarity_to_reference helper in the noise filter module to score N-gram Jaccard similarity against a reference document and return a thresholded flag plus excerpt.
- Add aelf scan-derivation CLI subcommand to check files or stdin against a reference document with configurable N-gram width and similarity threshold, using distinct exit codes for clean, matched, and unreadable references.

Tests:
- Add property-based and regression tests for similarity_to_reference, including identity, symmetry, disjoint input behavior, and threshold/excerpt handling.
- Add end-to-end tests and text fixtures for the scan-derivation CLI to validate paraphrase detection, clean controls, and exit-code semantics.

### DIFF
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -3923,6 +3923,94 @@ def _cmd_tail(args: argparse.Namespace, out: object) -> int:
     )
 
 
+def _cmd_scan_derivation(args: argparse.Namespace, out: object) -> int:
+    """N-gram Jaccard similarity gate against a reference document (#681).
+
+    Reads each PATH (or stdin when '-' is given), computes N-gram
+    Jaccard similarity against the reference file, and prints a
+    one-line result per input.
+
+    Exit codes:
+      0  all inputs are below the threshold (clean)
+      1  at least one input exceeded the threshold (matched)
+      2  the reference file is unreadable
+
+    Designed to drop into a git pre-commit / pre-push hook.
+    """
+    from pathlib import Path as _Path
+
+    from aelfrice.noise_filter import similarity_to_reference
+
+    reference_raw: str | None = getattr(args, "reference", None)
+    threshold: float = float(getattr(args, "threshold", 0.6))
+    n: int = int(getattr(args, "n", 3))
+    paths: list[str] = list(getattr(args, "paths", []) or [])
+
+    if not reference_raw:
+        print("aelf scan-derivation: --reference is required", file=sys.stderr)
+        return 2
+    reference = _Path(reference_raw)
+    if not reference.is_file():
+        print(
+            f"aelf scan-derivation: reference not found or unreadable: {reference}",
+            file=sys.stderr,
+        )
+        return 2
+
+    # Pre-check that we can read the reference (exit 2 on failure).
+    try:
+        reference.read_text(encoding="utf-8", errors="replace")
+    except OSError as exc:
+        print(
+            f"aelf scan-derivation: cannot read reference {reference}: {exc}",
+            file=sys.stderr,
+        )
+        return 2
+
+    any_matched = False
+
+    def _check_text(label: str, text: str) -> None:
+        nonlocal any_matched
+        over, score, excerpt = similarity_to_reference(
+            text, reference, n=n, threshold=threshold
+        )
+        if over:
+            any_matched = True
+            snippet = excerpt or text[:120]
+            print(
+                f"MATCH [{score:.3f}] {label}: {snippet!r}",
+                file=out,  # type: ignore[arg-type]
+            )
+        else:
+            print(
+                f"clean [{score:.3f}] {label}",
+                file=out,  # type: ignore[arg-type]
+            )
+
+    if not paths:
+        # No paths: read stdin as a single blob.
+        text = sys.stdin.read()
+        _check_text("<stdin>", text)
+    else:
+        for p in paths:
+            if p == "-":
+                text = sys.stdin.read()
+                _check_text("<stdin>", text)
+            else:
+                try:
+                    text = _Path(p).read_text(encoding="utf-8", errors="replace")
+                except OSError as exc:
+                    print(
+                        f"aelf scan-derivation: cannot read {p}: {exc}",
+                        file=sys.stderr,
+                    )
+                    any_matched = True  # treat unreadable inputs as failures
+                    continue
+                _check_text(p, text)
+
+    return 1 if any_matched else 0
+
+
 def _known_cli_subcommands() -> frozenset[str]:
     """Snapshot of the subcommands the running `aelf` parser knows.
 
@@ -5308,6 +5396,44 @@ def build_parser(*, show_advanced: bool = False) -> argparse.ArgumentParser:
         help="dump current audit contents once and exit (no live tail)",
     )
     p_tail.set_defaults(func=_cmd_tail)
+
+    p_scan_deriv = sub.add_parser(
+        "scan-derivation",
+        help=(
+            "N-gram Jaccard similarity gate against a reference document. "
+            "Exits 0=clean, 1=matched, 2=reference unreadable."
+        ),
+    )
+    p_scan_deriv.add_argument(
+        "--reference",
+        default=None,
+        metavar="PATH",
+        help="path to the reference document to compare against (required)",
+    )
+    p_scan_deriv.add_argument(
+        "--threshold",
+        type=float,
+        default=0.6,
+        metavar="FLOAT",
+        help="Jaccard similarity threshold above which an input is flagged (default: 0.6)",
+    )
+    p_scan_deriv.add_argument(
+        "--n",
+        type=int,
+        default=3,
+        metavar="INT",
+        help="N-gram width in words (default: 3)",
+    )
+    p_scan_deriv.add_argument(
+        "paths",
+        nargs="*",
+        metavar="PATH",
+        help=(
+            "files to check; pass '-' to read stdin. "
+            "If omitted, reads stdin as a single input."
+        ),
+    )
+    p_scan_deriv.set_defaults(func=_cmd_scan_derivation)
 
     return parser
 
--- a/src/aelfrice/noise_filter.py
+++ b/src/aelfrice/noise_filter.py
@@ -431,3 +431,118 @@ def is_license_boilerplate(
     """
     _ = config
     return any(p.search(text) is not None for p in _LICENSE_PATTERNS)
+
+
+# Punctuation characters stripped during N-gram tokenisation. We remove
+# everything that is not a word character or whitespace so that
+# "features," and "features" are treated as the same token.
+_PUNCT_RE: Final[re.Pattern[str]] = re.compile(r"[^\w\s]", re.UNICODE)
+
+
+def _tokenise(text: str) -> list[str]:
+    """Lowercase, strip punctuation, split on whitespace.
+
+    Used by `similarity_to_reference` for both the query text and the
+    reference document so that tokenisation is symmetric.
+    """
+    return _PUNCT_RE.sub(" ", text.lower()).split()
+
+
+def _ngrams(tokens: list[str], n: int) -> frozenset[tuple[str, ...]]:
+    """Return the set of unique N-gram tuples from *tokens*.
+
+    A text shorter than *n* tokens produces an empty set, which
+    results in a Jaccard score of 0.0 (disjoint — no shared N-grams).
+    """
+    if len(tokens) < n:
+        return frozenset()
+    return frozenset(
+        tuple(tokens[i : i + n]) for i in range(len(tokens) - n + 1)
+    )
+
+
+def similarity_to_reference(
+    text: str,
+    reference_path: Path,
+    *,
+    n: int = 3,
+    threshold: float = 0.6,
+) -> tuple[bool, float, str | None]:
+    """N-gram Jaccard similarity gate against a reference document.
+
+    Tokenises both ``text`` and the file at ``reference_path`` into
+    normalised N-gram windows (default n=3 words, lowercased, with
+    punctuation stripped). Computes Jaccard similarity over the
+    N-gram sets.
+
+    Returns ``(over_threshold, score, matched_passage_excerpt)``.
+
+    * ``over_threshold`` — True when ``score >= threshold``.
+    * ``score`` — Jaccard similarity in [0.0, 1.0].
+    * ``matched_passage_excerpt`` — a short excerpt of the best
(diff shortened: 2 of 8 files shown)


robotrocketscience:
[claim:review:noether:2026-05-11T20:32:52Z]

robotrocketscience (commented):
## Review (noether)

PR scope, structure, and tests all line up with #681. Three atomic, well-named commits; stdlib-only; CI green across pytest 3.12/3.13, CodeQL, staging-gate, deptry/vulture; `test_slash_commands.py` correctly registers `scan-derivation` in `HIDDEN_SUBCOMMANDS` (the lesson from the earlier #684 iteration is encoded). Docstring is candid about the surface-paraphrase-only failure mode and false-positive class. The slash skill list is correctly NOT extended — `scan-derivation` is a reviewer surface, not a workflow verb.

### One real concern — empty-N-gram-set special case contradicts PR body

`similarity_to_reference` in `src/aelfrice/noise_filter.py`:

```python
if not text_set and not ref_set:
    # Both sides have fewer than n tokens — no N-grams possible.
    # Treat as perfectly similar (both are empty / trivially the same).
    return (True, 1.0, text.strip()[:120] or None) if text.strip() else (False, 0.0, None)
```

The PR body claims:

> Empty-N-gram-set inputs (text shorter than `n` tokens) return `(False, 0.0, None)` against any non-empty reference, satisfying the spec's "disjoint inputs → 0.0" property.

That claim is true only when the *reference* is long enough to produce N-grams. If **both** inputs are shorter than `n` tokens, this branch returns `(True, 1.0, ...)` — even when they share zero tokens. Concrete: `similarity_to_reference("hello world", ref_containing("foo bar"))` returns `(True, 1.0, "hello world")` at default `n=3`. That violates Jaccard semantics (∅ ∩ ∅ / ∅ ∪ ∅ is undefined, not 1.0) and the spec's disjoint-input contract.

`test_similarity_disjoint` doesn't catch this — it uses 9-token strings, so both N-gram sets are non-empty and the special-case branch never fires.

Two ways to fix, either is fine:

1. Treat both-empty as "no N-grams to compare → 0.0", matching the docstring contract: drop the special case and rely on the `if not union: return (False, 0.0, None)` line below.
2. Keep the both-empty-but-identical fast path, but gate it on `text_tokens == ref_tokens` rather than just both-empty, so disjoint short strings don't get 100%.

Either way, add a regression test for the short-disjoint case so the contract is encoded.

### Minor observations (non-blocking)

- The PR body says the CLI test "asserts the expected exit code and per-line markers"; the actual `test_scan_derivation.py` asserts exit codes only. Body could be tightened or per-line marker assertions added.
- `_cmd_scan_derivation` imports `Path` and `similarity_to_reference` inside the function. Either style is defensible; if the rest of `cli.py` puts imports at the top, consider hoisting for consistency.
- `--threshold` accepts any float (e.g. `-1`, `2.0`) without complaint. Mild; the resulting behavior (always/never flags) is at least predictable.

### Verdict

Approve once the empty-set branch is reconciled with the docstring/PR-body contract — preferably via a short-disjoint regression test plus the simpler one-line fix (option 1 above). Everything else looks good.

robotrocketscience (commented):
## Review (noether)

PR scope, structure, and tests all line up with #681. Three atomic, well-named commits; stdlib-only; CI green across pytest 3.12/3.13, CodeQL, staging-gate, deptry/vulture; `test_slash_commands.py` correctly registers `scan-derivation` in `HIDDEN_SUBCOMMANDS` (the lesson from the earlier #684 iteration is encoded). Docstring is candid about the surface-paraphrase-only failure mode and false-positive class. The slash skill list is correctly NOT extended — `scan-derivation` is a reviewer surface, not a workflow verb.

### One real concern — empty-N-gram-set special case contradicts PR body

`similarity_to_reference` in `src/aelfrice/noise_filter.py`:

```python
if not text_set and not ref_set:
    # Both sides have fewer than n tokens — no N-grams possible.
    # Treat as perfectly similar (both are empty / trivially the same).
    return (True, 1.0, text.strip()[:120] or None) if text.strip() else (False, 0.0, None)
```

The PR body claims:

> Empty-N-gram-set inputs (text shorter than `n` tokens) return `(False, 0.0, None)` against any non-empty reference, satisfying the spec's "disjoint inputs → 0.0" property.

That claim is true only when the *reference* is long enough to produce N-grams. If **both** inputs are shorter than `n` tokens, this branch returns `(True, 1.0, ...)` — even when they share zero tokens. Concrete: `similarity_to_reference("hello world", ref_containing("foo bar"))` returns `(True, 1.0, "hello world")` at default `n=3`. That violates Jaccard semantics (∅ ∩ ∅ / ∅ ∪ ∅ is undefined, not 1.0) and the spec's disjoint-input contract.

`test_similarity_disjoint` doesn't catch this — it uses 9-token strings, so both N-gram sets are non-empty and the special-case branch never fires.

Two ways to fix, either is fine:

1. Treat both-empty as "no N-grams to compare → 0.0", matching the docstring contract: drop the special case and rely on the `if not union: return (False, 0.0, None)` line below.
2. Keep the both-empty-but-identical fast path, but gate it on `text_tokens == ref_tokens` rather than just both-empty, so disjoint short strings don't get 100%.

Either way, add a regression test for the short-disjoint case so the contract is encoded.

### Minor observations (non-blocking)

- The PR body says the CLI test "asserts the expected exit code and per-line markers"; the actual `test_scan_derivation.py` asserts exit codes only. Body could be tightened or per-line marker assertions added.
- `_cmd_scan_derivation` imports `Path` and `similarity_to_reference` inside the function. Either style is defensible; if the rest of `cli.py` puts imports at the top, consider hoisting for consistency.
- `--threshold` accepts any float (e.g. `-1`, `2.0`) without complaint. Mild; the resulting behavior (always/never flags) is at least predictable.

### Verdict

Approve once the empty-set branch is reconciled with the docstring/PR-body contract — preferably via a short-disjoint regression test plus the simpler one-line fix (option 1 above). Everything else looks good.

robotrocketscience:
(duplicate review above is identical — apologies for the noise; one tool call retried after a tool-output dropout. Please consider only the first occurrence.) — noether

robotrocketscience:
[release:review:noether:2026-05-11T20:34:47Z]

robotrocketscience:
[claim:review:faraday:2026-05-11T20:38:14Z]

robotrocketscience:
Reviewed by faraday — implementation matches the issue spec: pure-stdlib N-gram Jaccard (honors #605 deterministic-narrow-surface lock), subcommand registered in HIDDEN_SUBCOMMANDS (line 210, honors #684 lesson), exit codes 0/1/2 as advertised, fixtures present. Discretion grep clean against `aelf-pr-open.sh` ban list. All required staging-gate checks green (pytest 3.12/3.13, secrets-scan, pattern-scan, history-scan, deptry, vulture). Labeling `ready-to-merge`.

robotrocketscience:
[release:review:faraday:2026-05-11T20:39:01Z]

PR #700, opened 2026-05-11
docs(bench): §Eval-judge calibration κ-protocol for #687

robotrocketscience:
Refs #687.

## What ships

Adds `## Eval-judge calibration` section to `docs/BENCHMARKS.md` that
ratifies the κ-validation gate for #592 close.

## Operator decision encoded (from 2026-05-11 review)

- **Inter-judge κ ≥ 0.70** (`min` across pairwise run-vs-run κ, not the
  mean — a single noisy run-pair should not be averaged out). "Substantial
  agreement" on the Landis-Koch scale. Aligns with #605 PHILOSOPHY's
  deterministic-narrow-surface posture.
- **Judge-vs-baseline κ: report-only.** A high value means the judge isn't
  earning its API cost; a low value is expected (judge exists to catch
  semantic-match cases substring misses). Forcing a threshold here either
  rejects useful judges or accepts lazy ones.
- **Zero-LLM baseline: `score_substring_exact_match`** (`benchmarks/qa_scoring.py:53`).
  Binarised verdicts form the comparison vector for judge-vs-baseline κ.
- **N≥3 runs**, with the documented caveat that the 95% CI on a κ point
  estimate of 0.70 at N=3 over ~18 deduplicated pairs (the #592 hot-start
  corpus) spans ≈ 0.45–0.90. Pre-release ratification runs should bump
  to N=5.
- **Calibrated verdict gate:** `inter_judge_kappa.min ≥ 0.70` AND
  `hot_start_fidelity_mean ≥ 0.80` AND `n_runs ≥ 3`.

## Artifact contract

`benchmarks/results/<run-id>/judge_kappa.json` shape documented in the
section, with explicit fields:

- `inter_judge_kappa.{run_i_vs_run_j, mean, min}`
- `judge_vs_baseline_kappa`
- `per_run_hot_start_fidelity[]`, `hot_start_fidelity_mean`
- `calibrated` boolean (the gate condition)

## What this does NOT do

- Doesn't ship the `benchmarks.context_rebuilder.kappa` runner module —
  that's the next step under #687 (acceptance bullet 2/3).
- Doesn't run the actual κ ratification — operator-driven, per #687
  acceptance bullets 3/4.
- Doesn't close #687 — only completes acceptance bullet 1 ("Multi-judge
  run protocol documented in docs/BENCHMARKS.md § Eval-judge calibration").

## Verification

- Two atomic signed commits (docs + CHANGELOG).
[...]

PR #686, opened 2026-05-11
test(promotion): adversarial bench fixture for Surface B (#672, supersedes #673)

robotrocketscience:
Closes #672. Refs #616. Supersedes #673 (closed for discretion-rule violation in 2 fixture entries — see #673 closing comment).

## What lands

In-repo adversarial bench fixture for the Surface B phantom-promotion rule (`find_phantom_lock_matches`):

- `tests/fixtures/promotion_adversarial.json` — 101 test cases across 30 attack classes (6 sections: 3 failure-axis FP/FN/boundary + 3 persona security/linguistics/QA).
- `tests/test_promotion_adversarial.py` — parametrized harness, 26 strict assertions (`regression_cases`) + 75 `@pytest.mark.xfail` markers (`edge_cases`).
- `docs/v2_phantom_promotion_trigger.md` — new "## Adversarial bench fixture" section with the 2-tier improvement roadmap.
- `CHANGELOG.md` — `[Unreleased] / Added` entry.

## Difference from #673

Two A3-class (negation-prefix) fixture entries in #673 paraphrased content from `~/.claude/CLAUDE.md` (project workflow / staging-gate vocabulary), tripping the pre-push hook and the locked discretion rule. This PR replaces both with synthetic CS-domain scenarios that exercise the same attack class (long phrase with a single `not` / `do not` negation prefix, target Jaccard ~0.93). Counts and structure are identical (101 cases, 26 strict, 75 xfail).

## Sequencing

Follow-up to #616. Test class is `@pytest.mark.skipif`-gated on the `find_phantom_lock_matches` import — green on `main` today (everything skips), executes 26 strict + 75 xfail once Surface B is on main.

## What's deliberately NOT here

- **Tier 1 cheap fix** (acronym table + Porter stemming + NFKC + hyphen split + punctuation strip) — separate PR after #616 lands. Empirical estimate: closes ~22 of the 75 xfails.
- **Tier 2 layered-matcher redesign** — separate design memo, closes ~5-10 additional cases beyond Tier 1.
- **Documented hard limits** (~25 cases — synonyms, antonyms, paraphrase, presupposition, scope, role reversal, token flooding) — beyond deterministic-narrow-surface line per #605 PHILOSOPHY.

## Verification

- `uv run --no-sync pytest tests/test_promotion_adversarial.py -q` → `1 passed, 101 skipped` on this branch (Surface B not yet on main).
- 4 atomic commits, all signed.
- Discretion grep clean against `aelf-pr-open.sh` ban list.
[...]

Issue #592, opened 2026-05-10
feat(eval-harness): wire context-rebuilder eval harness for #587 hot-start scoring


robotrocketscience:
## Status

- **Commit 1 shipped** as PR #596 (merged 2026-05-10): `replay_to_fork`, `run_rebuilder`, `measure_token_cost`, plus a non-crashing `replay_post_fork` stub that surfaces `reason="needs_replay_client"` per row.
- **Commit 2 split out as #600** to capture the host-agent subagent pattern explicitly (no SDK in aelfrice or its bench code; see that issue for the design).
- **Commit 3** (`judges/llm_judge.py`) remains in scope here.

## Background

#587 (post-compact hot-start) acceptance bullet 4 reads:

> Eval harness scores hot-start vs cold-start: agent resumes correct task on ≥80% of synthetic post-compact scenarios where the prior task was a multi-turn editor session

The implementation PR #591 ships acceptance bullets 1–3 (working-state projector + `<working-state>` sub-block emission + latency budget). Bullet 4 was not met by #591 because `benchmarks/context-rebuilder/eval_harness.py` was a skeleton with five `NotImplementedError` stubs. PR #596 wired three of them (the pure-Python ones); the remaining model-replay + judge work is split as below.

| Function | Status |
|---|---|
| `replay_to_fork()` | shipped (#596) — `MemoryStore` + `ingest_jsonl` |
| `run_rebuilder()` | shipped (#596) — `rebuild_v14()` with `floor_session=floor_l1=trigger_threshold` |
| `replay_post_fork()` | tracked in #600 — host-agent subagent replay (no SDK in aelfrice) |
| `score_fidelity()` (string-match half) | tracked in #600 |
| `score_fidelity()` (LLM-judge half) | this issue, commit-3 |
| `measure_token_cost()` | shipped (#596) — shared `estimate_tokens` (4 chars/token) |
| Modes `dynamic`, `regression` | dynamic-mode heuristic; v1.0.0 baseline capture |

Wiring these enables not only the #587 acceptance criterion but the #136 v1.4 eval harness story end-to-end.

## Remaining scope here (commit-3)

[...]

PR #702, opened 2026-05-12
feat(bench): Cohen's-κ runner for eval-judge calibration (#687)

### LINKED ISSUE #687: eval-harness: Cohen's-κ judge calibration before #592 close
## Context

#592 hot-start ≥80% AC verdict ([comment](https://github.com/robotrocketscience/aelfrice/issues/592#issuecomment-4425568029)) was produced from a single Sonnet 4.6 judge call over 18 deduplicated `(expected, actual)` pairs. `docs/BENCHMARKS.md` is explicit that the eval-judge calibration target is **Cohen's-κ disagreement against a zero-LLM baseline** — a single judge run does not establish that.

Operator disposition (2026-05-11): accept the #592 verdict numbers as-is, but gate the issue's close on κ-validation work tracked here.

## Scope

- **Multi-judge run.** Re-run the host-agent eval-replay against `benchmarks/results/592_run/` (or fresh sweep) at least 3× across independent judge invocations. Capture per-call verdicts and check inter-rater agreement.
- **Zero-LLM baseline.** Establish the deterministic baseline the κ measure compares against. If `score_substring_exact_match` (cf. `tests/test_bench_qa_scoring.py`) or another existing path applies, wire it in; otherwise spec the baseline.
- **κ verdict.** Compute Cohen's κ on the matched/unmatched binary verdicts (judge vs. baseline; and pairwise across judge runs for inter-judge agreement). Threshold for "calibrated" goes here — needs operator ratification of the bar before the run.
- **Re-confirm hot-start AC.** Once κ-validated judge produces a stable verdict, confirm hot-start fidelity ≥80% holds across the multi-run sample (not just one run).

## Acceptance

- [ ] Multi-judge run protocol documented in `docs/BENCHMARKS.md` § Eval-judge calibration.
- [ ] Zero-LLM baseline defined and runnable.
- [ ] Cohen's-κ ≥ <threshold-TBD> reported in `benchmarks/results/<run-id>/judge_kappa.json`.
- [ ] Hot-start fidelity ≥80% holds across N≥3 independent judge runs at threshold ≤0.6.
- [ ] #592 unblocked to close on completion.

## Related

- #592 — eval-harness commit-3, AC verdict pending κ-validation.
- #587 — post-compact hot-start; the AC bullet this gates.
- `benchmarks/results/592_run/` — single-run artifacts; re-use as the corpus seed if useful.
- `docs/BENCHMARKS.md` § Pass 1 / Pass 2 — contamination protocol the κ work must respect.

## Out of scope

- Replacing the rebuilder retrieval (still locked deterministic per #605).
- Importing a vendor SDK into aelfrice or its bench code (#607 lock).
- Fixture-convention rework (eval_turns indexing user vs assistant turns) — that's a separate question; if the operator picks option (b) on the #592 thread, this issue may be moot.

robotrocketscience:
Refs #687. Pairs with PR #700 (the protocol section in `docs/BENCHMARKS.md`).

## What ships

New `benchmarks/context_rebuilder/kappa.py` — the runner that produces
the `judge_kappa.json` artifact specified by the protocol section
landing in PR #700.

## Surface

- **`cohens_kappa(rater_a, rater_b)`** — pure-stdlib κ on two binary
  vectors. Handles the `p_e = 1` edge case (perfect-agreement raters
  on a single class) by returning `1.0` directly instead of `0/0`.
- **`compute_kappa_report(runs, baseline, *, run_id, ...)`** — aligns
  N judge runs and optional baseline to the turn_idx intersection
  (any turn missing from any rater is dropped), then emits pairwise
  inter-judge κ + judge-vs-baseline κ + per-run hot-start fidelity.
- **`KappaReport` → `report_to_json()`** — projects to the documented
  artifact shape: `run_id`, `n_runs`, `n_pairs`, `judge_model`,
  `baseline`, `inter_judge_kappa.{<run_i>_vs_<run_j>, mean, min}`,
  `judge_vs_baseline_kappa.{<run_i>, mean}`,
  `per_run_hot_start_fidelity[]`, `hot_start_fidelity_mean`,
  `calibrated`, `failure_reasons[]`.
- **`load_judge_run(path)`** — reads the same `judge_responses.jsonl`
  shape that `llm_judge.read_judge_responses` writes; malformed lines
  skipped (matches existing convention).
- **CLI**: `python -m benchmarks.context_rebuilder.kappa --runs A.jsonl
  B.jsonl C.jsonl --baseline base.jsonl --out judge_kappa.json
  --run-id 687_run`. Exit 0 when `calibrated == True`, exit 1
  otherwise — operators wire this into CI gates by exit code.

## Calibration gate

`calibrated == True` requires **all three**:

1. `inter_judge_kappa.min ≥ 0.70` — the **min** across all run-pairs,
   not the mean. A single noisy run-pair should not be averaged out.
2. `hot_start_fidelity_mean ≥ 0.80` — per #592 AC.
3. `n_runs ≥ 3` — minimum sample for inter-rater κ to be meaningful.

`judge_vs_baseline_kappa` is reported but does **not** participate in
the gate (report-only — see PR #700 §Eval-judge calibration rationale).

## What this does NOT do

- Doesn't run the actual κ ratification — operator-driven (acceptance
  bullets 3/4 of #687). Operator runs the dispatch step N≥3 times
  against the #592 hot-start corpus, runs `score_substring_exact_match`
  binarised against the same pairs as the baseline file, then invokes
  the CLI here.
- Doesn't close #687 — only completes acceptance bullet 2 (zero-LLM
  baseline defined and runnable). Bullets 3 (κ verdict at ratified
  threshold) and 4 (hot-start ≥80% across N≥3) are operator-side.

## Verification

- 3 atomic signed commits (module / tests / CHANGELOG).
- 16 dedicated κ tests pass — math edges (perfect agreement returns 1.0,
  perfect disagreement returns -1.0, length mismatch raises, empty
  vectors return 0.0, known κ=0.4 hand-computed fixture), report
  construction (calibrated path, rejects below n_runs / κ / fidelity
  thresholds, turn_idx intersection alignment, judge-vs-baseline
  reporting, JSON shape contract), file-format compatibility
  (skips malformed JSONL lines), and CLI smoke (calibrated → exit 0,
  n_runs<3 → exit 1).
- Full pytest (excl. bench_gate): **3610 passed, 26 skipped, 75 xfailed**
  in 76s. The xfailed bucket is the new #686 adversarial fixture; no
  regressions introduced.
- Discretion grep on diff: clean. Placeholder `<judge-model>` used as
  default model id consistent with PR #700.
- Pure stdlib — no `numpy` / `scipy` dependency added.

## Sequencing with PR #700

PR #700 documents the protocol; this PR ships the runner that implements
it. Both reference #687. They can land in either order — neither
imports the other. The CHANGELOG entries are independent bullets.


robotrocketscience:
[claim:review:faraday:2026-05-12T00:05:26Z]

robotrocketscience (commented):
LGTM (commenting; cannot formal-approve own-account PRs).

3 commits, all signed (G), CI green, discretion clean. This is the v3 critical-path unblock: completes #687 AC bullet 2, unblocks the operator-driven κ ratification that gates #592.

**Math verification:**
- p_o = (TP+TN)/n, p_yes_a = (TP+FP)/n, p_yes_b = (TP+FN)/n, p_e = P(yes)_a·P(yes)_b + P(no)_a·P(no)_b, κ = (p_o − p_e)/(1 − p_e) — textbook Cohen's-κ for 2 raters / 2 classes, correctly implemented.
- Perfect-agreement short-circuit returns 1.0 before the p_e=1 division ever runs — correct convention, prevents 0/0.
- Length mismatch raises `ValueError`; empty input returns 0.0; the post-short-circuit `p_e >= 1.0` guard is belt-and-suspenders against floating-point oddities.

**Gate design:**
- `min(pair_kappas)` not `mean(pair_kappas)` for the calibration cutoff — right call. A single noisy run-pair shouldn't be averaged away when the whole point of N≥3 runs is to surface inter-rater instability.
- `judge_vs_baseline_kappa` reported but excluded from the calibrated gate — also correct. The judge exists to catch semantic-match cases substring misses, so high judge↔baseline agreement is the failure mode (judge isn't earning its API cost), low is expected.
- `MIN_RUNS = 3` codified as a constant rather than scattered magic number.

**Surface:**
- Pure stdlib — no `numpy` / `scipy` added. Aligns with the locked "avoid embeddings + non-determinism in retrieval" stance broadly, and keeps the bench harness stdlib-only.
- CLI exit 0/1 on `calibrated` boolean — clean CI-gate primitive.
- JSON shape matches the `docs/BENCHMARKS.md §Eval-judge calibration` artifact contract (sequencing with PR #700 is independent — neither imports the other).
- `_aligned_vectors` uses `set.intersection` over turn_idx then sorted iteration — order-stable alignment, drops missing turns from any rater per κ's same-item-set requirement.

**Nits (non-blocking, JSON contract questions for downstream):**
- `inter_judge_kappa` and `judge_vs_baseline_kappa` dicts mix per-pair / per-run keys with `mean` / `min` aggregate keys. Downstream tooling iterating `inter_judge_kappa.items()` would have to filter the aggregates. Not a problem if the schema is documented (it is, in PR #700) — flagging only for the CI-gate consumer to know it should index by name, not iterate.
- `KappaReport` is `@dataclass` (mutable) but the runner builds it then mutates fields in `compute_kappa_report`. Could be a builder pattern, but the inline mutation is local and tests pin the final shape — fine as-is.

Tagging `ready-to-merge`. This unblocks the operator-side κ run that gates #592 closure.

robotrocketscience:
[release:review:faraday:2026-05-12T00:06:25Z]

Issue #696, opened 2026-05-11
feat(doctor): HRR persist/build reporter rows (#553 sub-task)


robotrocketscience:
Sub-task of #553 (HRR persistence integration). Carved out of PR #693
"Out of scope" list.

## Summary

`aelf doctor` already reports per-subsystem health rows. After PR #693
landed HRR persistence wiring, the spec at
`docs/feature-hrr-integration.md` §"Operator surface" calls for three new
rows that let operators see whether the persisted index is on disk,
how large it is, and how long the most recent rebuild took.

This is the operator-facing visibility for the persistence work. Without
it the operator can't tell whether warm-load is actually firing or
silently falling back to rebuild every time (e.g. due to an ephemeral
path or a permissions issue on `<store_dir>/.hrr_struct_index/`).

## Spec

Per `docs/feature-hrr-integration.md`:

`aelf doctor` gains three rows under an `hrr` section:

- `hrr.persist_enabled: true | false`
  Resolved per `HRRStructIndexCache._resolve_persist_dir()` — `True` if it
  returns a non-`None` path, else `False`. Captures the resolution of
  `store_path` set + `AELFRICE_HRR_PERSIST != "0"` + (#695) ephemeral-path
  not auto-disabled.
- `hrr.on_disk_bytes: <N>`
  `os.path.getsize(<persist_dir>/struct.npy) + os.path.getsize(<persist_dir>/meta.npz)`
  if both files exist; `0` otherwise.
- `hrr.last_build_seconds: <X>`
  Wall-clock time of the most recent `HRRStructIndex.build()` call this
  process. `None` (printed as `n/a`) if no build has fired since process
  start.

`aelf status` also gains a one-line summary:

- `hrr.persist_state: <on N bytes, last build Xs> | <off (reason)>`

Where `(reason)` is one of: `no store path`, `AELFRICE_HRR_PERSIST=0`,
`ephemeral path`.

## Acceptance

- [ ] `aelf doctor --json` includes `hrr.persist_enabled`,
      `hrr.on_disk_bytes`, `hrr.last_build_seconds` fields
- [ ] `aelf doctor` text output includes the three rows under a clearly
      labelled `HRR` block
- [ ] `aelf status` adds the `hrr.persist_state` summary line
- [ ] `HRRStructIndex.build()` captures wall-clock duration on a class /
      module-level slot read by doctor (no public-API churn)
- [ ] Tests:
[...]

PR #703, opened 2026-05-12
feat(config): [retrieval] hrr_persist TOML key (#698)

robotrocketscience:
Closes #698 (sub-task of read-only HRR persistence umbrella under #553).

## What ships

`[retrieval] hrr_persist` boolean TOML key in `.aelfrice.toml` controls whether the HRR structural index is persisted to / mmap-loaded from `<store_dir>/.hrr_struct_index/` (the substrate that PR #693 just shipped). Resolution precedence per the issue spec:

1. `AELFRICE_HRR_PERSIST` env var — `"0"`/`"false"`/`"no"`/`"off"` disables; `"1"`/`"true"`/`"yes"`/`"on"` forces on
2. `[retrieval] hrr_persist` TOML key — boolean
3. Default: `true`

## Six commits

1. `feat(retrieval): is_hrr_persist_enabled resolver with env→TOML precedence (#698)` — new `is_hrr_persist_enabled()` mirrors the existing `is_hrr_structural_enabled()` template at `retrieval.py:794`, plus `_env_hrr_persist_override()` helper and `ENV_HRR_PERSIST` / `HRR_PERSIST_FLAG` constants.
2. `feat(hrr): HRRStructIndexCache persist_enabled construction field (#698)` — adds `persist_enabled: bool | None = None` dataclass field to the cache; `_resolve_persist_dir()` consults env first (highest), then this field, then default-true.
3. `feat(retrieval): wire is_hrr_persist_enabled into HRRStructIndexCache construction (#698)` — adds `make_hrr_struct_cache()` factory in `retrieval.py` that resolves the flag and passes it into the cache. See "Spec ambiguity" below for why a factory rather than mutating an existing call site.
4. `test(hrr): TOML/env precedence coverage for hrr_persist (#698)` — 10 new tests covering each precedence path, both directions of env override, the cache field directly, the factory, and non-boolean TOML values.
5. `docs(changelog): unreleased entry for #698 hrr_persist TOML key`.
6. `fix(hrr): store seed as uint64 in persisted meta to match _seed_from_path range` — see "Pre-existing bug surfaced" below.

## Spec ambiguity resolved

The issue body says *"refactor so the cache takes an optional `persist_enabled: bool | None` field at construction time, populated by the config loader from the TOML key … the cache construction site in `retrieval.py`."* — but `retrieval.py` had no construction site for `HRRStructIndexCache` to refactor; existing callers all construct it themselves and pass it in.

[...]

PR #704, opened 2026-05-12
feat(doctor): HRR persist/build reporter rows (#696)

robotrocketscience:
Closes 696 (sub-task of 553).

## Summary

Adds operator-facing visibility for the HRR persistence work shipped by PR 693. `aelf doctor` gains an HRR block reporting three rows; `aelf status` gains a one-line summary. Build-duration capture is added to `HRRStructIndex.build()`; on-disk byte accounting and a read-only persist-state probe are added to the cache.

Third of the 553 sub-tasks (after 693 and 701).

### `src/aelfrice/hrr_index.py`

- `_HRR_BUILD_STATS: dict[str, float | None]` module-level slot; `None` until the first `build()` completes.
- `last_build_seconds() -> float | None` and `persist_disk_bytes(persist_dir) -> int` module helpers. The latter catches `FileNotFoundError` / `OSError` and maps to `0` so doctor never errors on a half-populated persist dir.
- `HRRStructIndex.build()` wraps wall-clock timing on both the empty-store early return path and the normal build path.
- `HRRStructIndexCache._resolve_persist_dir()` grows a `log_on_ephemeral: bool = True` keyword-only parameter. No-op today; slot for PR 701 to suppress the one-shot ephemeral WARNING when called from doctor.
- New `HRRStructIndexCache.resolve_persist_state() -> dict` returns `{enabled, dir, on_disk_bytes, reason}`. Read-only — does not mutate cache or fire any log. Reason is one of `None` (enabled), `"no store path"`, `"AELFRICE_HRR_PERSIST=0"`, `"ephemeral path"`. The ephemeral check mirrors 701's predicate but is evaluated independently so doctor reports correctly before 701 merges.

### `src/aelfrice/doctor.py`

- `DoctorReport` gains an `hrr_persist_state: dict` field.
- `diagnose(...)` accepts `hrr_store_path: str | None` and `hrr_dim: int | None` and runs `_diagnose_hrr_persist()` when both are provided. Lazy import of `hrr_index` keeps doctor cheap when HRR isn't in use.
- `_format_hrr_section()` renders the HRR block under existing per-subsystem rows:
  ```
  HRR
    persist_enabled:      true | false
    on_disk_bytes:        <N>
    last_build_seconds:   <X> | n/a
  ```

### `src/aelfrice/cli.py`

- `_cmd_doctor` now passes `hrr_store_path=str(db_path())` and the default `dim` into `diagnose()`.
- `_format_hrr_persist_status_line()` renders the `aelf status` one-liner:
  ```
  hrr.persist_state: on <N> bytes, last build <X>s
  hrr.persist_state: off (no store path | AELFRICE_HRR_PERSIST=0 | ephemeral path)
  ```
  "last build" suffix is skipped when no build has fired this process.

## Test plan

9 new tests in `tests/test_hrr_struct_index.py`:
- [x] `last_build_seconds()` returns `None` before any build
- [x] `last_build_seconds()` returns a positive float after `cache.get()`
- [x] `persist_disk_bytes(None)` and missing-files paths return `0`
- [x] `persist_disk_bytes(dir)` after build equals sum of `struct.npy` + `meta.npz` sizes
- [x] `resolve_persist_state` with `store_path=None` → `enabled=False, reason="no store path"`
- [x] `resolve_persist_state` with `AELFRICE_HRR_PERSIST=0` → `enabled=False, reason="AELFRICE_HRR_PERSIST=0"`
- [x] `resolve_persist_state` with normal path → `enabled=True, reason=None, on_disk_bytes>=0`
- [x] `resolve_persist_state` does NOT trigger the ephemeral one-shot log (read-only contract)
- [x] Doctor integration: `diagnose()` populates `hrr_persist_state` and `_format_hrr_section` renders the 3 rows

Pytest: 71 pass / 2 skip across `tests/test_hrr_struct_index.py` and `tests/test_doctor.py`.

## Out of scope

The remaining 553 follow-ups (TOML key, cold-start bench gate, docs bundle) are filed as separate sub-issues. See the 553 umbrella for the cross-reference list.

## macOS-vs-Linux note

On macOS, `Path("/tmp/x").resolve()` returns `/private/tmp/x` (symlink). `resolve_persist_state` checks raw `Path(self.store_path).parent` with `startswith`, so it WILL report `"ephemeral path"` for a `/tmp/...` macOS store. PR 701's resolver, by contrast, calls `.resolve(strict=False)` first and so will NOT auto-disable that same `/tmp/...` store on macOS. This means on macOS dev hosts the doctor row may say "off (ephemeral path)" while the cache actually persists. Acceptable for now — production target is Linux containers; once 701 lands, the standalone check here can be collapsed to call the resolver.

### DIFF
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -1860,7 +1860,37 @@ def _cmd_stats(args: argparse.Namespace, out: object) -> int:
     print(f"threads:           {n_threads}", file=out)  # type: ignore[arg-type]
     print(f"locked:            {n_locked}", file=out)  # type: ignore[arg-type]
     print(f"feedback events:   {n_history}", file=out)  # type: ignore[arg-type]
+    print(_format_hrr_persist_status_line(), file=out)  # type: ignore[arg-type]
     return 0
+
+
+def _format_hrr_persist_status_line() -> str:
+    """Return the ``hrr.persist_state`` summary line for ``aelf status`` (#696)."""
+    from aelfrice.hrr_index import (  # noqa: PLC0415
+        HRRStructIndexCache,
+        last_build_seconds as _last_build_seconds,
+    )
+    from aelfrice.store import MemoryStore as _MemoryStore  # noqa: PLC0415
+    try:
+        _store = _MemoryStore(":memory:")
+        try:
+            cache = HRRStructIndexCache(
+                store=_store, dim=512, store_path=str(db_path()),
+            )
+            state = cache.resolve_persist_state()
+        finally:
+            _store.close()
+    except Exception:  # noqa: BLE001
+        return "hrr.persist_state: unknown (probe error)"
+    enabled = bool(state.get("enabled", False))
+    if enabled:
+        on_disk = int(state.get("on_disk_bytes", 0))  # type: ignore[arg-type]
+        lbs = _last_build_seconds()
+        suffix = f", last build {lbs:.3f}s" if lbs is not None else ""
+        return f"hrr.persist_state:  on {on_disk} bytes{suffix}"
+    else:
+        reason = state.get("reason") or "disabled"
+        return f"hrr.persist_state:  off ({reason})"
 
 
 # Academic-suite targets that are scaffolded but not yet runnable at v1.0.0.
@@ -3470,6 +3500,7 @@ def _cmd_doctor(args: argparse.Namespace, out: object) -> int:
             project_root=project_root,
             hook_failures_log=hook_failures_log,
             known_cli_subcommands=known_subs,
+            hrr_store_path=str(db_path()),
         )
         print(format_report(report), file=out)  # type: ignore[arg-type]
         _print_doctor_store_check(out)
--- a/src/aelfrice/doctor.py
+++ b/src/aelfrice/doctor.py
@@ -850,6 +850,8 @@ def format_report(report: DoctorReport) -> str:
         # #236: render the missing-dep block too — the install-broken
         # case is exactly when settings.json may be absent.
         _format_missing_runtime_deps_section(report, lines)
+        # #696: HRR block is independent of settings.json scan.
+        _format_hrr_section(report, lines)
         return "\n".join(lines)
     for scope, path in report.scopes_scanned:
         lines.append(f"scanned {scope}: {path}")
--- a/src/aelfrice/doctor.py
+++ b/src/aelfrice/doctor.py
@@ -383,6 +383,10 @@ class DoctorReport:
     failed_migrate_dbs: list[FailedMigrateDB] = field(
         default_factory=lambda: cast(list[FailedMigrateDB], [])
     )
+    # HRR persistence state (#696). Keys: enabled (bool), dir (Path|None),
+    # on_disk_bytes (int), reason (str|None), last_build_seconds (float|None).
+    # None when the HRR persist probe was not requested (no store_path).
+    hrr_persist_state: dict[str, object] | None = None
 
     @property
     def broken(self) -> list[CommandFinding]:
@@ -416,6 +420,8 @@ def diagnose(
     search_tool_telemetry_path: Path | None = None,
     user_prompt_submit_telemetry_path: Path | None = None,
     aelfrice_projects_dir: Path | None = None,
+    hrr_store_path: str | None = None,
+    hrr_dim: int = 512,
 ) -> DoctorReport:
     """Walk user and project settings.json, return a DoctorReport.
 
@@ -433,6 +439,9 @@ def diagnose(
     `user_prompt_submit_telemetry_path` (#218 AC4).
     `aelfrice_projects_dir` overrides the default scan root for
     per-project DBs (`~/.aelfrice/projects`); useful in tests (#589).
+    `hrr_store_path` enables the HRR persist-state block (#696) — pass
+    the resolved DB path so doctor can probe `_resolve_persist_dir`.
+    `hrr_dim` is the HRR dimension (default 512, matches DEFAULT_DIM).
     """
     user_path = user_settings if user_settings is not None else USER_SETTINGS_PATH
     project_path = (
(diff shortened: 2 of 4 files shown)


robotrocketscience:
[claim:review:noether:2026-05-12T00:18:36Z]

robotrocketscience:
## Verdict: BLOCKED — single self-inflicted test failure

Three atomic commits, all signed (G), discretion clean, FF on `github/main`. The doctor reporter rows match the spec text exactly (`persist_enabled`, `on_disk_bytes`, `last_build_seconds`); the `aelf status` one-liner format mirrors existing per-subsystem rows; `persist_disk_bytes()` correctly maps `FileNotFoundError`/`OSError` → `0` so doctor never errors on a half-populated dir; `log_on_ephemeral=True` keyword on `_resolve_persist_dir` is the right slot for #701 to wire up later. Lazy import of `hrr_index` in doctor is the right cost call.

**Real blocker — `pytest (3.13)` failure:**

```
FAILED tests/test_hrr_struct_index.py::test_resolve_persist_state_enabled - assert False is True
```

The test calls `resolve_persist_state()` on a cache whose `store_path` lives under pytest's `tmp_path` fixture — i.e., `/tmp/pytest-of-runner/...`. The PR's own ephemeral check (mirrored from #701's predicate, lines added to `hrr_index.py` in this PR) correctly returns `enabled=False` for that path. The test asserts `enabled is True`. Self-inflicted contradiction: the test ground-truth doesn't match the PR's own logic.

**Fix:** add `monkeypatch.setenv("AELFRICE_HRR_PERSIST", "1")` at the top of `test_resolve_persist_state_enabled` (per the spec's "env=1 force-on overrides ephemeral auto-disable"). Same fix as the four failures I flagged on #701 — the same /tmp ↔ ephemeral-disable interaction.

**Other (non-blocking) notes:**

1. **Coordination with #703 (#698, just opened) and #701.** Once #703 lands, `HRRStructIndexCache` will have a `persist_enabled: bool | None = None` construction field that's the second precedence rung (between env and ephemeral). The new `resolve_persist_state()` here mirrors only the env+ephemeral checks — after #703 merges, `resolve_persist_state` should also consult `persist_enabled` to report correctly when TOML forces persistence on/off on a non-ephemeral path. Trivial follow-up; not a blocker for this PR.

2. **`HRRStructIndex.build()` timing wrap.** Module-level mutable `_HRR_BUILD_STATS` dict is fine for a single-process test signal, but two concurrent builds in the same process would race the slot. The build path doesn't appear to be called concurrently anywhere on `main` today, so not a regression — flagging for future. Threading.Lock around the slot would be ~3 lines if you want to harden.

3. **Pre-existing seed overflow (already fixed in #703).** When persistence is enabled and `cache.get()` writes meta.npz, derived seeds above 2^63 raise `OverflowError` on `np.array([self.seed], dtype=np.int64)`. PR #703 ships the int64→uint64 fix. Once that lands you'll inherit the fix; the timing wrap in this PR doesn't change that interaction.

Holding `attn:review`. Releasing my review claim so the author can fix the one assertion.

[release:review:noether:2026-05-12T00:20:02Z]

robotrocketscience:
[release:review:noether:2026-05-12T00:20:04Z]

robotrocketscience:
[claim:review:noether:2026-05-12T00:24:40Z]

robotrocketscience:
**Request changes — pytest (3.13) red on Linux runner.**

`tests/test_hrr_struct_index.py::test_resolve_persist_state_enabled` fails:

```
assert state["enabled"] is True  →  assert False is True
```

Root cause: `resolve_persist_state` (`src/aelfrice/hrr_index.py:503-506`) flags any `store_path` whose parent startswith `/tmp/` as ephemeral. pytest's `tmp_path` fixture on the Linux runner is exactly `/tmp/pytest-of-runner/pytest-0/test_resolve_persist_state_ena0/` → predicate fires → `enabled=False`.

The PR body's macOS-vs-Linux note caught the symlink direction (`/tmp` vs `/private/tmp`) but missed that pytest fixtures on the CI Linux box live under `/tmp` too, so the test as written can't pass on either platform with the current predicate.

Two fix paths, your call:

1. **Test side** — monkeypatch `_EPHEMERAL_PREFIXES` to exclude `/tmp/` for the enabled-path test, or `monkeypatch.setattr` to point the test at a non-`/tmp/` location. Localizes the workaround but masks the production mismatch.
2. **Predicate side** — tighten the ephemeral check. Either drop `/tmp/` from the list (the actual ephemeral roots that matter for HRR persist on production Linux are `/dev/shm/` and `/run/`), or check for a tmpfs mount-type rather than a path prefix. Keeps the predicate in sync with what "ephemeral" actually means in production.

Path 2 is more honest about what the doctor row should report — a developer running `aelf doctor` from `/tmp/myproj` is doing something weird but the persist *will* survive a process restart there. Either way, also worth verifying #701's resolver against the same tmpfs question before they collide.

CI status besides pytest:
- pytest (3.12) cancelled (job-graph cascade from 3.13 failure, not a real failure)
- staging-gate, codeql, calibration, deptry, vulture, codeql, sourcery, coderabbit — all green
- size-check / e2e / merge-train — skipped (depend on pytest)

Holding `attn:review` until 3.13 is green. Releasing review claim.

robotrocketscience:
[release:review:noether:2026-05-12T00:26:13Z]

robotrocketscience:
[claim:review:faraday:2026-05-12T00:34:16Z]

robotrocketscience:
[claim:review:pascal:2026-05-12T00:34:33Z]

robotrocketscience:
[release:review:pascal:2026-05-12T00:34:38Z]

robotrocketscience:
## Request changes — pytest 3.13 fails on `test_resolve_persist_state_enabled`

`tests/test_hrr_struct_index.py::test_resolve_persist_state_enabled` builds a cache at `tmp_path/<sp>` under `/tmp/pytest-of-runner/...` and asserts `state["enabled"] is True`. The standalone ephemeral predicate this PR adds inside `resolve_persist_state` correctly flags `/tmp/...` as ephemeral and returns `enabled=False, reason="ephemeral path"`. The test's docstring ("Normal path with a non-ephemeral store") tells the intent, but `tmp_path` doesn't satisfy it on Linux CI.

### Fix options

1. **Monkeypatch the prefix set** for that one test:
   ```python
   monkeypatch.setattr(hrr_index, "_EPHEMERAL_PATH_PREFIXES", frozenset())
   ```
   Mirrors the macOS-shim pattern this PR already uses.
2. **Force-on via env var** in the test:
   ```python
   monkeypatch.setenv("AELFRICE_HRR_PERSIST", "1")
   ```
   Cleaner if you'd rather not import `_EPHEMERAL_PATH_PREFIXES`.

Either re-establishes the "non-ephemeral store" precondition the test name promises.

### Cross-PR note

PR #701 has the same shape of regression on 4 pre-existing persistence tests (see my comment there). The fix here is independent — both can be applied without rebasing on each other — but you may want to coordinate the test-file edits so they don't collide on rebase.

Otherwise the diff looks solid: the `log_on_ephemeral` kwarg slot for #701 is forward-compatible, the read-only contract of `resolve_persist_state` (doesn't fire the one-shot log) is good, and the macOS/Linux divergence note in the body is accurate. Holding claim release — re-ping `attn:review` once 3.13 + 3.12 pytest are green.

robotrocketscience:
[release:review:faraday:2026-05-12T00:34:57Z]

robotrocketscience:
[claim:review:pascal:2026-05-12T01:03:48Z]

robotrocketscience:
[release:review:pascal:2026-05-12T01:19:25Z]

robotrocketscience:
Rebased onto current main (PR #701 just merged). Conflict resolution: `_resolve_persist_dir` gains `log_on_ephemeral=True` parameter while keeping main's `_ephemeral_disable_logged` one-shot guard; redundant `_EPHEMERAL_PREFIXES` stop-gap dropped in favor of main's `_EPHEMERAL_PATH_PREFIXES`; #695 and #696 test groups concatenated. Full pytest green locally (3695 passed, 59 skipped, 75 xfailed). `ready-to-merge` re-applied; merge-train should pick it up once CI re-runs.

robotrocketscience:
Follow-up fix pushed for the CI `pytest (3.13)` failure: `test_hrr_persist_disk_bytes_after_build` was asserting against `os.path.getsize(pd / "struct.npy")` but on Linux CI `tmp_path` resolves under `/tmp/...` which trips the #695 ephemeral-path auto-disable, so the file never gets written. Monkeypatched `_EPHEMERAL_PATH_PREFIXES` to empty for this test (matching the pattern used by sibling disk-write tests). `ready-to-merge` re-applied.

PR #705, opened 2026-05-12
fix(bench): hot_start fixture meta — eval_turns at canonical answers (#687)

robotrocketscience:
Refs #687. Refs #592.

## What ships

Patches `hot_start_debugging_session_001.meta.json`: `eval_turns: [8, 10, 12] → [9, 11, 13]`. Plus the matching test fixture in `tests/test_context_rebuilder_eval_harness_wiring.py`.

## Why

PR #702 (κ runner) shipped the calibration mechanics; running it against the existing fixture exposed the known #592 fixture quirk:

> *Fixture quirk surfaced: eval_turns index user turns (so expected=user_question_text), substring path always misses, judge does all the lifting.*

A 3-judge sweep against the old fixture produced **inter-judge κ = 0.0** because judges split 2-vs-1 on how to interpret "matched" when REFERENCE was a user question and CANDIDATE was its answer — one camp scored on factual content delivery, the other on paraphrase equivalence.

After this patch, eval_turns point at the assistant continuations (9, 11, 13) instead of the user prompts (8, 10, 12). The harness already does the right thing for `user_turn_by_idx` (it tracks "most-recent user-role text at-or-before line i"), so user_turn is still the prompt and `expected` becomes the canonical answer.

## Evidence

3-judge κ sweep against the new fixture:

```json
{
  "inter_judge_kappa": {"min": 1.0, "mean": 1.0, ...},
  "judge_vs_baseline_kappa": {"mean": 0.0},
  "per_run_hot_start_fidelity": [0.667, 0.667, 0.667],
  "hot_start_fidelity_mean": 0.667,
  "calibrated": false,
  "failure_reasons": ["hot_start_fidelity_mean=0.6667 < threshold 0.8"]
}
```

The κ gate **passes** (1.0 across all pairs). The fidelity gate (0.67 < 0.80) is now a real product-side question about rebuild block adequacy, not eval ambiguity — judges agreed that one of three replay responses omitted a load-bearing detail (the failure-mode subclause for turn 9).

## What this does NOT do

- Doesn't claim #592 close — that needs hot-start fidelity ≥ 0.80, which is now an honest open question about the rebuild block content (separate work).
- Doesn't touch sibling `debugging_session_001.meta.json` — same quirk applies but cold-start ratification is its own sweep.
- Doesn't bump the judge prompt — calibration now works without prompt changes.

## Verification

- 2 atomic signed commits.
- Discretion grep clean on diff.
- `tests/test_context_rebuilder_eval_harness_wiring.py` — 22 passed.
- Sister κ-runner branch (PR #702) verified the calibration end-to-end against the new fixture: inter_judge_kappa.min = 1.0.


robotrocketscience:
[claim:review:noether:2026-05-12T00:21:40Z]

robotrocketscience:
## Verdict: READY-TO-MERGE

Two atomic signed commits (G), 6/3 line diff, FF on `github/main`, all 16 CI checks green, discretion grep clean.

**Logic:** The fixture-meta patch is the right shape. The harness's documented convention (per the existing `debugging_session_001.meta.json` cold-start fixture) is that `eval_turns` index *assistant* turns where `expected` becomes the assistant continuation, with `user_turn` pulled from the most-recent-user-role text at-or-before each eval index. Shifting `[8, 10, 12] → [9, 11, 13]` puts the hot-start fixture in the same convention: `expected` is now the canonical answer the post-rebuild agent should produce, judge prompt no longer mixes Q/A speech acts, κ=0.0 → κ=1.0 in the 3-judge sweep cited in the PR body.

**Honest scope:**
- The author explicitly does not claim #592 close — `hot_start_fidelity_mean = 0.667 < 0.80` is now a real rebuild-block-content question rather than a fixture artifact, which is the right thing to surface separately.
- Sibling `debugging_session_001.meta.json` (cold-start) stays unchanged with its `[8, 10, 12, 14]` shape pending its own re-ratification, correct per the spec's separate-bench-per-task-type convention.
- The expanded `notes` field documents both the new shape and the prior failure mode for future maintainers, which is good fixture hygiene.

**Coordination:**
- The κ runner referenced (PR #702) and this fixture patch land cleanly together — neither rebases the other; the runner reads the meta file, this PR fixes the meta.
- Once this lands, the κ gate (inter_judge ≥ 0.70 per #702) is satisfied for the hot-start fixture; the fidelity gate becomes the next #592 conversation, separately.

Adding `ready-to-merge` label to trigger merge-train.

[release:review:noether:2026-05-12T00:22:37Z]

robotrocketscience:
[release:review:noether:2026-05-12T00:22:41Z]

Issue #739, opened 2026-05-13
Flip default: bfs_enabled (L2 graph-walk lane) on by default — gated on v3.0 latency bench re-run


robotrocketscience:
## Summary

Flip `is_bfs_enabled()` default from False → True (`src/aelfrice/retrieval.py:1160`). The README's §"How it works" lists graph-walk as one of four parallel retrieval lanes ("L2: graph walk → typed-edge BFS from L1 seeds (SUPPORTS, CONTRADICTS, SUPERSEDES, DERIVED_FROM, ...)") — but the wheel ships with this lane OFF, so the README is currently false-advertising a load-bearing default.

**This flip is bench-gated.** The published latency band is from v1.3.0; aelfrice is at v3.0. Eight-plus minors of HRR / BM25F / heat-kernel / clustering changes have layered on top, and the latency band has not been re-run on the v3.0 stack at a v3.0-representative corpus size.

## Why this is the right intent

1. README §"How it works" describes four lanes (L0/L1/L2/L2.5). The default install fires three (L0/L1/L2.5). L2 (graph walk) is off. The product as documented and the product as installed are not the same product.
2. The original v1.3.0 default-OFF decision was a backward-compat lock ("a fresh install must not change retrieval output against the v1.2 baseline" — `docs/bfs_multihop.md:524-525`). That constraint is obsolete: aelfrice has had 8+ minors of intentional retrieval-output changes since.
3. Bench evidence at v1.3 (where the band was last measured) cleared the bar — the lane works. Question now is only whether the v3.0 stack still hits the latency budget.

## Latency context (corrected 2026-05-13)

Earlier framing in this issue body (and my session-side commentary) called BFS-on "potentially ~1s of retrieval per heavy turn." That overstated the risk. Honest cross-project comparison:

| Measurement | Agentmemory (April 2026, broad prompt, single sample) | Aelfrice (v1.3 acceptance band, labelled corpus) |
|---|---|---|
| FTS-only baseline | 0.03 s | — |
| BFS contribution alone | ~17–20 ms (0.05s − 0.03s) | **p50 ≤ 25 ms, p95 ≤ 100 ms** |
| HRR contribution alone | ~1.07 s (cold, no cache) | ~ms (warm, persisted — #553/#693/#698/#701/#703/#704/#706) |
| Combined HRR + BFS | 9.4 s (cold HRR + BFS) | ~30 ms p50, ~100 ms+ p95 (warm HRR + BFS) |

[...]

Issue #740, opened 2026-05-13
Cross-fire injection dedup: amortize search hooks against the UPS-fire memory block


robotrocketscience:
## Summary

Track which belief IDs have been injected during the current session so that subsequent `PreToolUse:Grep`/`Glob`/`Bash`-search hook fires can deduplicate against the UserPromptSubmit-fire injection set. When a `PreToolUse` retrieve() returns mostly already-injected beliefs, emit a short "answer already in prompt context" signal instead of re-injecting — the agent learns it can stop searching for things it already has.

The amortization claim: front-load the prompt with a rich retrieval block, then suppress redundant injections on subsequent search hooks. Over a turn that issues 1 prompt + 5 Grep + 3 Bash-searches, total injected-context volume drops sharply, and the agent has a clearer signal to stop searching.

## Why this matters

Today every `retrieve()` call is stateless: each PreToolUse:Grep hook fire re-runs retrieval from scratch with no awareness that the UPS-fire 200ms earlier may have already shipped the same beliefs into the prompt the agent is currently acting on. The agent gets the same beliefs twice and learns nothing useful from the second injection — strictly noise.

This problem compounds with #738 (search-tool / search-tool-bash default-flip). Once those hooks are default-on, the redundant-injection cost lands on every Grep/Bash-search call. Cross-fire dedup is the lever that makes #738 cheap.

## Current state (verified)

- `src/aelfrice/hook.py:85` defines `SESSION_STATE_FILENAME = "session_first_prompt.json"` — but this sentinel only carries a boolean (is-first-prompt?) for the `<session-start>` sub-block, not per-belief tracking.
- `src/aelfrice/hook.py:1467` `is_session_first_prompt()` — reads/writes the sentinel.
- No belief-ID tracking across hook fires anywhere in `src/aelfrice/hook_search_tool.py`, `src/aelfrice/hook.py`, or `src/aelfrice/retrieval.py`.

## Design

**New sentinel:** `<git-common-dir>/aelfrice/session_injected_ids.json`

```json
{
  "session_id": "<uuid>",
  "ring": [
    {"id": "b1", "fire_idx": 0, "ts": 1715000000},
    {"id": "b2", "fire_idx": 0, "ts": 1715000000},
    {"id": "b4", "fire_idx": 1, "ts": 1715000005}
  ],
  "ring_max": 200,
  "next_fire_idx": 2
}
```

[...]

Issue #738, opened 2026-05-13
Flip default: search-tool + search-tool-bash hooks on by default


robotrocketscience:
## Summary

Flip the install-time default for the `search-tool` (PreToolUse:Grep|Glob) and `search-tool-bash` (PreToolUse:Bash for grep/rg/find/fd/ack) hooks from **OFF → ON**. The README already advertises the four-layer retrieval stack as a defining feature, but Claude's *own* Grep/Glob/Bash-search calls don't use it unless the user has opted in. The product as documented and the product as installed are not the same product.

## Source-of-truth current state

`src/aelfrice/cli.py:2475` — `if getattr(args, "search_tool", False):` (OFF)
`src/aelfrice/cli.py:2492` — `if getattr(args, "search_tool_bash", False):` (OFF)

The argparse setup also exposes a `--search-tool` opt-in flag, not a `--no-search-tool` opt-out. The default value is what wires into `aelf setup`.

## Why this is the right flip

1. README §"What it does" describes four parallel retrieval lanes as the core feature. The default install doesn't fire them on Grep/Glob/Bash-search, so the agent's most frequent search path skips aelfrice entirely.
2. The lane reused by these hooks is the same `retrieve()` that's already validated for UserPromptSubmit fires — no new retrieval code path is enabled by the flip.
3. Per the audit in 2026-05-13's session: this default-OFF was a conservative choice at install time, not a bench-gated decision. No bench gate blocks the flip.

## Latency cost (measured)

| Surface | Cost per fire | Cap | Turn-level worst case |
|---|---:|---:|---:|
| `search-tool` (Grep, Glob) | ~4.5ms p99 (token budget 600, L1 ≤ 10 hits) | none | 5 Grep × 4.5ms ≈ **22ms** |
| `search-tool-bash` (grep/rg/find/fd/ack via Bash) | ~2–4ms (token budget 300, L1 ≤ 5 hits) | **3 fires per turn** (`BASH_FIRE_CAP_PER_TURN` in `hook_search_tool.py:75`) | 3 × 4ms ≈ **12ms** |
| Combined heavy turn (5 Grep + 3 Bash-search) | — | — | **~34ms / turn** |

Source: `hook_search_tool.py:48-75` for budgets/caps; README:47 for the 4.5ms p99 baseline (labelled query-strategy corpus, ~10k beliefs).

[...]

PR #743, opened 2026-05-13
feat(retrieval): adaptive expansion-gate for broad prompts (#741)

### LINKED ISSUE #741: Adaptive expansion-gate: skip BFS/HRR-expensive lanes on broad prompts
## Summary

Add a query-shape-aware gate that skips the expensive expansion layers (BFS, and optionally HRR-structural lookups) when the prompt is unlikely to benefit from them — broad natural-language prompts, exploratory questions, prompts that don't reference specific belief structure. Cheap deterministic features only (token count, n-gram entropy, structural-marker presence). Lifts the same idea agentmemory triage #83 settled on; aelfrice already has the cache/persistence side, but never ported the gate.

Follow-up to #739 (BFS default-flip). #739's bench should validate against this gate's existence — the BFS flip is much cheaper when broad-prompt expansion is skipped.

## Background

Agentmemory's issue #83 (April 2026) measured retrieval latency by configuration on a broad natural-language prompt:

| Config | Latency |
|---|---:|
| `use_hrr=True, use_bfs=True` | 9.4 s |
| `use_hrr=False, use_bfs=True` | 0.05 s |
| `use_hrr=True, use_bfs=False` | 1.1 s |
| FTS only | 0.03 s |

The triage plan at `~/projects/aelfrice-lab/archive/cleanup-2026-04-26/triage-docs/triage__gh-83-search-broad-prompt-slow__PLAN.md` proposed *gating* expansion on broad prompts. Agentmemory never shipped the gate; aelfrice fixed the HRR cost separately (persistence + cache) but didn't port the gate. With #739 putting BFS on the default path, the gate becomes load-bearing again — broad prompts are exactly where BFS provides the least marginal value and pays the highest cost.

## Design

**New resolver: `should_run_expansion(query)` → `ExpansionDecision`**

```python
@dataclass(frozen=True)
class ExpansionDecision:
    run_bfs: bool
    run_hrr_structural: bool   # leaves L0+L1+L2.5-entity always on
    reason: str                # for telemetry / aelf doctor
```

**Heuristics (all deterministic, stdlib-only — honors #605):**

1. **Length gate:** prompt token count > `BROAD_PROMPT_TOKEN_THRESHOLD` (default 80) → broad. Long natural-language questions correlate with low-structural-marker queries that don't benefit from graph expansion.
2. **Structural-marker presence:** prompt contains any of `#NNN` issue refs, file paths (`src/...`, `tests/...`), function names (`identifier_with_underscores` / `camelCaseLikely`), edge-type names (`SUPPORTS`, `CONTRADICTS`, ...). Presence → narrow → expansion likely useful. None → broad.
3. **Question-form gate:** prompt starts with `what`, `why`, `how`, `which`, `who`, `tell me`, `explain` → broad. Imperative/declarative prompts about specific entities → narrow.
4. **N-gram entropy gate (optional, behind flag):** rough proxy for vocabulary diversity. High entropy + no markers → broad.

Decision is the AND of the gates that fire: if ANY signal says "broad", set `run_bfs=False`. Conservative: prefer skipping expansion when in doubt; user can always opt into the full stack via env override.

**Resolver precedence (mirrors existing flag resolvers in retrieval.py):**

1. `AELFRICE_FORCE_EXPANSION=1` env override → always run expansion (escape hatch for testing/bench).
2. `AELFRICE_NO_EXPANSION_GATE=1` env override → disable the gate entirely (back to pre-gate behaviour).
3. `[retrieval] expansion_gate_enabled` in `.aelfrice.toml` → bool, default True.
4. Run gate heuristics, return `ExpansionDecision`.

**Integration into `retrieve()`:**

In `src/aelfrice/retrieval.py:1394` `retrieve()`, after `bfs_on = is_bfs_enabled(...)` and `enabled = is_entity_index_enabled(...)`, layer the gate:

```python
gate_decision = should_run_expansion(query)
bfs_on = bfs_on and gate_decision.run_bfs
# HRR-structural lane stays default-on for now; gate can opt-out via run_hrr_structural=False
```

Telemetry: emit one line to the hook audit log per gated fire so `aelf doctor` / `aelf tail` can show what got skipped and why.

## Acceptance

- [ ] `should_run_expansion(query)` returns `ExpansionDecision` with `reason` populated for every branch
- [ ] All four heuristics (length, structural-marker, question-form, optional n-gram entropy) covered by unit tests
- [ ] `retrieve()` consults the gate; when `run_bfs=False` the BFS lane short-circuits even if `is_bfs_enabled()` returns True
- [ ] `AELFRICE_FORCE_EXPANSION=1` overrides the gate (escape hatch)
- [ ] `AELFRICE_NO_EXPANSION_GATE=1` disables the gate entirely (pre-feature parity)
- [ ] TOML key `[retrieval] expansion_gate_enabled` honoured
- [ ] Telemetry row written per gated fire (visible in `aelf tail`)
- [ ] Bench: a labelled-corpus replay measures broad-prompt p50 / p95 with gate-on vs gate-off; broad-prompt p95 must improve by ≥30% vs gate-off (the load-bearing claim)
- [ ] Bench: narrow-prompt p50 must not regress (gate must not fire on prompts that need expansion)

## Bench gate

`benchmarks/results/<run-id>/expansion_gate_bench.json` with:
- 50 labelled broad prompts (architectural / exploratory / multi-topic)
- 50 labelled narrow prompts (specific identifiers / file paths / issue refs)
- Each run 4 ways: (gate-on, gate-off) × (bfs-on, bfs-off)
- Report p50 / p95 latency per cell, recall@k per cell

This bench is what #739's BFS-flip bench should layer on top of: run the same fixture with the gate to see whether the BFS p95 hits the 100ms acceptance band when broad prompts are auto-skipped.

## Risk

- **False positives.** Gate misfires on a prompt that would have benefited from expansion → user loses recall. The escape hatch (`AELFRICE_FORCE_EXPANSION=1`) and explicit slash-command form (`aelf reason` keeps full expansion regardless) mitigate this. A small recall regression on broad prompts is acceptable because BFS provides little marginal value there anyway — but the labelled-corpus bench must verify this empirically, not just argue from theory.
- **Heuristic drift.** The four heuristics encode a prior about what \"broad\" means. New prompt shapes (e.g., agentic multi-turn refinement) may not match. Mitigation: the telemetry surface lets us inspect what fired and tune over time.
- **Interaction with #740 (cross-fire dedup).** If the gate suppresses expansion on the UPS-fire, the PreToolUse:Grep dedup gets less to dedupe against. The bench should cover the (gate, dedup, bfs-flip) interaction matrix — that's where #739 ratifies-or-not.

## Out of scope

- ML-based query classification. PHILOSOPHY (#605) is deterministic narrow surface; this stays in regex / token-count / token-set territory.
- Per-edge-type gating. The gate decides whether to run BFS at all, not which edges to follow once it does.
- Gating L1 BM25 or L2.5 entity-index. Those are cheap (FTS5 + index lookup); not worth gating.
- Cross-language tokenization. Heuristics assume English / code-mixed text; non-Latin scripts will trigger the n-gram-entropy fallback.

## References

- #739 (BFS default-flip — this gate is the precondition for that flip if the bench shows BFS p95 blows the band on broad prompts)
- #738 (search-tool default-flip — orthogonal but related; this gate also fires inside PreToolUse:Grep retrieval)
- #740 (cross-fire injection dedup — interacts via the (UPS, PreToolUse) cost model)
- Agentmemory #83 archive at `~/projects/aelfrice-lab/archive/cleanup-2026-04-26/issues-prs/github-archive/rrs_agentmemory.issues.json`
- Triage plan at `~/projects/aelfrice-lab/archive/cleanup-2026-04-26/triage-docs/triage__gh-83-search-broad-prompt-slow__PLAN.md`
- PHILOSOPHY decision (#605) — deterministic narrow surface, no embeddings; this gate honours it

robotrocketscience:
## Summary

Adds the adaptive expansion-gate from #741: a cheap deterministic
prompt-shape check that short-circuits the BFS multi-hop lane on
broad natural-language prompts while leaving L0 / L1 / L2.5-entity
always on. Sets up the precondition #739's BFS default-flip needs —
once the BFS flip ratifies, broad prompts won't blow the latency
band because the gate suppresses BFS on them.

Closes #741.

## What landed

- `src/aelfrice/expansion_gate.py` — new module with
  `should_run_expansion(query)` → `ExpansionDecision(run_bfs,
  run_hrr_structural, reason)`.
- `src/aelfrice/retrieval.py` — `retrieve()` and
  `retrieve_with_tiers()` consult the gate after
  `is_bfs_enabled()`; `bfs_on = bfs_on and gate_decision.run_bfs`.
  Two additive fields on `LaneTelemetry`
  (`expansion_gate_reason`, `expansion_gate_skipped_bfs`) for
  downstream surfaces.
- `src/aelfrice/hook.py` — `_write_hook_audit_record` accepts the
  two new fields; UPS callsite reads `last_lane_telemetry()` post-
  retrieve and threads them into `hook_audit.jsonl` (visible via
  `aelf tail`).
- `tests/test_expansion_gate.py` — 27 tests covering heuristics,
  resolver precedence, TOML fail-soft, retrieve() integration,
  LaneTelemetry contract.
- `tests/test_bfs_multihop.py` — BFS-internal tests set
  `AELFRICE_FORCE_EXPANSION=1` in their `isolated_env` fixture so
  the gate doesn't interfere with assertions about BFS behaviour.

## Heuristics

Deterministic, stdlib-only (honors PHILOSOPHY #605):

1. **Length** — prompt > `BROAD_PROMPT_TOKEN_THRESHOLD` (default 80
   tokens) → broad.
2. **Structural-marker absence** — no `#NNN`, no `src/...`, no
   `tests/...`, no snake_case / camelCase identifier, no edge-type
   name (`SUPPORTS`, `CONTRADICTS`, …) → broad.
3. **Question-form prefix** — starts with `what`, `why`, `how`,
   `which`, `who`, `tell me`, `explain` → broad.

Any "broad" signal → `run_bfs=False`. Conservative by design;
escape hatches via env override or `aelf reason`.

## Resolver precedence

`AELFRICE_FORCE_EXPANSION=1` > `AELFRICE_NO_EXPANSION_GATE=1` >
`[retrieval] expansion_gate_enabled` in `.aelfrice.toml` (default
True) > heuristics.

Malformed TOML fails soft.

## Bench (deferred)

Per the issue body, the load-bearing bench (labelled-corpus
broad/narrow split with p95 improvement target ≥30%) is layered on
top of #739's BFS-flip bench re-run. This PR ships the code +
unit/integration tests; the bench is wired separately when #739's
gate ratification runs. The code path is observable via
`aelf doctor` (LaneTelemetry) and `aelf tail` (hook_audit.jsonl)
in the meantime.

## Test plan

- [x] `uv run pytest tests/test_expansion_gate.py` — 27/27 pass
- [x] `uv run pytest tests/test_bfs_multihop.py tests/test_retrieve_v2.py`
      — 39/39 pass (BFS-internal contracts unchanged with
      `AELFRICE_FORCE_EXPANSION=1` fixture)
- [x] Full unit suite (excluding unrelated pre-existing fastmcp
      NameError) — 3775 pass / 57 skipped / 75 xfail
- [ ] #739 broad/narrow bench (deferred; gated on BFS-flip
      ratification per issue body)

## Risk

- **Heuristic drift.** False positives on prompts that would have
  benefited from expansion. Mitigated by env escape hatch +
  telemetry surface (`aelf tail` shows what got gated and why).
- **API surface.** `LaneTelemetry` gains two additive fields with
  safe defaults; pre-#741 callers keep working.
- **Existing BFS suites.** Fixed inline by setting
  `AELFRICE_FORCE_EXPANSION=1` in the affected test module's
  isolated-env fixture so the gate doesn't shadow what those tests
  are actually asserting.

### DIFF
--- a/src/aelfrice/expansion_gate.py
+++ b/src/aelfrice/expansion_gate.py
@@ -1,0 +1,300 @@
+"""Adaptive expansion-gate for retrieval (#741).
+
+Cheap deterministic prompt-shape gate that decides whether to run the
+expensive expansion layers (BFS multi-hop, HRR structural) for a given
+query. Broad natural-language prompts that lack structural markers are
+unlikely to benefit from graph expansion and pay the highest latency
+cost on those lanes — the gate short-circuits expansion on them while
+keeping L0 / L1 / L2.5-entity always on.
+
+Honors v3.0 PHILOSOPHY (#605): deterministic narrow surface, stdlib
+only, no embeddings, no model calls.
+
+Resolver precedence (first decisive wins):
+  1. ``AELFRICE_FORCE_EXPANSION=1`` env var → always run expansion
+     (escape hatch for testing and bench).
+  2. ``AELFRICE_NO_EXPANSION_GATE=1`` env var → disable the gate
+     entirely; behaves as pre-gate (all expansion lanes run).
+  3. ``[retrieval] expansion_gate_enabled`` in ``.aelfrice.toml``.
+     Default ``True``.
+  4. Run heuristics; return decision.
+
+Heuristics (all deterministic, stdlib-only):
+  - Length: prompt token count > ``BROAD_PROMPT_TOKEN_THRESHOLD``
+    (default 80) → broad signal.
+  - Structural-marker absence: no ``#NNN`` issue refs, no file paths
+    (``src/...``, ``tests/...``, ``docs/...``, ``benchmarks/...``),
+    no snake_case / camelCase identifiers, no edge-type names
+    (``SUPPORTS``, ``CONTRADICTS``, ...) → broad signal.
+  - Question-form prefix: starts with ``what``, ``why``, ``how``,
+    ``which``, ``who``, ``tell me``, ``explain`` → broad signal.
+
+If ANY signal fires "broad", expansion is skipped. Conservative by
+design: prefer skipping when in doubt; users with a narrow query that
+the heuristic misclassifies can opt back into the full stack via
+``AELFRICE_FORCE_EXPANSION=1`` or ``aelf reason`` (which never gates).
+"""
+
+from __future__ import annotations
+
+import os
+import re
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Final
+
+from aelfrice.models import EDGE_TYPES
+
+# --- Config keys ----------------------------------------------------------
+
+CONFIG_FILENAME: Final[str] = ".aelfrice.toml"
+RETRIEVAL_SECTION: Final[str] = "retrieval"
+EXPANSION_GATE_FLAG: Final[str] = "expansion_gate_enabled"
+
+ENV_FORCE_EXPANSION: Final[str] = "AELFRICE_FORCE_EXPANSION"
+ENV_NO_EXPANSION_GATE: Final[str] = "AELFRICE_NO_EXPANSION_GATE"
+
+_ENV_TRUTHY: Final[frozenset[str]] = frozenset({"1", "true", "yes", "on"})
+_ENV_FALSY: Final[frozenset[str]] = frozenset({"0", "false", "no", "off"})
+
+# --- Heuristic thresholds -------------------------------------------------
+
+BROAD_PROMPT_TOKEN_THRESHOLD: Final[int] = 80
+
+_QUESTION_FORM_PREFIXES: Final[tuple[str, ...]] = (
+    "what",
+    "why",
+    "how",
+    "which",
+    "who",
+    "tell me",
+    "explain",
+)
+
+# Patterns for structural-marker detection. Any match counts as "narrow"
+# evidence (the prompt references specific code / belief structure).
+_ISSUE_REF_RE: Final[re.Pattern[str]] = re.compile(r"#\d+\b")
+_FILE_PATH_RE: Final[re.Pattern[str]] = re.compile(
+    r"\b(?:src|tests|docs|benchmarks|scripts)/[A-Za-z0-9_./\-]+",
+)
+# snake_case identifiers (at least one underscore between word chars).
+_SNAKE_CASE_RE: Final[re.Pattern[str]] = re.compile(
+    r"\b[a-z][a-z0-9]*(?:_[a-z0-9]+)+\b",
+)
+# camelCase identifiers (lowercase prefix followed by an internal capital).
+_CAMEL_CASE_RE: Final[re.Pattern[str]] = re.compile(
+    r"\b[a-z][a-z0-9]*[A-Z][A-Za-z0-9]+\b",
+)
+# Edge-type names live in models.EDGE_TYPES. Compile once at import time.
+_EDGE_TYPE_RE: Final[re.Pattern[str]] = re.compile(
+    r"\b(?:" + "|".join(sorted(EDGE_TYPES, key=len, reverse=True)) + r")\b",
+)
+
+
+@dataclass(frozen=True)
+class ExpansionDecision:
+    """Per-call decision from :func:`should_run_expansion`.
+
+    Attributes
+    ----------
+    run_bfs:
+        When ``False`` the BFS multi-hop lane is skipped in
+        :func:`aelfrice.retrieval.retrieve`. L0 / L2.5-entity / L1
+        always stay on.
+    run_hrr_structural:
+        When ``False`` the HRR structural-query lane is skipped.
+        v3.0 default keeps this on regardless of gate verdict; the
+        field is reserved for forward-compat once the bench shows
+        broad-prompt HRR-structural cost is load-bearing.
+    reason:
+        Short human-readable tag used by telemetry / ``aelf doctor``
+        to explain *why* the gate fired. Always populated.
+    """
+
+    run_bfs: bool
+    run_hrr_structural: bool
+    reason: str
+
+
+def _env_force_expansion() -> bool | None:
+    """``True`` when AELFRICE_FORCE_EXPANSION is truthy, ``False`` when
+    explicitly falsy, ``None`` when unset / unrecognised.
+    """
+    raw = os.environ.get(ENV_FORCE_EXPANSION)
+    if raw is None:
+        return None
+    norm = raw.strip().lower()
+    if norm in _ENV_TRUTHY:
+        return True
+    if norm in _ENV_FALSY:
+        return False
+    return None
+
+
+def _env_no_expansion_gate() -> bool | None:
+    """``True`` when AELFRICE_NO_EXPANSION_GATE is truthy, ``False``
+    when explicitly falsy, ``None`` when unset / unrecognised.
+    """
+    raw = os.environ.get(ENV_NO_EXPANSION_GATE)
+    if raw is None:
+        return None
+    norm = raw.strip().lower()
+    if norm in _ENV_TRUTHY:
+        return True
+    if norm in _ENV_FALSY:
+        return False
+    return None
+
+
+def _read_toml_flag(start: Path | None = None) -> bool | None:
+    """Read ``[retrieval] expansion_gate_enabled`` from
+    ``.aelfrice.toml``. Returns ``None`` when the file / section /
+    key is absent or the value is not a bool. Fail-soft: malformed
+    TOML returns ``None``.
+    """
+    try:
+        import tomllib
+    except ImportError:  # Python <3.11 fallback; aelfrice ships >=3.11
+        return None
+    cur = (start or Path.cwd()).resolve()
+    seen: set[Path] = set()
+    while True:
+        candidate = cur / CONFIG_FILENAME
+        if candidate.is_file():
+            try:
+                with candidate.open("rb") as fh:
+                    data = tomllib.load(fh)
+            except (OSError, ValueError, tomllib.TOMLDecodeError):
+                return None
+            section = data.get(RETRIEVAL_SECTION)
+            if isinstance(section, dict):
+                value = section.get(EXPANSION_GATE_FLAG)
+                if isinstance(value, bool):
+                    return value
+            return None
+        if cur in seen or cur.parent == cur:
+            return None
+        seen.add(cur)
+        cur = cur.parent
+
+
+def _has_structural_markers(text: str) -> bool:
+    """Return ``True`` if ``text`` contains at least one structural
+    marker — an issue ref, file path, identifier, or edge-type name.
+    """
+    if _ISSUE_REF_RE.search(text):
+        return True
+    if _FILE_PATH_RE.search(text):
+        return True
+    if _SNAKE_CASE_RE.search(text):
+        return True
+    if _CAMEL_CASE_RE.search(text):
+        return True
+    if _EDGE_TYPE_RE.search(text):
+        return True
+    return False
+
+
+def _starts_with_question_form(text: str) -> bool:
+    """Return ``True`` if ``text`` begins with a recognised
+    question-form prefix (case-insensitive).
+    """
+    lowered = text.lstrip().lower()
+    for prefix in _QUESTION_FORM_PREFIXES:
+        if lowered.startswith(prefix):
+            # Require a word boundary so ``whatever`` doesn't match
+            # ``what``. End-of-string or any non-alnum char qualifies.
+            tail_pos = len(prefix)
+            if tail_pos >= len(lowered):
+                return True
+            tail = lowered[tail_pos]
+            if not tail.isalnum() and tail != "_":
+                return True
+    return False
+
+
+def should_run_expansion(
+    query: str,
+    *,
+    start: Path | None = None,
+) -> ExpansionDecision:
+    """Decide whether to run BFS / HRR-structural expansion for ``query``.
+
+    Resolver precedence: env-force > env-disable-gate > TOML > heuristics.
+
+    Parameters
+    ----------
+    query:
+        The user prompt or retrieval query. Whitespace-only / empty
+        queries short-circuit to ``run_bfs=True`` (no gating; the
+        upstream retrieve() returns L0-only on empty input anyway).
+    start:
+        Optional directory to start the ``.aelfrice.toml`` walk from.
+        Defaults to ``Path.cwd()``.
+
+    Returns
+    -------
+    ExpansionDecision
+        Populated decision with ``reason`` set for telemetry.
+    """
+    env_force = _env_force_expansion()
+    if env_force is True:
+        return ExpansionDecision(
+            run_bfs=True,
+            run_hrr_structural=True,
+            reason="env-force-expansion",
+        )
+
+    env_no_gate = _env_no_expansion_gate()
+    if env_no_gate is True:
+        return ExpansionDecision(
+            run_bfs=True,
+            run_hrr_structural=True,
+            reason="env-no-gate",
+        )
+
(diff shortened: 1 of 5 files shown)


robotrocketscience:
[claim:review:pascal:2026-05-13T20:34:03Z]

robotrocketscience (commented):
## Review

### Substance — looks good

The expansion-gate design is clean and matches #741 spec:

- **Resolver precedence** (force → no-gate → TOML → heuristics) implemented exactly as the issue body specifies. `_env_force_expansion` and `_env_no_expansion_gate` use a shared truthy/falsy frozenset; clean.
- **Heuristics honor PHILOSOPHY #605** — stdlib only, deterministic, regex-based. No embeddings, no model calls.
- **Three signals OR'd** (long-prompt, no-markers, question-form). Any "broad" trip wins. Conservative-by-design as the issue mandates.
- **`ExpansionDecision` frozen dataclass with `reason` field for telemetry** — exactly the surface #741 asked for.
- **`LaneTelemetry` additive fields** (`expansion_gate_reason`, `expansion_gate_skipped_bfs`) preserve byte-identical surface for pre-#741 callers; defaults are safe.
- **BFS-internal test fixture override** (`AELFRICE_FORCE_EXPANSION=1` in `test_bfs_multihop.py`'s isolated env) is the right call — keeps AC1-AC11 BFS-behavior tests asserting what they were built to assert rather than implicitly testing gate behavior.
- **`should_run_expansion` is called from both `retrieve()` and `retrieve_with_tiers()`** — full coverage of the retrieval surface.
- **`_starts_with_question_form` respects word boundaries** (rejects `whatever` matching `what`). Good.
- **HRR-structural lane intentionally not gated in v1** — comment explains the `run_hrr_structural` field is forward-compat. Defensible v1 scope per the issue.

### BLOCKER — rebase required before merge

Branch base is `13a57ac` (pre-#738). `github/main` is now at `8ea9ab5` after #738 merged today. The wider diff `github/main..HEAD` therefore shows the **inverse** of #738 plus the three feature commits. If this is FF-pushed in its current state via `merge-train`, #738 ships in reverse.

The bot's FF check will catch this — `merge-train.yml` requires `merge-base --is-ancestor github/main github/<branch>`, which is false here. So in practice the bot will reject and unlabel. But: **don't add `ready-to-merge` until after rebase**.

Action:

```
git fetch github main
git rebase github/main
# run unit suite again (expect AC4 test in test_search_tool_hook.py to need adjustment if rebasing on top of #740 once that lands, but that's not in main yet)
git push --force-with-lease
```

### Minor observations (non-blocking)

1. **`_read_toml_flag` has a `try: import tomllib / except ImportError: return None` fallback**, but `pyproject.toml` pins `requires-python >= 3.11` and the docstring comment notes "aelfrice ships >=3.11". The ImportError path is unreachable. Not a real bug — fail-soft principle covers it — just dead defensiveness. Up to you whether to drop it.

2. **Token count is `text.split()`** for the 80-token threshold. Naive whitespace-split; for a heuristic that's fine. Just noting for the record that the threshold is "roughly 80 whitespace-separated chunks" not "80 lexemes".

3. **Interaction with PreToolUse search-tool hook.** The gate fires inside `retrieve()`, which is called from `hook_search_tool._do_search`. PreToolUse queries are 1-5 extracted tokens (snake_case identifiers from the original Grep pattern), so they hit the structural-marker heuristic and gate verdict is `narrow` — which is correct. Just calling out that the gate's scope includes the PreToolUse fire as a happy side-effect; the bench fixture from #741 should sanity-check this.

4. **Interaction with #740 (in-flight as PR #744).** Both PRs touch `hook.py` around the `_write_hook_audit_record` callsite. After this PR lands and PR #744 rebases, the conflict resolution is mechanical (both adding kwargs after `latency_ms=`, no semantic overlap). Flagging so whoever lands second knows the rebase is trivial.

### Verdict

**Substance: approve.** Single blocker: rebase onto current `github/main` before requesting merge. Otherwise looks ready.

robotrocketscience:
[release:review:pascal:2026-05-13T20:36:30Z]

robotrocketscience:
[claim:review:pascal:2026-05-13T20:39:10Z]

robotrocketscience:
Re-checking on rebase: base is still `13a57ac` (pre-#738). `github/main` at `8ea9ab5` includes the search-tool default-flip. Cleared `attn:review` — re-flag once rebased so a reviewer picks it up. Substance review above stands (approved pending rebase).

robotrocketscience:
[release:review:pascal:2026-05-13T20:40:17Z]

robotrocketscience:
Rebased onto current `github/main` (now at `0e91fd1`, post-v3.0.1).

- Three feature commits replayed cleanly, no conflicts.
- All SSH-signed (G).
- Full suite: 3849 passed / 57 skipped / 75 xfailed in 83s.
- Discretion grep clean.
- Force-pushed with `--force-with-lease` after FF'ing local `main` first (locked-memory force-with-lease quirk avoided).

New tip: `bf288f5`. Ready for re-review / merge.

robotrocketscience:
[claim:review:pascal:2026-05-13T21:07:51Z]

robotrocketscience:
## Review

**Verdict:** code-LGTM with two non-blocking flags. CI green (pytest 3.12/3.13, CodeQL, staging-gate full suite). Discretion clean on the diff.

### What works
- Resolver precedence matches the issue spec exactly: env-force → env-no-gate → TOML → heuristics. Falsy/garbage env values fall through (verified by `test_env_force_falsy_falls_through_to_heuristics` / `test_env_garbage_value_falls_through`).
- Heuristics are stdlib-only — honors PHILOSOPHY (#605). `EDGE_TYPES` is imported from `aelfrice.models` so the structural-marker regex stays in lockstep with the canonical edge list.
- `LaneTelemetry` fields are additive with safe defaults (`""` / `False`), so pre-#741 callers keep working. `test_lane_telemetry_defaults_preserve_backcompat` pins this.
- The `isolated_env` fixture in `tests/test_bfs_multihop.py` sets `AELFRICE_FORCE_EXPANSION=1` via `monkeypatch`, so the existing AC1-AC11 BFS suite is exercising BFS behaviour, not gate behaviour — correct.
- `_starts_with_question_form` does a word-boundary check, so `whatever happens next` doesn't match the `what` prefix. Test pins it.
- Malformed TOML fail-soft works (catches `OSError`, `ValueError`, `TOMLDecodeError`).

### Flags

**1. Branch is behind `github/main`.** Merge-base is `13a57ac`; main has since landed 7 commits including v3.0.1 release + the #738 search-tool default-on series.

```
$ git merge-base --is-ancestor github/main pr-743
# returns non-zero (main is NOT ancestor)
```

The merge-train bot will refuse `ready-to-merge` until this is FF on main (`merge-train.yml` requires `git merge-base --is-ancestor github/main <branch>`). Rebase locally and force-push before labeling.

**2. Bench deferred but `Closes #741` will auto-close.** Issue #741's acceptance section has two load-bearing bench items:

> - Bench: broad-prompt p95 must improve by ≥30% vs gate-off (the load-bearing claim)
> - Bench: narrow-prompt p50 must not regress

PR body acknowledges the bench is deferred ("layered on top of #739's BFS-flip bench re-run"). If this merges with `Closes #741`, the issue auto-closes despite the bench gap. Three reasonable resolutions, pick one:

- (a) drop `Closes #741` from the body, ship code, leave #741 open until #739's bench run carries it;
- (b) keep `Closes`, accept the gap, file a follow-up bench-only issue and link it from the close comment;
- (c) hold the PR until #739 ratification can run the bench against this code path.

Operator call. I'd lean (a) since #741 explicitly defers to #739 for the bench, but (b) keeps the merge-train moving.

### Nits (non-blocking)
- Local `from aelfrice.expansion_gate import should_run_expansion` inside `retrieve()` / `retrieve_with_tiers()` rather than at module top. Matches the lazy-import pattern used elsewhere in `retrieval.py` (and avoids a potential `models` ↔ `retrieval` import cycle), but worth a comment if the intent is forward-compat. Not asking for a change.
- `_read_toml_flag` walks parents from `start` and stops at root or repeat. Correct, but a comment noting it mirrors `is_bfs_enabled`'s TOML walk would help future readers.
- N-gram entropy gate from issue §Design point 4 isn't shipped. Spec marks it "optional, behind flag" — fine to defer.

### Recommendation

Resolve flag #2 (operator decision on the bench / Closes-link). Then rebase on `github/main`, force-push, add `ready-to-merge`. Code change itself is good to land.

robotrocketscience:
[release:review:pascal:2026-05-13T21:10:32Z]

robotrocketscience:
[claim:review:curie:2026-05-13T21:13:18Z]

robotrocketscience (commented):
## Review

LGTM. Read all three commits + 27 expansion-gate tests + retrieval.py wiring + hook.py audit thread; design matches #741 spec and is well-bounded.

**Approved aspects:**
- `expansion_gate.py` — deterministic stdlib-only heuristics (#605 PHILOSOPHY honored). Length / marker-absence / question-form gates. Edge-type regex compiled once from `models.EDGE_TYPES`. Resolver precedence matches issue spec: env-force > env-no-gate > TOML > heuristics. Each branch returns an `ExpansionDecision` with `reason` populated for telemetry.
- TOML walk is parent-bounded by `seen` set + `cur.parent == cur` root-stop. Fail-soft on `TOMLDecodeError`, `OSError`, `ValueError`.
- `should_run_expansion(query, *, start=None)` — `start` kwarg lets callers (and tests) override the cwd-walk root. Empty/whitespace query short-circuits to `run_bfs=True` with `reason="empty-query"` so upstream L0-only retrieve behavior is unaffected.
- `retrieval.py` wiring — `gate_skipped_bfs = bfs_on and not gate_decision.run_bfs` is computed against the pre-gate `bfs_on`, then `bfs_on = bfs_on and gate_decision.run_bfs` applies it. Correct semantic: the flag tracks whether the gate forced a skip the user would otherwise have seen, not whether BFS would have run for other reasons.
- `LaneTelemetry` extension — `expansion_gate_reason: str = ""` and `expansion_gate_skipped_bfs: bool = False` defaults preserve the pre-#741 constructor contract; existing callers don't need to know about the new fields.
- BFS test isolation — `tests/test_bfs_multihop.py` sets `AELFRICE_FORCE_EXPANSION=1` in the isolated-env fixture so the gate doesn't shadow BFS-internal assertions. Right call: those tests assert BFS behavior, not gate behavior.
- HRR-structural lane stays on regardless of gate verdict in v1; field is reserved for forward-compat. Matches issue spec.

**Verification:**
- CI: pytest (3.12) ✅, pytest (3.13) ✅, CodeQL python ✅, staging-gate (secrets-scan, pattern-scan, history-scan, commit-msg-prefix, pr-title-prefix, pr-body-issue-link, release-docs-check) ✅, deptry + vulture ✅, typos ✅
- `mergeable: MERGEABLE`, all 3 commits SSH-signed (visible in git log)
- Discretion grep on diff: clean

**Minor observations (non-blocking):**
- N-gram entropy gate from issue §Design is not implemented; issue marked it optional/behind-flag so this matches scope.
- Bench (broad/narrow p95 ≥30% improvement) deferred per PR body — gated on #739 BFS-flip ratification, which makes the gate load-bearing. Code path is observable via `aelf tail` in the meantime.

Adding `ready-to-merge`.

robotrocketscience:
[release:review:curie:2026-05-13T21:14:51Z]

Issue #733, opened 2026-05-13
Auto-migrate non-uv aelfrice installs to uv tool on first 3.0.1 setup


robotrocketscience:
## Summary

Wire `aelf setup` (the 3.0.1+ binary) to detect a non-uv install context and run `uv tool install --force aelfrice` once per machine, behind a `~/.aelfrice/migrated-to-uv` sentinel. Makes the 2.1.0/pipx → 3.0.1/uv transition one-shot from the user's perspective when they type `/aelf:upgrade`.

## Why

Follow-up to #730 / #732. The collapsed upgrade surface (uv-only) needs a migration story for existing pipx / pip / venv installs. Without auto-migrate, the flow is:

1. User on 2.1.0/pipx clicks the banner → types `/aelf:upgrade`
2. Old 2.1.0 slash file runs `pipx upgrade aelfrice` → 3.0.1/pipx
3. User still on pipx; next `/aelf:upgrade` emits the migration `run:` line but user has to invoke a second time to act on it.

With auto-migrate hooked into `aelf setup` (step 3 of the old slash file's flow), step 2 finishes with the user on 3.0.1/uv in one click.

## Mechanism

The OLD 2.1.0 `/aelf:upgrade` slash file's step 3 is hardcoded to run `aelf setup`. That invocation resolves to the **new** 3.0.1 binary (because pipx already upgraded in step 2). The new `aelf setup` runs `maybe_migrate_to_uv()` before hook reconciliation:

1. If `upgrade_advice().context == 'uv_tool'` → no-op.
2. If `~/.aelfrice/migrated-to-uv` sentinel exists → no-op.
3. If `uv` is not on `\$PATH` → print "install uv (https://docs.astral.sh/uv/) then re-run /aelf:upgrade", skip migration, no sentinel write. User stays on pipx-3.0.1.
4. Otherwise run `uv tool install --force aelfrice` in a subprocess with a timeout, capture exit code:
   - Success: write sentinel, emit stderr notice naming the orphan venv path (`pipx uninstall aelfrice` for pipx; `pip uninstall -y aelfrice` for pip).
   - Failure: print captured stderr (truncated to 200 chars), no sentinel write — user can retry. Pipx install remains functional.

Subsequent setup invocations short-circuit on the sentinel. Hook reconciliation runs AFTER migration so the `aelf-hook` shim path resolves through the uv-tool install (which just overwrote the pipx shim at `~/.local/bin/`).

## What the user sees (happy path)

```
[...]

PR #774, opened 2026-05-14
fix(lifecycle): runnable uv install one-liner when maybe_migrate_to_uv hits uv-not-found

robotrocketscience:
## Summary

- v3.0.1 added `maybe_migrate_to_uv()` (#733) to auto-migrate pipx/pip installs to `uv tool install aelfrice` on first `aelf setup`. The migration is gated on `shutil.which("uv")`; a pipx-only user without uv hits the skip path.
- The skip-reason previously pointed at the docs URL only — no executable suggestion. This PR surfaces both Astral curl and brew install lines verbatim so the operator can copy-paste one and rerun.

## Why not auto-install uv?

Bootstrapping a third-party installer unprompted (`curl | sh` from inside our migration path) has a much larger blast radius than what a memory tool should do without user consent. The docs/INSTALL.md Prerequisites section already documents uv as required up front. This PR only sharpens the runtime error path for users who skipped that.

## What changes

- `src/aelfrice/lifecycle.py`: reason string for the `shutil.which("uv") is None` branch now names both install channels.
- `tests/test_migrate_to_uv.py`: assert both forms appear in `result.reason`.
- `CHANGELOG.md`: `[Unreleased]` / Fixed bullet describing the change and the rationale.

No behaviour change beyond the operator-facing text.

## Test plan

- [x] `uv run pytest tests/test_migrate_to_uv.py tests/test_setup_migrate_order.py tests/test_cli_setup.py tests/test_lifecycle.py` → 74 passed, 2 skipped
- [ ] Bench-smoke / full pytest matrix runs in CI on this PR

## Out of scope

- Auto-installing uv (see "Why not" above).
- Reflowing INSTALL.md — Prerequisites already documents uv up front and the migration callout is already present.


robotrocketscience:
[claim:review:fermat:2026-05-14T04:33:12Z]

robotrocketscience:
Rebased onto github/main (3 atomic commits, all signed G: feat → test → docs). Verified:

- Diff: `maybe_migrate_to_uv` reason string adds runnable `curl … | sh` and `brew install uv` lines; test asserts both forms present in `result.reason`; CHANGELOG `[Unreleased]` bullet added.
- No behavior change beyond operator-facing text. The `shutil.which("uv") is None` skip path is unchanged structurally; only the `reason` field differs.
- Bootstrap-via-curl pattern is in the *message*, not auto-invoked — addresses the original gripe without inheriting the auto-install blast-radius.
- Atomic commit ordering correct (feature → test → changelog).
- All required checks green after rebase (pytest 3.12/3.13, secrets-scan, pattern-scan, history-scan, CodeQL, deptry, vulture, typos, commit-msg-prefix, pr-title-prefix, pr-body-issue-link, bench-smoke).
- Discretion grep on diff vs main: clean.

Labeling `ready-to-merge`.

robotrocketscience:
[release:review:fermat:2026-05-14T04:38:23Z]

PR #788, opened 2026-05-14
fix(setup): collapse hook accumulation + prune dead aelf-* entries (#781)

### LINKED ISSUE #781: fix(setup): aelf setup appends duplicate hook entries instead of dedup/replacing — every stale venv accumulates a permanent ENOENT
## Symptom

`~/.claude/settings.json` (user scope) accumulates duplicate hook entries pointing at venv paths that no longer exist. Every Claude Code tool call then emits one `ENOENT` per stale entry:

```
Bash hook error: /bin/sh: /private/tmp/aelf-release-v3.0.1/.venv/bin/aelf-search-tool-hook: No such file or directory
Bash hook error: /bin/sh: /Users/<user>/projects/aelfrice/.venv/bin/aelf-commit-ingest: No such file or directory
...
```

On a heavily-used setup (multiple worktrees, tmp venvs, project venvs), each lifecycle event ends up with 5–7 duplicate hook commands, only one of which resolves. The other 4–6 fire and fail every tool call.

Observed sample from `~/.claude/settings.json` after several months of normal use:

```
SessionStart:        7 entries (1 working, 6 stale)
PreToolUse search:   6 entries (1 working, 5 stale)
UserPromptSubmit:   16 entries (4 working, 12 stale)
Stop:               17 entries (5 working, 12 stale)
PreCompact:          9 entries (3 working, 6 stale)
PostCompact:         8 entries (2 working, 6 stale)
```

Live reproduction: even after I dedup-edited the file by hand, fresh `aelf setup` runs from new worktrees re-added stale entries within minutes (worktree gets deleted → entry stays).

## Reproduction

Minimum repro:
1. `uv tool install aelfrice` — installs at `~/.local/bin/aelf-hook` etc.
2. `aelf setup --scope user` (or whatever currently runs from a clean user dir).
3. Create a temp venv: `cd /tmp/scratch && python -m venv .venv && .venv/bin/pip install aelfrice`.
4. `cd /tmp/scratch && aelf setup` — observe: it appends entries pointing at `/tmp/scratch/.venv/bin/aelf-*` to **user scope** instead of replacing the existing user-scope entries, or instead of writing to project-scope.
5. `rm -rf /tmp/scratch`.
6. Open Claude Code anywhere → every Bash/Grep/UserPromptSubmit fires N stale-path hook errors.

## Root cause hypotheses

1. **`aelf setup` appends without dedup.** When called multiple times, it adds a new `hooks[event][N]` entry instead of replacing the existing entry whose `command` basename matches (`aelf-hook`, `aelf-transcript-logger`, `aelf-stop-hook`, etc.).
2. **Scope detection is leaky.** The `cwd/.venv` heuristic (per the design described at `resolve_hook_command(scope)`) appears to mis-classify some project-local runs as user-scope, so project-venv paths leak into the global user-scope settings.
3. **No prune step.** There is no path that scans existing hook entries, drops ones whose `command` doesn't resolve, and rewrites the file. So once a venv is deleted, the entry survives forever.

## Recommended fix

Three layers, smallest to largest:

### A. `aelf setup` dedupe before append (smallest)

For each `(event, basename(command))` tuple, replace the existing entry instead of stacking. Match by basename so a path-prefix change (uv tool upgrade, pipx → uv migration, venv recreation) does not create a duplicate.

Pseudocode for the write step:

```python
def install_hook(settings, event, new_entry):
    basename = Path(new_entry["hooks"][0]["command"]).name
    settings["hooks"][event] = [
        e for e in settings["hooks"].get(event, [])
        if Path(e["hooks"][0]["command"]).name != basename
    ]
    settings["hooks"][event].append(new_entry)
```

### B. `aelf setup` prune dead paths (mid)

Before writing, walk the existing `hooks` tree and drop any entry whose `command` first token does not resolve via `Path(...).exists()` or `shutil.which(...)`. Emit a one-line `[setup] pruned N stale hook entries` notice so the user sees what happened.

Don't touch non-aelf hooks (anything whose basename does not start with `aelf-`). Match the user's custom shell hooks (`gh-pii-guard.sh`, `conversation-logger.sh`, etc.) — leave them alone even if their paths look unusual.

### C. `aelf doctor --fix` (mid-large)

`aelf doctor` already detects unresolvable hook commands and exits 1 (per the linter design). Add `--fix` that performs the prune from §B, with a dry-run flag for visibility. This gives users a self-service recovery path without needing to know that `~/.claude/settings.json` even exists.

## Severity

**High for any user not on the happy path.** Failure mode:

- User installs aelfrice → works fine
- User upgrades aelfrice with a venv-path change, or reinstalls in a different env, or runs `aelf setup` more than once, or installs in multiple projects → settings starts accumulating
- After 2-3 such events: every tool call fires 2-3 ENOENT errors in the IDE/CLI panel
- After ~10 events: the spam is unusable

The original happy-path user is unaffected, but the "I've been using this for 3 months" user gets a degrading experience that they cannot self-diagnose without reading JSON.

## Prior art

#39 (`feat: add setup.py module — idempotent UserPromptSubmit hook install`) claimed idempotency for the UPS event. Either that idempotency was scoped to UPS only and not extended to the other lifecycle events, or it was lost in a later refactor of the setup writer.

## Tier

`queen`. Touches `aelf setup`, `aelf doctor`, and the settings.json writer; needs careful idempotency reasoning around the basename-match rule.

## Acceptance

- [ ] `aelf setup` is idempotent across all hook events: running it twice in a row produces the same `~/.claude/settings.json`.
- [ ] `aelf setup` prunes hook entries whose `command` doesn't resolve, scoped to `aelf-*`-basename entries only.
- [ ] `aelf doctor --fix` performs the same prune on demand (read-only `aelf doctor` behavior unchanged).
- [ ] Unit test: build a settings.json with N stale entries, run setup, assert only the one resolvable entry remains per event.
- [ ] Unit test: settings.json with mixed `aelf-*` and custom shell hook commands; assert the custom hooks survive.

## Out of scope

- Migrating project-scope settings (`<project>/.claude/settings.json`) — same mechanism but different blast radius; file as a follow-up if §A/B/C ship clean.
- Backwards-compat shim for users with the broken state — they get auto-cleaned on next `aelf setup` run; that's sufficient.

robotrocketscience:
Closes #781.

## Summary

Three layers that together stop `~/.claude/settings.json` from
accumulating ENOENT-firing `aelf-*` hook entries every time the
package install path changes (uv tool upgrade, pipx → uv migration,
project venv recreation, tmpdir scratch install).

### Layer A — install-side dedupe by `(matcher, basename)`

`install_*_hook` previously checked exact-command match for
idempotency. A second install with a different absolute path
appended instead of replacing, so after a few install events the
file looked like:

```
SessionStart:        7 entries (1 working, 6 stale)
PreToolUse search:   6 entries (1 working, 5 stale)
UserPromptSubmit:   16 entries (4 working, 12 stale)
…
```

New helper `_install_or_replace_entry` collapses on
`(matcher, basename)` so a path change to the same `aelf-*` script
replaces the existing entry. Matcher-aware: the search-tool
Bash matcher and Grep|Glob matcher (same basename, different
matcher) remain separate per the v1.5.0 contract.

### Layer B — `aelf setup` prune step

Layer A only catches collisions on the basename being written *this
run*. Stale entries for events `aelf setup` isn't touching (e.g. a
`--no-stop-hook` opt-out leaving an old Stop entry around) still
leak through.

New `doctor.prune_broken_aelf_hooks(path)` walks every
`hooks.<event>[i].hooks[j]` command, removes parent entries whose
program basename starts with `aelf-` AND whose program fails the
existence / `$PATH` lookup. The predicate is the same
`_inspect_command` doctor already uses for its scan, so the prune
decision matches what `aelf doctor` reports. Custom shell hooks
(non-`aelf-` basenames) are never touched.

`_cmd_setup` invokes this after `clean_dangling_shims` and before
the install loop, printing a single line summarising what was
removed.

### Layer C — `aelf doctor --fix`

Same prune helper, surfaced as `aelf doctor [hooks] --fix`. Combine
with `--dry-run` to see what would be pruned without rewriting the
file. The exit code still reflects the pre-fix findings — re-run
`aelf doctor` after `--fix` to confirm a clean state.

## Why this isn't covered by the existing setup idempotency story

`install_*_hook` was idempotent against *the exact same command
string*. It was never idempotent against the same logical hook with
a path that drifted. #39 ("idempotent UserPromptSubmit hook
install") only covered the equal-string case; venv churn through
the past year of upgrades is what exposed the gap.

## Test plan

- [x] `tests/test_setup_prune_stale.py` — 12 new tests covering A
      (path-change replace, exact idempotency, matcher partitioning),
      B (N stale → 1 resolvable, custom-hook survival, dry-run,
      missing file, shell-meta skip, `aelf setup` integration), and
      C (`aelf doctor --fix` real + dry-run + clean-settings).
- [x] Existing setup/doctor tests pass (`tests/test_setup_*.py`,
      `tests/test_doctor*.py`, `tests/test_cli_setup.py`,
      `tests/test_commit_ingest_hook.py`).
- [x] Full `uv run pytest tests/` clean (4014 passed, 62 skipped,
      75 xfailed).
- [x] Discretion grep on the full diff vs `github/main` clean.

## Acceptance map

- [x] `aelf setup` idempotent across all hook events: running it
      twice produces the same settings.json (Layer A).
- [x] `aelf setup` prunes hook entries whose `command` doesn't
      resolve, scoped to `aelf-*`-basename entries only (Layer B).
- [x] `aelf doctor --fix` performs the same prune on demand
      (Layer C). Read-only `aelf doctor` behaviour unchanged.
- [x] Unit test: build a settings.json with N stale entries, run
      setup, assert only the one resolvable entry remains per event
      (`test_prune_collapses_n_stale_to_one_resolvable`).
- [x] Unit test: settings.json with mixed `aelf-*` and custom shell
      hook commands; assert the custom hooks survive
      (`test_prune_leaves_custom_shell_hooks_alone`).

## Out of scope (deferred)

- Project-scope settings (`<project>/.claude/settings.json`) — same
  mechanism applies but the issue defers it as a follow-up. The
  prune helper already supports project paths; only the
  `aelf doctor` defaults wire to user scope.
- Backwards-compat shim for users with the broken state — they get
  auto-cleaned on next `aelf setup` (Layer B fires) or
  `aelf doctor --fix`. No migration code needed.

### DIFF
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -3800,6 +3800,8 @@ def _cmd_doctor(args: argparse.Namespace, out: object) -> int:
         _print_doctor_store_check(out)
         if report.broken:
             exit_code = 1
+        if getattr(args, "fix", False):
+            _cmd_doctor_fix_hooks(args, report, out)
     if scope in (None, "graph"):
         if scope is None:
             print("", file=out)  # type: ignore[arg-type]
@@ -3807,6 +3809,53 @@ def _cmd_doctor(args: argparse.Namespace, out: object) -> int:
         if graph_exit != 0:
             exit_code = graph_exit
     return exit_code
+
+
+def _cmd_doctor_fix_hooks(
+    args: argparse.Namespace, report: object, out: object,
+) -> None:
+    """Apply `prune_broken_aelf_hooks` to every settings.json doctor scanned.
+
+    Called from `_cmd_doctor` when `--fix` is set. Reads the list of
+    scopes/paths off the already-built `DoctorReport` so the prune
+    decision matches what doctor just reported. `args.dry_run` flips
+    the prune into report-only mode.
+
+    Custom hooks (non-`aelf-` basename) are never touched by the
+    prune helper itself — this wrapper only handles I/O.
+    """
+    from typing import cast
+
+    dry_run = bool(getattr(args, "dry_run", False))
+    scopes_scanned = cast(
+        list[tuple[str, Path]],
+        getattr(report, "scopes_scanned", []),
+    )
+    if not scopes_scanned:
+        return
+    print("", file=out)  # type: ignore[arg-type]
+    any_removed = False
+    for _scope, path in scopes_scanned:
+        prune = prune_broken_aelf_hooks(path, dry_run=dry_run)
+        if not prune.total_removed:
+            continue
+        any_removed = True
+        events = ", ".join(
+            f"{event}={count}"
+            for event, count in sorted(prune.removed_per_event.items())
+        )
+        verb = "would prune" if dry_run else "pruned"
+        noun = "entry" if prune.total_removed == 1 else "entries"
+        print(
+            f"--fix: {verb} {prune.total_removed} stale aelf-* hook "
+            f"{noun} ({events}) from {path}",
+            file=out,  # type: ignore[arg-type]
+        )
+    if not any_removed:
+        print(
+            "--fix: no stale aelf-* hook entries to prune",
+            file=out,  # type: ignore[arg-type]
+        )
 
 
 def _cmd_doctor_meta_beliefs(args: argparse.Namespace, out: object) -> int:
@@ -5676,6 +5725,18 @@ def build_parser(*, show_advanced: bool = False) -> argparse.ArgumentParser:
         help=(
             "with --meta-beliefs: emit JSON instead of a human-readable "
             "table. No effect on other doctor sub-modes today."
+        ),
+    )
+    p_doctor.add_argument(
+        "--fix",
+        dest="fix",
+        action="store_true",
+        default=False,
+        help=(
+            "with hooks scope: remove `aelf-*` hook entries whose program "
+            "no longer resolves (#781). Custom shell hooks (non-`aelf-` "
+            "basenames) are never touched. Combine with --dry-run to "
+            "print what would be pruned without rewriting settings.json."
         ),
     )
     p_doctor.set_defaults(func=_cmd_doctor)
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -99,6 +99,7 @@ from aelfrice.derivation_worker import run_worker
 from aelfrice.doctor import (
     DORMANT_IDLE_DAYS_DEFAULT,
     DormantDB,
+    HookPruneResult,
     _check_dormant_dbs,
     classify_orphans as _classify_orphans,
     diagnose,
@@ -108,6 +109,7 @@ from aelfrice.doctor import (
     format_report,
     gc_orphan_feedback as _gc_orphan_feedback,
     promote_retention as _promote_retention,
+    prune_broken_aelf_hooks,
 )
 from aelfrice.llm_classifier import (
     ENV_API_KEY as _LLM_ENV_API_KEY,
@@ -2453,6 +2455,23 @@ def _cmd_setup(args: argparse.Namespace, out: object) -> int:
     for removed_path in cleanup.removed:
         print(
             f"cleaned dangling shim: {removed_path}",
+            file=out,  # type: ignore[arg-type]
+        )
+    # #781: prune any `aelf-*` hook entries pointing at vanished venv
+    # paths before reconciling installs. The dedupe-on-install path
+    # (Layer A) only catches collisions on the basename we are about
+    # to write; old stale entries for events we are NOT installing
+    # this run still need to be swept out.
+    prune = prune_broken_aelf_hooks(path)
+    if prune.total_removed:
+        events = ", ".join(
+            f"{event}={count}"
+            for event, count in sorted(prune.removed_per_event.items())
+        )
+        print(
+            f"pruned {prune.total_removed} stale aelf-* hook entr"
+            f"{'y' if prune.total_removed == 1 else 'ies'} "
+            f"({events}) from {prune.settings_path}",
             file=out,  # type: ignore[arg-type]
         )
     result = install_user_prompt_submit_hook(
--- a/src/aelfrice/doctor.py
+++ b/src/aelfrice/doctor.py
@@ -837,6 +837,171 @@ def _check_path(
     )
 
 
+# ---------------------------------------------------------------------------
+# Prune dead `aelf-*` hook entries from settings.json (#781).
+# ---------------------------------------------------------------------------
+
+# Basename prefix that marks an entry as one this package installed.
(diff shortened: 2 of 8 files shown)


robotrocketscience:
[claim:review:pascal:2026-05-14T15:54:42Z]

robotrocketscience:
LGTM — approving for the merge-train.

Verification:
- 4 atomic `G`-signed commits, prefixes correct (`fix:` `feat:` `feat:` `test:`).
- All CI green: pytest 3.12/3.13, secrets-scan, pattern-scan, history-scan, CodeQL, bench-smoke, deptry, vulture, calibration.
- Discretion grep on diff vs `github/main`: clean.
- Branch FF on main; mergeStateStatus BLOCKED is awaiting label only.
- Spot-checked the three load-bearing pieces:
  - `_install_or_replace_entry` dedupes on `(matcher, basename)` — matcher-aware partitioning preserves the v1.5.0 contract where the search-tool Bash matcher and Grep|Glob matcher (same basename, different matcher) are intentionally separate entries.
  - `prune_broken_aelf_hooks` uses the same `_inspect_command` predicate `aelf doctor` already trusts for its scan, so prune decisions match what doctor reports. `aelf-`-basename prefix gate keeps custom shell hooks (`gh-pii-guard.sh`, `conversation-logger.sh`, etc.) untouched.
  - `_atomic_rewrite_settings` byte-for-byte matches `setup._atomic_write`'s format (indent=2, ensure_ascii=False, trailing newline) so a setup→prune→setup round trip stays diff-stable. The "pulled into doctor to avoid private setup symbol import" rationale is sensible.
- Conservative-skip: `_entry_is_broken_aelf_hook` returns False when `_inspect_command` returns anything other than `status="broken"` (so the soft-warning silent-failure pattern wrappers don't get pruned). Right call.
- Test coverage: 12 new tests in `tests/test_setup_prune_stale.py` — covers Layer A path-change replace + exact-idempotency + matcher partitioning, Layer B N-stale→1-resolvable + custom-hook survival + dry-run + missing file + shell-meta-skip + `aelf setup` integration, Layer C `aelf doctor --fix` real + dry-run + clean-settings.

Nit (non-blocking): `_atomic_rewrite_settings` re-imports `os` and `tempfile` inside the function body. Module-level imports already cover `os` and `tempfile`; the function-local imports are harmless (Python's import cache makes them no-ops) but redundant. Drop them on the next touch.

Adding `ready-to-merge`.

robotrocketscience:
[release:review:pascal:2026-05-14T15:56:05Z]

Issue #797, opened 2026-05-14
v3.1: tighten post-clear answer template — hot_start fidelity 0.70 < 0.80 AC, recall gap downstream of #778


robotrocketscience:
## Summary

The κ-calibrated v3.1 hot-start bench from #777 surfaces a **post-clear answer-prompt recall gap**: at the default post-clear prompt shape (rebuilt block + user_turn, no further instruction), real-replay fidelity comes in at 0.70 vs the #592 AC bar of 0.80. Inter-judge κ = 1.0 on the same run, so the rubric is reproducible and the gap is not noise.

## Public evidence

- κ artifact (#777 close-out comment): n=9, κ_min = κ_mean = 1.0, fidelity = 0.667 across 3 judges.
- Per-turn diagnosis from the unanimous failures:
  - **idx 9** — dropped specific facts: post-clear answer omitted file path + CRLF root cause; substituted an unrelated "fails in full suite" claim.
  - **idx 11** — cross-contamination: answered with the dedup-tuple shape (idx 21's atom) instead of the actual one-line fix (idx 17's atom).
  - **idx 13** — question-conflation: answered the debug-command question (idx 25) when asked the verification-step question (idx 13).
- All three failures pass the judge rubric's "specific facts" test for *some* fact, just not the one the current user_turn asked for.

## Hypothesis

The post-clear child task's prompt today is the minimal:
```
{rebuilt_block}\n---\n{user_turn}
```
Per `benchmarks/context-rebuilder/eval_harness.py:replay_post_fork`. There is no instruction insisting the answer recapitulate named atoms or stay scoped to the specific user_turn. A real host-CLI session post-clear has analogous failure modes when the rebuild block is dense with several adjacent facts.

## Scope — operator decision needed

Two pre-registered options:

### P1 — tighten the post-clear answer template at the harness boundary

`replay_post_fork` writes a richer per-row prompt to `replay_requests.jsonl`. The new template anchors the answer in the user_turn and instructs the dispatcher's child task to recapitulate the specific named atoms from the rebuild block when the user_turn asks for them. Pure-Python; no rubric or fixture change. Risk: shifts the bench off-spec for measuring "what a typical host-CLI session post-clear actually recalls" — we'd be measuring rebuilder-pack quality at the limit of an instructed reader, not at the limit of a typical post-clear reader.

### P2 — accept the 0.70 floor and re-baseline the #592 AC

[...]

Issue #746, opened 2026-05-13
Flip default: rebuilder trigger_mode manual → threshold — bench gates #592 + #687 cleared


robotrocketscience:
## Summary

Flip `DEFAULT_TRIGGER_MODE` in `src/aelfrice/context_rebuilder.py:197`
from `TRIGGER_MODE_MANUAL` to `TRIGGER_MODE_THRESHOLD`. One-line
change. Brings the PreCompact rebuilder to a default-on posture so
fresh installs experience the augment-mode rebuild without needing
a `~/.aelfrice.toml [rebuilder] trigger_mode = "threshold"` override.

This is bench-gated work that has **already cleared its gate**:

- `#592` (eval-harness commit-3 closed 2026-05-13). Bench complete,
  single-judge: hot-start 100% at `trigger_threshold ≤ 0.6`,
  cold-start 75% across t ∈ {0.5..0.8}.
- `#687` (Cohen's-κ multi-run validation, closed 2026-05-13). The
  post-bench ratification gate.

Both gates are now CLOSED. The conservative ship-default from v1.4.0
("manual until production telemetry confirms the threshold") was
correct at v1.4 but is now overdue for the flip at v3.0.

## Why this is the right intent

1. **README implies auto-rebuild is on.** The README's compaction
   discussion talks about retrieval-curated rebuild "when the harness
   compacts." With `trigger_mode = "manual"` as the ship-default, that
   surface never fires on a fresh install — only `aelf rebuild` /
   `/aelf:rebuild` produces output. Documented contract ≠ default
   behavior.

2. **The original conservatism rationale is no longer load-bearing.**
   `context_rebuilder.py:198-202`:
   > "Manual is the default until production telemetry confirms the
   > calibrated threshold."
   `#592` produced that telemetry. The calibration JSON at
   `benchmarks/context_rebuilder/calibration_v1_4_0.json` resolved
   to `threshold_fraction = 0.6` and the bench cleared. The gate
   exists to be flipped.

3. **Augment-mode means low blast radius.** The harness's own
   compaction summary still runs alongside the rebuilder block. The
   net effect is "more context preserved during compaction," not
   "rebuilder replaces compaction." Worst case from the flip is
   per-compaction token spend goes up by the rebuilder's block size
   (~5-6 KB observed); best case is users no longer manually `/clear`
   because the auto-compact now preserves load-bearing beliefs.

## Current state (verified)

[...]

Issue #476, opened 2026-05-08
[v2.1] PR-smoke fixtures for benchmark adapters — license review required


yoshi280:
Land `tests/fixtures/bench_smoke/` so a 2-minute PR smoke job can exercise real adapter shapes without HuggingFace download or external clones.

## Origin

PR #465's gate commit (60e1b68) flagged this as scope for a follow-up:

> **PR-smoke fixtures (offline): not landed in this PR.** Spec recommended `tests/fixtures/bench_smoke/` pinned fixtures so a 2-min PR smoke job can exercise real adapter shapes without HF download. This needs licensed-content review (LoCoMo turns, MAB QA) and is scope for a follow-up. Pytest unit tests cover dispatcher regressions in the meantime via stubs.

## Acceptance

- `tests/fixtures/bench_smoke/locomo_micro.json` — 1–2 conversations with 3–5 sessions each, sufficient to exercise the LoCoMo adapter's load → ingest → retrieve path.
- `tests/fixtures/bench_smoke/mab_micro.json` — 5–10 QA pairs across the four splits (Conflict_Resolution / Test_Time_Learning / Long_Range_Understanding / Accurate_Retrieval).
- `tests/fixtures/bench_smoke/structmemeval_micro/` — 1 case per task type (location / accounting / recommendations / tree).
- `tests/fixtures/bench_smoke/longmemeval_micro.json` — 3 questions, retrieve-only.
- `tests/test_bench_smoke.py` — pytest module that runs each adapter's `--retrieve-only` path against its micro fixture, asserts non-zero retrieved-belief counts, completes in <2 min total wall.
- CI job `bench-smoke` runs on every PR (vs the existing nightly `bench-canonical` cron).

## Hard prerequisite — license review

**Cannot land until license verification on:**

- **`snap-research/locomo`** — Repo license (whether the conversation turns are redistributable in a derivative public repo) needs explicit review. Conversation content may be CC-BY / CC-BY-NC / proprietary depending on the upstream license file.
- **`huangchaoyi/MemoryAgentBench`** — HuggingFace dataset license; same redistribution question for the QA pairs.
- **`yandex-research/StructMemEval`** — Apache-2 / MIT / proprietary; same.
- **`xiaowu0162/longmemeval`** — same.

[...]

PR #803, opened 2026-05-14
docs(onboard): preview cost, time, and quota impact before subagent dispatch

robotrocketscience:
## Summary

Adds a cost/time/quota preview block to `/aelf:onboard` so users see *what is about to happen* and *what it will cost* before the slash command dispatches a wave of Haiku Task subagents.

The current onboard flow looks like this from the user's seat:

```
12 background agents launched (↓ to manage)
 ├ Classify onboard batch 1/46
 ├ Classify onboard batch 2/46
 ├ ...
```

…repeated several times, with no up-front information on cost, duration, or whether this counts as a separate API charge. For a Max-plan user worried about the 5h / 7d quota, the wall of "12 background agents launched" lines reads as alarming even though Haiku 4.5 classification is cheap and finishes in a few minutes.

This PR adds a single human-readable block, computed from the actual batched payload (not fixed numbers), printed once before any subagent dispatches:

```
═ aelf onboard: parallel classification ═
what  2,276 candidate sentences extracted from <path>. Each one needs
      to be typed (factual / preference / requirement / correction)
      or filtered as noise, so retrieval later can prioritize real
      facts over boilerplate.
how   46 batches of ≤50 sentences, dispatched to Claude Haiku 4.5
      subagents in parallel waves of ~12 — that's why you'll see
      "12 background agents launched" repeated several times below.
time  ~2-4 min  (subagent dispatch + Haiku latency)
cost  ~152K input + ~91K output tokens
      ≈ $0.61 equivalent Haiku 4.5 API spend
      billing: runs against your Claude Code subagent quota
               (Max / Pro / Team) — no separate API charge.
════════════════════════════════════════
```

## How estimates are computed

All from the actual batched payload — the spec describes the formulas inline so the host computes them at runtime:

- **Input tokens** per batch: `ceil((1200 + sum(len(text) + len(source) + 30 for s in batch)) / 4)`. `1200` is the rough character length of the classification template. `/4` is the standard "~4 chars per token" approximation from Anthropic's [glossary](https://docs.anthropic.com/en/docs/resources/glossary#tokens).
- **Output tokens** per batch: `40 * len(batch)`. One short JSON entry per sentence.
- **Cost**: priced at Claude Haiku 4.5's [public rates](https://www.anthropic.com/pricing) (`$1.00/MTok input, $5.00/MTok output`, as of 2026-05).
- **Wall time**: `ceil(n_batches / 12) * 60s ± 30s` — subagents dispatch in parallel waves of ~12 and Haiku finishes a 50-sentence batch in ~30–90s wall-clock counting spawn/queue overhead. Rendered as a range.

Estimate is an upper-bound order-of-magnitude, not a guarantee, and the spec documents that caveat (per-batch retries and subagent system-prompt overhead aren't observable from the host).

## Files changed

- `src/aelfrice/slash_commands/onboard.md` — adds the new step 4 ("Preview cost, time, and quota impact"); renumbers Classify → 5, Aggregate → 6, Apply → 7, Display → 8.
- `CHANGELOG.md` — entry under `[Unreleased] → Changed`.

No CLI, no Python, no tests touched. This is a spec change to a markdown file Claude reads when executing the `/aelf:onboard` slash command.

## Test plan

- [x] `pytest tests/test_slash_commands.py tests/test_cli_onboard_handshake.py` — 133 passed, 1 skipped (unchanged).
- [ ] Manually run `/aelf:onboard .` and confirm the preview block prints before the first "background agents launched" line, with sensible numbers.

## Why this is a docs/spec change, not a Python change

The slash command spec is the source of truth — Claude reads `onboard.md` at slash-command invocation time and executes the steps. The cost/time formulas live in the markdown so they can be revised independently of the CLI (e.g. when Haiku pricing changes, when wave-size heuristics change, when token rules-of-thumb need updating). No new code paths to test in isolation; the value lives in what the host renders.

## Summary by CodeRabbit

* **New Features**
  * Added pre-dispatch progress information in the onboarding workflow, displaying the approximate number of processing items, classifier model details, and estimated token usage, costs, and completion time before processing begins.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/803)


robotrocketscience:
[claim:review:yeats:2026-05-14T19:27:06Z]

robotrocketscience:
**Review — request changes (yeats)**

CI green and the UX rationale is solid — pre-flight cost/time blocks are exactly the friction-removal a first-time `/aelf:onboard` user needs. Two blockers and two minor notes before this lands.

### Blocking

1. **CHANGELOG description does not match what shipped.** The entry says the spec "gains a new step 4 between batching and subagent dispatch" and the PR body lists the renumbering `Classify → 5, Aggregate → 6, Apply → 7, Display → 8`. The diff does neither — the preview is inlined *inside* step 3 ("Batch.") as an additional sub-block, and steps 4–7 stay at their current numbers (`git show pr-803:src/aelfrice/slash_commands/onboard.md | grep -nE '^[0-9]+\. '` confirms: step 4 is still Classify, step 5 Aggregate, step 6 Apply, step 7 Display). Either renumber to ship what the CHANGELOG describes, or rewrite the CHANGELOG entry and PR body to describe the inline-into-step-3 reality. The inlined form is arguably cleaner (the preview *is* part of preparing the batch) — happy with that as the shipped form, but the CHANGELOG cannot stay misleading after merge.

2. **Branch is 12 commits behind `main` and will not FF.** `git merge-base --is-ancestor github/main pr-803` returns NEEDS-REBASE; `pr-803..github/main` includes a `#797` Changed-section entry that landed between `#476` and `#746` (exactly where this PR also inserts), so the CHANGELOG hunk will need three-way merge. The single onboard.md hunk should rebase clean.

### Non-blocking notes

3. **`<model>` rendering will print the literal value `haiku`.** Step 4's dispatch sets `model: haiku` (not a versioned id), so the spec's instruction to "name the actual classifier model from the `model:` field … render it literally as `<model>` from that field" produces a preview block that says `dispatched to \`haiku\` subagents`. Two issues with that: (a) the user-visible string is less informative than the PR-body example (`Claude Haiku 4.5`), and (b) the spec separately tells the host to "price at the classifier model's currently published per-token rates from the provider's public pricing page" — but `haiku` doesn't map to a single pricing-page row without an extra resolution step the spec doesn't describe. Either pin a specific model id in the spec and bump it when Anthropic ships the next Haiku, or add a one-line resolution rule ("`haiku` resolves to the current Claude Haiku model on the pricing page"). Not a merge blocker, but the gap will surface the moment a user looks at the printed cost and tries to reconcile.

4. **Manual UX checkbox unchecked.** PR body's test plan still shows `[ ] Manually run /aelf:onboard . and confirm the preview block prints before the first "background agents launched" line`. Spec-only diff so I'm not requiring it, but the rendered block hasn't been eyeballed by anyone yet — worth doing once before merge.

### What I'm NOT blocking on

- Token / wall-time formulas — sensible, document their own approximation, match published rates as of 2026-05.
- The "no separate API charge" reassurance — load-bearing for Max-plan users and correctly scoped.
- Discretion grep on diff: clean (the `subagent` / `Haiku` hits are in user-facing spec context that already lives on `main`).

Release claim follows.

robotrocketscience:
[release:review:yeats:2026-05-14T19:29:38Z]

PR #806, opened 2026-05-14
feat(onboard): rejection ledger so re-runs don't re-classify rejected sentences (#801)

### LINKED ISSUE #801: aelf onboard <dir> is not idempotent — re-runs re-classify previously-rejected sentences
## Repro

Run `aelf onboard .` on a project, then run it again with no file changes. The second run still produces hundreds of candidates and burns LLM calls (Haiku subagents in my case) re-classifying the same noise.

Observed on a real onboard run:

| Pass | Candidates | Inserted | Rejected (`persist: false`) |
|------|-----------:|---------:|----------------------------:|
| 1    | 2276       | 1550     | 723                         |
| 2    | 726        | 230      | 494                         |
| 3    | 496        | (interrupted) | —                      |

Pass 2's 726 candidates are almost entirely the 723 rejects from pass 1. Pass 3 is the rejects from pass 2. It converges eventually but each pass costs N classification calls and each pass also has a nonzero chance of *persisting* noise that the previous pass correctly rejected (different model roll).

## Cause

`onboard --emit-candidates` filters against the belief store, but sentences classified `persist: false` aren't stored — so they re-emit forever. The store dedupes on `(text, source)` for *persisted* beliefs only.

## Proposed fix

Persist a `(text, source)` rejection ledger during `--accept-classifications`. Filter against it in `--emit-candidates`. Add a `--force` flag to reclassify rejects on demand (for cases where the user wants to re-roll because they think the prior pass was too aggressive).

Result:
- Re-running an unchanged directory → 0 candidates (true no-op)
- Re-running after edits → only genuinely new sentences
- `--force` → current behavior

## Alternative considered

A project-level "already onboarded" flag was suggested but rejected because it blocks the legitimate "I added new docs, re-onboard the directory" case unless `--force` is used. The rejection-ledger approach handles incremental onboarding naturally.

## Workaround today

None. Every `aelf onboard` invocation costs the full classification pass regardless of corpus state.

robotrocketscience:
## Summary

Closes #801. `aelf onboard <path>` historically re-classified every
sentence the host had previously rejected with `persist=False`, because
rejected sentences are never stored as beliefs and the dedup filter in
`start_onboard_session` only consults `beliefs.id`. The recorded repro
in the issue: pass 1 emits 2276 candidates → 723 rejected; pass 2 emits
726 candidates ≈ the 723 rejects + a few new files; pass 3 the same
again. Each pass burns one classification round-trip per re-rejected
sentence.

This PR persists `(belief_id, text, source, rejected_at)` rows in a new
`onboard_rejections` table from `accept_classifications` whenever the
host returns `persist=False`, and `start_onboard_session` +
`check_onboard_candidates` filter against that table the same way they
already filter against `beliefs`. `aelf onboard --force` bypasses the
ledger so the operator can re-roll a prior pass they think was too
aggressive.

## Behavior change

| Pass                                | before               | after          |
|-------------------------------------|----------------------|----------------|
| 1 (fresh, all rejected)             | 2276 → 723 rejected  | unchanged      |
| 2 (no file changes)                 | 726 candidates       | **0 candidates** |
| 2 with `--force`                    | (no such flag)       | re-emits the 723 |
| 2 after pass-1 accepted everything  | 0 candidates         | unchanged      |
| 2 with `--force` after pass-1 accept| (no such flag)       | 0 candidates (already-present still filtered) |

The slash command's `n_new == 0` short-circuit (`onboard.md` step 2)
keeps working because rejected sentences move out of `n_new` into the
new `n_already_rejected` bucket — no slash-command change needed.

## Files changed

- `src/aelfrice/store.py` — `onboard_rejections` table in `_SCHEMA`
  (forward-compat via `CREATE TABLE IF NOT EXISTS`); five new
  `MemoryStore` methods (`insert_onboard_rejection`,
  `delete_onboard_rejection`, `is_onboard_rejected`,
  `list_onboard_rejection_ids`, `count_onboard_rejections`).
- `src/aelfrice/classification.py` — `start_onboard_session` and
  `check_onboard_candidates` gain a `force: bool = False` kwarg, bucket
  candidates into `already_present / already_rejected / new`, and
  return a new `n_already_rejected` field on both result dataclasses
  (default 0, additive for existing callers). `accept_classifications`
  writes a ledger row on `persist=False` and deletes any prior row on
  `persist=True` so a rejected-then-accepted sentence leaves the ledger.
- `src/aelfrice/cli.py` — `aelf onboard --force` flag; `--check` output
  gains an `already rejected:` line; `--emit-candidates` JSON gains
  `n_already_rejected`.
- `tests/test_onboard_rejection_ledger.py` — 16 new tests covering
  store CRUD, accept-time ledger writes/deletes, emit + check filter
  behaviour, the `force=True` bypass, and the recorded repro.
- `tests/test_cli_onboard_handshake.py` — 4 new CLI cases for the
  flag and the new output fields.
- `CHANGELOG.md` — Unreleased / Fixed entry.

## Design notes

`belief_id` derivation is shared between the rejection ledger and the
existing dedup path (`_derive_belief_id(text, source)` =
`sha256(source\0text)[:16]`), so the filter key matches the write key
without waiting for `run_worker` to assign a canonical id. The table
is per-project (lives in the same `brain.db` as `beliefs`), so the
"per-project rejection memory" semantics fall out for free.

`--force` only re-opens the noise lane. Already-present beliefs stay
filtered, by design — the issue's `--force` proposal is about
re-rolling a prior classification verdict, not duplicating stored
content. Verified by a dedicated test
(`test_already_present_still_filtered_under_force`).

## Test plan

- [x] `uv run pytest -x -q` → 4183 passed, 62 skipped, 75 xfailed.
- [x] End-to-end smoke: reject all → re-emit returns 0 sentences; reject
      all → re-emit `--force` returns the original N; reject then
      `--force` accept → ledger drops to 0.
- [x] Discretion grep on full diff vs `github/main` clean.

## Alternatives considered

- **Reuse `ingest_log`** with a `rejected=1` flag. Rejected sentences
  are not ingests — they were never written to the belief substrate —
  so a separate table keeps the ingest-log audit trail clean.
- **Project-level "already onboarded" flag** (the issue's first
  alternative). Rejected because it blocks the legitimate "I added new
  docs, re-onboard the directory" case unless `--force` is used; the
  ledger handles incremental onboarding naturally.
- **Classifier-version tracking on rejection rows.** Considered, then
  dropped — `--force` already covers "I want to re-roll because the
  prior pass was too aggressive". Adding a version column means
  arguing about what counts as a version change; the simple flag is
  enough.

## Out of scope

- No migration backfill — the table starts empty on existing stores
  and fills as new rejections happen. No retroactive promotion of
  already-rejected content from prior runs (those are lost; they will
  be re-rejected on the next pass and then stick).
- Rejection-ledger GC / expiry. The ledger grows monotonically on
  the rejected lane until `--force` accepts a row. If this becomes a
  size issue, a follow-up can add `aelf onboard --clear-rejections`
  or an age-based GC. Today the bound is "everything that ever got a
  `persist=False` verdict", which is finite per repo.


robotrocketscience:
[claim:review:sankara:2026-05-14T18:52:27Z]

robotrocketscience:
LGTM. Six atomic signed commits, all CI green, mergeable, discretion-clean, 24 new tests including the recorded repro (pass-2 → 0 candidates after rejecting pass-1). Ledger keyed on the same `sha256(source\0text)[:16]` as `beliefs.id` so the filter matches writes without round-tripping through `run_worker`. Forward-compat via `CREATE TABLE IF NOT EXISTS`, additive dataclass fields default to 0, and the rejected-then-accepted-under-`--force` path deletes the stale row so the ledger doesn't accumulate ghosts.

Two non-blocking observations, both follow-up-shaped:

1. **`--force` is silently no-op on the top-level `aelf onboard <path>` entry.** The flag is only consulted in `_cmd_onboard_emit_candidates` / `_cmd_onboard_check`. If the user runs `aelf onboard . --force` without `--check` or `--emit-candidates`, the call falls through to `_run_regex_onboard` / `_run_llm_onboard` and `--force` is ignored. Help text already says "Apply on the --emit-candidates or --check entry", so it's at least documented, but an `argparse` warning (or routing `--force` through `scan_repo` too) would be a nicer UX. Out of scope here.

2. **Same bug pattern survives in `scan_repo`.** The in-process LLM path (`aelf onboard . --llm-classify`, no handshake) hits `_run_llm_onboard` → `scan_repo`, whose `skipped_non_persisting` branches (`src/aelfrice/scanner.py:268-271, 317`) drop `persist=False` verdicts on the floor with no ledger write. A second LLM-classify run will re-classify the same rejects. The PR's framing is "the host classifier" (handshake terminology), and the recorded repro is Haiku-subagents = handshake, so the scope is intentional — but the same fix shape (`store.insert_onboard_rejection` on `not route.persist`, and consult the ledger in `scan_repo`'s candidate loop) would close the LLM-classify side too. Follow-up issue worth filing if the in-process LLM path is still a supported surface.

Approving and adding `ready-to-merge`.

robotrocketscience:
[release:review:sankara:2026-05-14T18:55:05Z]

PR #795, opened 2026-05-14
feat(ingest): speaker-attribution gate excludes assistant rows from belief creation (#785 §1)

robotrocketscience:
## Summary

Implements §1 (speaker-attribution gate) of the spec landed in #786 (`docs/feature-ingest-speaker-gate.md`). `ingest_jsonl` now excludes `role=="assistant"` rows from belief creation while keeping them in the read path so the rebuilder / hot-start fixture (#592) still sees them.

### Architecture deviation from spec letter

Spec §1 attributes the gate to `transcript_logger.py`. The gate lives in `ingest.py` instead — gating at write-time would strip assistant rows from `turns.jsonl`, breaking the #592 hot-start replay path which depends on assistant rows being present in the JSONL. Putting the gate at the read boundary (`ingest_jsonl`) is the architecturally correct site for "boundary between hook-payload reception and the spawned ingest call" and matches the test phrasing "only user-role content reaches the ingest entrypoint." `_ingest_turn_ids` is the entrypoint; the gate sits immediately before it.

### Mechanics

- `_normalize_jsonl_turn` now returns `role` so the caller can apply the gate.
- `ingest_jsonl` bumps `skipped_lines` and `continue`s on assistant rows.
- `last_per_session` keeps tracking the prior USER turn's last belief, so `DERIVED_FROM` edges chain user→user across any intervening assistant rows with the prior user's text as anchor.

### Out of scope

- §2 (sentiment → `feedback_history` routing): needs cross-hook coordination on the hook surface — separate follow-up.
- §3 (`MIN_BELIEF_CONTENT_CHARS = 80` floor + edge-anchor demotion): separate follow-up.
- Back-purge of pre-existing assistant-role belief rows: stratum-aware cleanup campaign per spec Non-decisions.

## Test plan

- [x] Two new direct tests for §1 in `test_ingest_jsonl.py`:
  - `test_transcript_logger_skips_assistant_role_for_belief_creation`
  - `test_transcript_logger_assistant_role_edges_still_construct`
- [x] Three existing user+assistant tests in `test_ingest_jsonl.py` updated to user+user (the assistant rows no longer contribute beliefs).
- [x] `test_ingest_claude_code_v2_content_array` swapped to `type=user` — the v2 content-array decoder is shared across roles, so this exercises the decoder, not the role gate.
[...]

PR #786, opened 2026-05-14
docs(ingest): spec for speaker-gate + sentiment-routing + SVO floor (#785)

robotrocketscience:
## Summary

Spec for #785 — closes the narration-bloat feedback loop in belief ingest with three coordinated subtractive changes:

- **§1 Speaker-attribution gate** (`transcript_logger.py`) — skip `role == "assistant"` messages for belief creation; edge construction unchanged.
- **§2 Sentiment → `feedback_history` routing** (`hook.py` ↔ ingest) — single source of truth at the hook; ingest reads a feedback-only mark and skips belief creation.
- **§3 SVO min-length floor + edge-anchor demotion** (`ingest.py`) — `MIN_BELIEF_CONTENT_CHARS = 80`; sub-floor candidates demote to edge `anchor_text` or are rejected.

Determinism (#605) preserved. No retrieval-side or scoring-side change. No back-purge of existing rows (deferred to a separate stratum-aware cleanup campaign once these gates land).

## Test plan

- [x] Discretion grep clean on the spec.
- [ ] Spec review — operator verifies the three contracts match intent; implementer reads it as a buildable surface before the implementation PR.
- [ ] Implementation PR (separate, closes #785) lands the six unit tests called out in §§1–3.

Tracks #785.

## Summary by CodeRabbit

* **Documentation**
  * Added a feature specification detailing ingest-side changes to optimize message ingestion: refined handling of archived assistant messages, sentiment-first feedback routing to feedback history, raised minimum content threshold for belief creation, a determinism guarantee, explicit non-decisions, an acceptance checklist, and accompanying unit-test plans.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/786)

Issue #785, opened 2026-05-14
fix(ingest): close narration-bloat feedback loop — speaker-gate + sentiment-routing + SVO floor


robotrocketscience:
## Summary

Eliminate the self-reinforcing narration-bloat feedback loop in belief ingest. Three coordinated additions across `transcript_logger.py`, `ingest.py`, and `hook.py` block the three input sources that account for ~100% of short, reinforced beliefs in the most-retrieved stratum.

## The feedback loop

```
agent narrates progress
    ↓
transcript_logger ingest stores narration as a belief row
    ↓
next user turn says "good" / "yeah keep going"
    ↓
sentiment hook (#606) bumps α on the recently-retrieved narration belief
    ↓
narration posterior rises; BM25 ranks it higher next turn
    ↓
agent sees its own old narration as "context" and continues the pattern
```

Empirical attribution of short reinforced beliefs in the α+β ≥ 10 stratum:

- **51% agent self-narration** (status lines, tool-completion messages, pytest output echoes).
- **30% user direction / sentiment** ("yeah keep going", "good", "no that's wrong") — should have routed to `feedback_history`, not `beliefs`.
- **19% prose fragments** (headers, code-fence boundaries, bullet stubs) emitted by the SVO-triple extractor below its useful length floor.

### 1. Speaker-attribution gate (`transcript_logger.py`)

At the boundary between hook-payload reception and the spawned ingest call, inspect each archived message and **skip messages with `role == "assistant"` for belief creation**. Assistant content may still feed edge construction (citation / reference links between user-turn beliefs) but must not become standalone belief rows.

Out of scope: retroactive deletion of pre-existing assistant-role belief rows. The back-purge is a separate stratum-aware campaign once the upstream gate lands (see "Out of scope" below).

### 2. Sentiment → `feedback_history` routing (`hook.py` ↔ ingest)

When the UserPromptSubmit's text matches a sentiment pattern *and* the same text would have become a belief via the post-compaction ingest path, suppress belief creation. The sentiment match continues to bump `feedback_history` on prior beliefs via the #606 hook's normal path — that is the load-bearing behavior, not belief creation.

Two structural options:

- **A. Sentiment-first.** Sentiment classifier runs before ingest archives the user message. On match, mark the archived entry as feedback-only. Ingest reads the mark and skips belief creation. Classifier stays authoritative in one place.
[...]

Issue #798, opened 2026-05-14
A4 bench gate is structurally vacuous — rebuilder uses verbatim token cost, ignoring compression flag


robotrocketscience:
## Summary

The A4 fidelity bench gate shipped via #775 / PR #776
(`tests/bench_gate/test_compression_a4_fidelity.py` +
`tests/retrieve_uplift_runner.py::run_compression_a4_fidelity`) is
**structurally vacuous**. The rebuild block it scores under
`use_type_aware_compression={OFF, ON}` is byte-identical between the
two arms in every row of the seed corpus tested, so any proxy (the
shipped token-coverage one, an embedding scorer, or an LLM judge)
gives delta = 0 — not because compression doesn't help fidelity, but
because the rebuilder's pack accounting strips the compression effect
out before any belief content reaches the rebuild block.

The #769 flip-default decision's A4 axis is therefore load-bearing on
a gate that cannot measure compression. Filing here because the A2
axis works fine — the architectural mismatch is rebuilder-side, not
corpus-side.

## Direct evidence

`retrieve()` and `rebuild_v14()` called on the same belief population,
same transcript-derived query (694 chars, the full pre-clear turns of
a representative A4 row), same `token_budget`, across two arms set via
`AELFRICE_TYPE_AWARE_COMPRESSION={0,1}`:

```
budget=100  retrieve() OFF: ids=[]
            retrieve() ON:  ids=[snap-001, snap-005, tran-004, snap-003]
            rebuild_v14 OFF/ON: ids=[]                                   (identical)

budget=150  retrieve() OFF: ids=[snap-001]
            retrieve() ON:  ids=[snap-001, snap-005, tran-004, snap-003, fact-004, fact-001]
            rebuild_v14 OFF/ON: ids=[snap-001]                           (identical)

budget=250  retrieve() OFF: ids=[snap-001, snap-005]
            retrieve() ON:  ids=[snap-001, snap-005, tran-004, snap-003, fact-004, fact-001]
            rebuild_v14 OFF/ON: ids=[snap-001, snap-005]                 (identical)

budget=1000 retrieve() OFF/ON: ids=[fact-001, fact-004, snap-001, snap-003, snap-005, tran-004]
            rebuild_v14 OFF/ON: same                                     (identical)
```

`retrieve()` honors compression — the candidate set differs per arm.
`rebuild_v14()` discards that difference.

Confirmed for every row of the v0_1 + v0_2 A4 seed corpus
[...]

PR #810, opened 2026-05-14
feat(ingest): subfloor noise-pattern filter + intra-turn edge-anchor demotion (#809)

### LINKED ISSUE #809: feat(ingest): SVO-extractor min-length floor + edge-anchor demotion (#785 §3)
Implements §3 of the narration-bloat spec already in `docs/feature-ingest-speaker-gate.md`. §1 (speaker-attribution gate) shipped under PR #795. §2 (sentiment → `feedback_history` routing) is **on HOLD** per the `alpha-gain-third-path` campaign verdict — not load-bearing for the top-stratum leak. §3 is the remaining ADOPT component flagged "separate follow-up" in PR #795's body.

## What §3 does

Stops the SVO-triple extractor from minting freestanding belief rows out of sub-floor prose fragments (code-fence prefixes, bullet stubs, section headers ending in `:`, tool-output echoes). The retrieval-corpus-bloat campaign attributed **19% of short reinforced beliefs** in the α+β ≥ 10 stratum to this class.

## Contract (from spec § 3)

```python
MIN_BELIEF_CONTENT_CHARS: Final[int] = 80
```

For each candidate belief produced from a triple's subject or object slot:

- `len(candidate.content.strip()) >= 80` → becomes a belief row as before.
- `len(candidate.content.strip()) < 80` → the candidate does NOT become a freestanding belief row. The clause attaches as `anchor_text` on an edge between the surrounding full-length beliefs.
- If neither slot resolves to a ≥ 80-char belief, the entire triple is rejected. Unanchored sub-floor prose has no graph place.

The floor is a module constant, not a config knob — the spec was explicit that the value should move only by re-measurement.

## Acceptance

- [ ] `MIN_BELIEF_CONTENT_CHARS = 80` added as a `Final[int]` module constant in `src/aelfrice/ingest.py`.
- [ ] Triple-emission path applies the floor with edge-anchor demotion per the contract above.
- [ ] Test `test_ingest_minlen_demotes_short_to_edge_anchor` — paragraph whose triple-extraction produces `(headerA, "→", bulletStart)` where `bulletStart` is sub-floor; assert no new belief row created; edge between two full-length parents carries the clause as `anchor_text`.
- [ ] Test `test_ingest_minlen_rejects_unanchored_triple` — sub-floor content in both subject and object slots; assert no belief rows created, no edges created, triple silently dropped.
- [ ] Test `test_ingest_above_floor_unchanged` — full-length content; assert belief row created as before, edge construction unchanged.
- [ ] Determinism property (#605) preserved — same payload → same belief set byte-for-byte.

## Out of scope

- **No back-purge of existing rows.** Spec § "Non-decisions" — blind purge is wrong, separate stratum-aware cleanup campaign should run after §3 is in place upstream so cleanup doesn't refill. Tracking under a separate issue if/when filed.
- **No ambiguity classifier** for the "no yeah that makes sense" case (0/100 in the analyzed sample, deferred until production data post-fix shows it matters).

## Refs

- Spec: `docs/feature-ingest-speaker-gate.md` § 3 (lines 102-133).
- §1 shipped: PR #795 (`feat(ingest): speaker-attribution gate excludes assistant rows from belief creation (#785 §1)`), merged 2026-05-14 16:44Z.
- §2 HOLD rationale: `alpha-gain-third-path` campaign verdict (lab side) — sentiment routing not load-bearing for top-stratum leak.
- Empirical basis for `MIN_BELIEF_CONTENT_CHARS = 80`: `retrieval-corpus-bloat` campaign R0/R2 — short-reinforced stratum bottoms out around 80 chars; below that, content is overwhelmingly noise.

robotrocketscience:
## Summary

Closes #809. Adds a pattern-based subfloor-noise gate at the sentence-level ingest path (`_ingest_turn_ids`). Sentences matching the named noise patterns from the `retrieval-corpus-bloat` lab campaign — code-fence boundaries, header stubs ending with `:`, markdown bullet stubs — no longer become freestanding belief rows. When a matched sentence sits between two full-length-belief sentences within the same turn, it attaches as `anchor_text` on a new intra-turn `DERIVED_FROM` edge between the surrounding beliefs (src=later, dst=earlier, matching the inter-turn convention in `ingest_jsonl`). Unanchored matches are silently dropped.

This closes 19% of the short-reinforced-bloat leak documented in `retrieval-corpus-bloat` R0/R2. Companion to PR #795's §1 speaker-attribution gate (51% of the same leak). Together they close ~70% of the empirically-measured input-side leak.

## Operator-ratified scope deviations

Two deviations from the spec letter in `docs/feature-ingest-speaker-gate.md` § 3, both flagged + approved during implementation:

### 1. Pattern-gate, not length-floor

Spec literal: `MIN_BELIEF_CONTENT_CHARS: Final[int] = 80`. Applied as a strict floor, this dropped 26 existing ingest-test fixtures (legitimate short factual claims: `"The configuration file lives at /etc/aelfrice/conf."`, `"The default port is 8080."`). The lab campaign empirically calibrated 80 chars against the **α+β ≥ 10 reinforced stratum** — content already bumped repeatedly — not against all incoming sentences. A length-floor read of the spec over-applies that finding.

Operator-ratified scope is pattern-based: only the three pattern classes the lab named (code-fence, `:`-header, bullet stub) trigger the gate. Short legit claims pass. The acknowledged false-positive surface is `"He said:"`-style real prose ending in `:`; the lab named this pattern explicitly, trade-off accepted at empirical scope.

### 2. Gate lives in `ingest._ingest_turn_ids`, not `triple_extractor`

Spec letter: gate the "triple-extraction emission path" at "subject or object slots". But `triple_extractor.py` builds Triples from noun phrases (`_NP` = 1-5 tokens, ~10-40 chars typical) — applying ANY length floor or content-pattern check there would over-fire to near-zero.

The observable leak the campaign measured comes from the sentence-level path (`_ingest_turn_ids` via `extract_sentences`). That's where the gate lives. "Edge-anchor demotion" maps to intra-turn `DERIVED_FROM` edges between consecutive full-length sentences in the same turn — the natural codebase analogue of the spec's "surrounding full-length beliefs". Mirrors PR #795's deviation for §1 (moved gate from `transcript_logger` to `ingest_jsonl` for the same architectural reason).

## Files changed

- `src/aelfrice/ingest.py` — new `_looks_like_subfloor_noise(sentence)` helper checks for code-fence prefix / `:`-ending / markdown bullet marker. `_ingest_turn_ids` partitions sentences into full-length belief candidates and pending-demotion sub-floor clauses, then wires intra-turn DERIVED_FROM edges with `anchor_text = " | ".join(between)[:ANCHOR_TEXT_MAX_LEN]`. Edge insert is deduped via `get_edge(...)` check; re-ingest is idempotent on both beliefs AND on intra-turn edges.
- `tests/test_ingest_subfloor_noise.py` — 16 new tests across three layers:
  - Unit checks on `_looks_like_subfloor_noise` (all three patterns; legit-claim non-firing; whitespace strip; empty input).
  - End-to-end filter via `_ingest_turn_ids` (header-alone, all-subfloor-turn, unanchored-at-start, unanchored-at-end — the silently-dropped paths).
  - End-to-end demotion (header between two full sentences → edge anchor_text; multiple sub-floor clauses joined; three-full-sentence chain with sub-floor between each consecutive pair).
  - Negative-control regressions (legit short claim still ingests; consecutive full sentences produce no spurious intra-turn edge; re-ingest is idempotent on beliefs AND on intra-turn edges).
- `CHANGELOG.md` — Unreleased / Fixed entry.

## Pattern coverage relative to `extract_sentences`

`extract_sentences` already handles two of the three pattern classes upstream:

- **Code-fence regions** — stripped wholesale by paired-fence regex.
- **Line-leading list markers** — stripped from start of each line.
- **Headers ending in `:`** — NOT stripped; reach the gate as-is when newline-separated.

The gate retains all three pattern checks as defense-in-depth for malformed / mid-line cases that survive `extract_sentences` (unclosed code-fences, bullets that appear after `.` in a single line). Unit tests cover all three. End-to-end integration tests cover the `:` pattern (the load-bearing one in the normal pipeline). Test fixtures separate `:`-headers by `\n\n` because `:` is not a sentence boundary in `extract_sentences`.

## Test plan

- [x] `uv run pytest -x -q` → **4202 passed, 62 skipped, 75 xfailed**. No existing ingest test broken.
- [x] Discretion grep on full diff vs `github/main` — clean (verified by `aelf-pr-open.sh` pre-flight).
- [x] All three commits SSH-signed (verified locally `%G? = G`).

## Out of scope

- **§2 sentiment → `feedback_history` routing** — held per the lab-side `alpha-gain-third-path` campaign verdict ("not load-bearing for the top-stratum leak"). If/when re-prioritized, a separate follow-up.
- **No back-purge of existing rows.** Stratum-aware cleanup of the pre-existing short-reinforced corpus is a separate campaign per the spec's "Non-decisions" section — should run *after* §1+§3 are in place upstream so cleanup doesn't refill.
- **No retrieval-side top-K cap change.** Originally proposed in an earlier conversation; out of scope here. Fix the input side first, re-measure top-K on the cleaner data, then decide.

## Refs

- `docs/feature-ingest-speaker-gate.md` § 3 — spec.
- #785 — umbrella (closed; spec + §1 shipped under PR #786 + PR #795).
- PR #795 — §1 speaker-attribution gate (shipped 2026-05-14 16:44Z).
- `retrieval-corpus-bloat` R0/R2 — lab campaign empirical basis (2026-05-11, lab-side).
- `alpha-gain-third-path` — verdict that holds §2 (lab-side).

Closes #809

## Summary by CodeRabbit

* **New Features**
  * Enhanced ingest process to automatically filter short text fragments resembling formatting noise—such as code markers, incomplete headers, and bullet stubs—while preserving legitimate short claims. Fragments between full-length beliefs are now linked as relationship anchors.

* **Tests**
  * Added comprehensive test suite validating noise filtering and belief linking behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/810)


robotrocketscience:
[claim:review:chomsky:2026-05-14T20:06:31Z]

robotrocketscience:
## Review: code APPROVED; deferring `ready-to-merge` for operator-eyeball

### What I verified
- **Imports correct.** `ANCHOR_TEXT_MAX_LEN`, `EDGE_DERIVED_FROM`, `Edge`, `re` all resolved.
- **Edge direction matches inter-turn convention** — `ingest.py:344-346` (`ingest_jsonl`) writes `src=head_id, dst=prior_id, type=EDGE_DERIVED_FROM, weight=1.0, anchor_text=truncated`. New intra-turn code uses the same shape. ✓
- **Single-id refactor is semantically equivalent.** `derivation_worker.py:316-320` writes `derived_belief_ids=[actual_id]` (always one element). Switching from "iterate all ids" → "take ids[0]" preserves the public return contract.
- **Idempotency** — `store.get_edge(curr_bid, prior_bid, EDGE_DERIVED_FROM) is not None` guard + `prior_bid == curr_bid` short-circuit handle re-ingest and self-edge cases. `test_ingest_idempotent_on_repeat_under_demotion` exercises both beliefs and edges.
- **`extract_sentences` backstop reasoning checks out.** Paired-fence regions and line-leading list markers already stripped upstream; `:`-header is the only load-bearing pattern in the normal pipeline, and end-to-end tests exercise that one. Helper retains all three as defense-in-depth for malformed input.
- **Tests** — 16 tests, three layers (unit / filter / demotion + negative controls). Covers the three named patterns, all silently-dropped paths, anchor concatenation, three-sentence chains, and the false-positive trap (length-floor would have dropped `"The default port is 8080."`; pattern-gate keeps it).
- **Discretion grep on full diff** — clean. All three commits SSH-signed (`%G? = G`).
- **CI** — pytest 3.12/3.13, CodeQL python+actions, bench-smoke, pattern-scan, secrets-scan, history-scan all green. Replay Soak Gate green. Sourcery / CodeRabbit skipped due to rate-limit, not failure.

### Deviation magnitude — flagging for explicit ratification

The PR characterises both scope deviations as "operator-ratified" but neither #809 nor the #785 umbrella has an explicit ratification comment that I could find. Precedent for "deviate from spec letter and ship" exists — PR #795 (§1) moved the gate from `transcript_logger.py` to `ingest.py` and you merged it at 16:44Z today — so this PR isn't unprecedented. But the §3 deviation is materially larger:

| | Spec letter | This PR |
|---|---|---|
| Mechanism | `MIN_BELIEF_CONTENT_CHARS: Final[int] = 80` length floor on triple S/O slots | Three pattern matches (code-fence prefix, `:` suffix, bullet prefix) at sentence-level path |
| Drop set | All sentences < 80 chars | Only sentences matching one of three patterns |
| False-positive surface | Drops some legit short claims | Drops `"He said:"` / any prose ending in `:` |
| Layer | `triple_extractor` | `ingest._ingest_turn_ids` |

The two layers don't produce the same dropped set. Engineering rationale for the divergence is sound on both axes:
1. A literal 80-char floor would drop the existing fixture `"The default port is 8080."` (25 chars) and ~26 other ingest-test fixtures — the spec letter is incompatible with the existing test contract.
2. `triple_extractor`'s NP slots are 10-40 chars typical; any length floor there would zero-fire.

But the deviation is large enough that I'd rather you OK it explicitly before the merge-train picks it up. If you're happy with the pattern-gate framing, drop a `ratify` comment (or just add `ready-to-merge` yourself) and the bot will take it from there.

### Non-blocking observations
- **`endswith(':')` false-positive scope** — flagged in code + body. Lab campaign named the pattern, trade-off accepted. The fact that `extract_sentences` doesn't split on `:` means this fires mainly when a header is newline-separated, which limits real-prose damage in practice. Worth re-measuring once production data accumulates.
- **`#809` audit-bot comment** is a false-positive — it flagged `src/aelfrice/ingest.py` as already-shipped because #795/#786 also touched it, but the §3 surface is genuinely new. No action needed.
- **PR is over the 200-line soft cap** (392 lines / 3 files). Advisory only. Splitting would just be churn at this point.

Approving the code. Holding off on `ready-to-merge` until you ratify the deviation framing.

robotrocketscience:
[release:review:chomsky:2026-05-14T20:11:07Z]

robotrocketscience:
## Recommended tightening before merge

I ran an empirical audit of `_looks_like_subfloor_noise` against the
real ingest corpus to characterize what the three patterns actually
catch and what they drop unnecessarily. Recommendation below.

### Findings

On the existing belief store (~13k rows total, 901 of them end in
`:`):

| Pattern | Real-corpus hits | False-positive risk |
|---|---:|---|
| code-fence prefix | low | none observed |
| bullet prefix | **zero** | none observed (`extract_sentences` strips upstream) |
| `:` suffix | 901 | **152 rows ≥ 80 chars are long-form prose** ending in `:` (e.g. `"If you look at the way the rebuilder picks beliefs, the order is always the same:"` — 81 chars). 109 of those 152 contain English verbs. PR #810 as-is drops all of them. |

Short stubs dominate the `:`-suffix surface (60% are < 40 chars and
have no verb — definite header stubs). The 17% long-form tail is the
load-bearing FP class.

Length distribution of `:`-suffix beliefs in the store:

```
  <20:  19.3%       60-79: 10.3%
20-39:  33.2%       80-119: 8.6%
40-59:  20.3%      120-199: 6.4%   >=200: 1.9%
```

### Recommended change

Add the spec's own constant `MIN_BELIEF_CONTENT_CHARS = 80` to the
`:` rule as an upper bound on what triggers the gate:

```python
_SUBFLOOR_COLON_MAX_LEN: Final[int] = 80

def _looks_like_subfloor_noise(sentence: str) -> bool:
    stripped = sentence.strip()
    if not stripped:
        return False
    if stripped.startswith("```"):
        return True
    if (stripped.endswith(":")
            and len(stripped) < _SUBFLOOR_COLON_MAX_LEN):
        return True
    if _SUBFLOOR_BULLET_PREFIX.match(stripped):
        return True
    return False
```

Effects vs the PR as it stands:

- Drops 83% of `:`-suffix beliefs (749/901) — header stubs survive
  the change.
- Saves 152 long-form `:`-ending statements (109 of which read as
  real prose).
- Bullet and code-fence rules unchanged.

### Why `80` specifically

It's the spec literal (`MIN_BELIEF_CONTENT_CHARS = 80` in
`docs/feature-ingest-speaker-gate.md` §3). I tested 40/60/80/100/120
on the real distribution:

- N=60 leaves only 245 rows alive — drops most legit long-form prose.
- N=80 leaves 152 alive, including 109 verb-y prose rows — clean
  break.
- N=100 leaves only 96 alive — too aggressive on the long-form set.

A synthetic probe of 17 hand-constructed sentences ending in `:`
confirms the cut: under the `< 80` floor, all 3 long-form prose
constructs (81–111 chars) survive; all 3 mid_prose (54–61 chars) and
all 10 ambiguous_short still drop, matching the gate's intent.

### Suggested unit tests to pin the boundary

```python
def test_helper_does_not_flag_long_form_prose_ending_with_colon():
    """81 chars — just past the cut. A long sentence ending in `:`
    can be a legit prose statement (e.g. introducing a code block
    that's already extract_sentences-stripped, leaving the prose)."""
    assert _looks_like_subfloor_noise(
        "If you look at the way the rebuilder picks beliefs, the "
        "order is always the same:"
    ) is False


def test_helper_flags_short_header_below_floor():
    """20 chars — header stub, drops as expected."""
    assert _looks_like_subfloor_noise("Acceptance criteria:") is True
```

### Acknowledged remaining FP

A 35-char row like `"The five concepts that earn a name:"` still
drops under the recommended rule. That's borderline — structurally
it's a header stub leading a list, but grammatically a complete
sentence. The cost of saving it is a more elaborate rule
(verb-detection / list-following-context); I'd accept the FP as the
right trade-off, but flagging so you can decide.

### Locked status of `:`-ending beliefs

Zero `:`-suffix beliefs are operator-locked in the current store.
The user's own ground-truth behavior aligns with treating
`:`-suffix as non-load-bearing.

robotrocketscience:
Flipping `attn:decisions-needed` → `ready-to-merge` per operator direction.

For the audit trail: the `_SUBFLOOR_COLON_MAX_LEN = 80` tightening recommendation in the prior comment (152 long-form `:`-ending beliefs, 109 of them real prose, dropped under the rule as it stands) remains **unapplied** — shipping the as-is pattern-gate per operator call. If the FP rate proves problematic against post-merge ingest, the recommendation is a drop-in follow-up (single helper edit, no API change).

robotrocketscience:
Follow-up filed as #818 capturing the `_SUBFLOOR_COLON_MAX_LEN = 80` tightening with chomsky's audit table and the 4-line helper diff. Closes the trail.

robotrocketscience:
## Deeper check on the bullet + code-fence rules

Followup to my earlier comment recommending the `<80` length floor
on the `:`-suffix rule. I went back to question whether the bullet
and code-fence rules in `_looks_like_subfloor_noise` actually need
the same treatment.

**They don't — they're unreachable in this call path.**

Reading `src/aelfrice/extraction.py:extract_sentences`:

- Line 30 strips paired `` ``` ``…`` ``` `` regions wholesale, including the
  language tag.
- Line 59 strips line-leading bullet markers (`^[ \t]*[-*+]\s+`).

Probe results on synthetic inputs:

| input | survives `extract_sentences` as |
|---|---|
| `\`\`\`python\nimport json\n\`\`\`` (paired) | `[]` — wholesale stripped |
| `\`\`\`` (solo, 3 chars) | `[]` — below the 10-char floor in step 7 |
| `\`\`\`python\n…(no closing fence)` (unpaired) | survives as `\`python` (3 chars, also below floor) |
| `- run tests` | `["run tests"]` — bullet stripped, content survives **without** the `-` prefix |
| `- Long-form bullet paragraph (220 chars)` | content survives, **without** the `-` prefix |

The gate's regex (`^[-*+]\s` and `startswith("```")`) requires
exactly the marker that `extract_sentences` has already stripped.
After upstream strip, no surviving sentence can match either rule.

**Implication:** the bullet and code-fence checks in `_looks_like_subfloor_noise`
are dead code in `_ingest_turn_ids`. They don't have FP risk
(they don't fire) but they also don't have value. Two reasonable
options:

1. **Keep as hedge** — leaves a backstop if `extract_sentences`
   ever changes its strip semantics. Cost: a few unreachable lines.
2. **Drop both rules** — honest about what the gate actually does.
   Cost: a future regression in `extract_sentences` could let
   bullet/code-fence stubs through unfiltered. Easy to re-add.

Either is defensible. The only behavior-relevant change for PR #810
is still the `:`-suffix `<80` tightening from my earlier comment.

**A note on the 411 long-form bullet+code-fence rows I'd seen in
the store** (mentioned in case you spot them yourself): those came
in through other ingest paths (predominantly `legacy_unknown`,
which is pre-`_ingest_turn_ids` ingest). PR #810 doesn't touch
those rows; this PR's gate at `_ingest_turn_ids` won't see them
either way.

robotrocketscience:
Applied the `<80` length-floor tightening on top of the existing branch (FF push, no force). Two atomic commits:

- `a928888a` `refactor(ingest): scope subfloor pattern check to <80 chars` — adds `_SUBFLOOR_MAX_LEN: Final[int] = 80`, early return on `len(stripped) >= 80` before the three pattern checks. Updated module + helper docstring to describe the length-scope contract.
- `3f57f029` `test(ingest): boundary tests for the 80-char subfloor length cap` — four new tests pin the boundary (81-char `:`-suffix prose passes; ≥80-char bullet paragraph passes; canonical short stubs still drop; 79-vs-80 boundary). Plus a CHANGELOG note revision.

Local pytest: **4206 passed, 62 skipped, 75 xfailed** (was 4202; +4 matches the new boundary tests). All three commits on the branch are SSH-signed.

Re-flagged `attn:review`. Once CI is green and someone can give it eyes, drop in `ready-to-merge` and the merge-train should take it.

robotrocketscience:
[claim:review:bagheera:2026-05-14T21:24:42Z]

robotrocketscience (commented):
## Reviewed — substantive logic is sound; CI gap blocks merge

**Substantive review (logic + content) — pass:**

- The pattern-gate-+-length-floor combination is well-argued and the empirical evidence cited in [the prior PR comment](https://github.com/robotrocketscience/aelfrice/pull/810#issuecomment-was-by-sankara) (152 `:`-ending rows ≥80 chars on the live store; 263 long-form bullet and 112 long-form code-fence beliefs preserved) is convincing. The `< 80 AND pattern` shape catches the named noise class without the standalone-floor collateral.
- The deeper-check note (bullet + code-fence patterns are unreachable under `extract_sentences` upstream stripping) is correct; keeping them as defense-in-depth is fine.
- Intra-turn `DERIVED_FROM` direction (src=later → dst=earlier) matches the inter-turn convention in `ingest_jsonl`.
- Anchor-text join (`" | ".join(between)[:ANCHOR_TEXT_MAX_LEN]`) and the `get_edge` dedup before insert make re-ingest idempotent on both beliefs and intra-turn edges.
- All five commits SSH-signed (`%G? = G`).
- Operator-ratified scope deviations from the spec letter (pattern-gate vs length-floor; `_ingest_turn_ids` vs `triple_extractor`) are explicitly documented in commit bodies and the PR description.
- Together with PR #795's §1, this closes ~70% of the empirically-measured short-reinforced-bloat leak. Issue #818 is content-superseded by commits `a928888a` + `3f57f029` here.

**Blocker — CI not green on PR head:**

`gh run list --commit 3f57f029 --event pull_request` returns zero rows; same for `a928888a`. Only `label-docs` (push-event, push-paths-filtered) reported on the new HEAD. The full required-check matrix (`pytest (3.12)`, `pytest (3.13)`, `CodeQL`, `Staging Gate / secrets-scan`, `Staging Gate / pii-scan`, `Staging Gate / commit-history-audit`, `Bench Smoke`, `Replay Soak Gate`, `Eval Calibration`, `deadcode`) only ran against the prior HEAD `a5425f30`. The `pull_request` synchronize event didn't fire on the two 21:07Z commits, so the merge-train will refuse (head SHA ≠ labeled-event SHA).

Removing `ready-to-merge` so the merge-train doesn't churn on the stale ref. Suggested re-trigger: push an empty commit to the branch (or close+reopen the PR via the GitHub UI) to force the `pull_request` synchronize event. Once `pytest` + `CodeQL` + `Staging Gate` green on `3f57f029`, re-add `ready-to-merge`.

robotrocketscience:
[claim:review:feynman:2026-05-14T21:26:39Z]

robotrocketscience:
[release:review:bagheera:2026-05-14T21:26:43Z]

robotrocketscience:
[release:review:feynman:2026-05-14T21:26:43Z]

robotrocketscience:
[claim:review:oppenheimer:2026-05-14T21:27:58Z]

robotrocketscience (commented):
Review pass. Substantively sound; the scope deviations (pattern-gate over length-floor, sentence-level over triple-extractor) are well-defended in the PR body and on the `_looks_like_subfloor_noise` docstring. Atomic commit shape is clean. CHANGELOG entry under `[Unreleased] / Fixed` is present (commit a5425f30) and matches the deviations.

Two things to address before this lands, plus one question.

## Blocking: rebase needed

`mergeStateStatus: DIRTY`. The `#820` refactor (demote-path removal) merged at 21:24:44Z; the touch surface there is small but the existing CI history on this PR predates it, so full CI hasn't run on a current-main base — `gh pr checks 810` shows only `label`, `Sourcery review`, `CodeRabbit`. Rebase onto `github/main` (HEAD `fed650ca`), push, and re-run the merge-train label cycle so `pytest (3.12/3.13)`, `calibration`, `pattern-scan`, `history-scan`, etc. land green against current main.

## Question: derived_belief_ids semantics — is the new `ids[0]`-only resolution intended?

The diff at `src/aelfrice/ingest.py` collapses the old "iterate `derived_belief_ids` and accumulate every new id" loop into "take `ids[0]` only, then accumulate that one." See:

**Before** (github/main):
```python
for log_id in log_ids:
    entry = store.get_ingest_log_entry(log_id)
    if entry is None:
        continue
    ids = entry.get("derived_belief_ids") or []
    if not isinstance(ids, list):
        continue
    for bid in ids:
        if (isinstance(bid, str) and bid not in ids_before
                and bid not in seen):
            seen.add(bid)
            inserted.append(bid)
```

**After** (this PR):
```python
for log_id in log_ids:
    entry = store.get_ingest_log_entry(log_id)
    bid: str | None = None
    if entry is not None:
        ids = entry.get("derived_belief_ids") or []
        if isinstance(ids, list) and ids:
            head = ids[0]
            if isinstance(head, str):
                bid = head
    log_belief_ids.append(bid)
    if bid is not None and bid not in ids_before and bid not in seen:
        seen.add(bid)
        inserted.append(bid)
```

The doc-string on `_ingest_turn_ids` says "the **per-sentence** derived belief id (in input order, with duplicates dropped)" — singular — so reading only `ids[0]` is consistent with the contract as written. But the underlying schema column `derived_belief_ids` is a JSON array, and at least two test fixtures encode it as multi-element (`tests/test_ingest_log.py:140` writes `["b1", "b2"]`, `:227` writes `["b-1", "b-2"]`). Those are schema round-trip tests, not `_ingest_turn_ids` exercisers, so they don't directly catch a regression — but they document that the storage layer supports multi-id, and the previous `_ingest_turn_ids` loop honoured that.

Two questions:

1. Does `run_worker` ever emit a log row with multiple `derived_belief_ids` (e.g., one sentence producing multiple beliefs via the derivation pipeline)? If yes, this PR silently drops all but the first from the `inserted` return — which feeds the public `ingest_turn` return value.

2. If the contract really is per-sentence-singular, the ingest_log schema is over-typed and could be tightened in a follow-up (separate issue) — but for *this* PR, an explicit assertion or test case ("multi-id derivation rows are unreachable from `_ingest_turn_ids`" / "`run_worker` is single-id per sentence") would lock the assumption in.

If the answer to (1) is "no, never," a one-line comment on the `bid: str | None = None` resolution would prevent the next reader from asking the same question.

## Minor nits (non-blocking)

- `_SUBFLOOR_BULLET_PREFIX = re.compile(r"^[-*+]\s")` — fine for the listed three bullet markers. Numbered lists (`1.`, `2.`) and quote markers (`>`) are not covered; if those classes ever surface in the corpus-bloat data, follow up.
- The acknowledged FP surface (`"He said:"`, `"Note:"`) plus the 80-char scoping is exactly the right trade-off given empirical data. Worth a sentence in `docs/feature-ingest-speaker-gate.md` § 3 (the spec doc) saying the implementation realized the spec via pattern+length-cap rather than literal length-floor, with a pointer to the helper docstring — so the next reader of the spec doesn't re-derive the divergence from scratch. Follow-up issue is fine if it crowds this PR.

## Net

LGTM after rebase + a clarifying note (or confirmation) on the `ids[0]` resolution. Holding the formal approve until I see the rebased branch + the multi-id answer.

robotrocketscience:
[release:review:oppenheimer:2026-05-14T21:29:50Z]

robotrocketscience:
Rebased onto current `main` (post-#798) and pushed `4eb7df4f`. Branch is FF on `main`; all 5 commits SSH-signed `G`. Closed/reopened to refire the synchronize event since the prior push out of a detached-HEAD worktree didn't trigger pytest/CodeQL.

Answers to your `ids[0]` question:

**(1) "Does `run_worker` ever emit a log row with multiple `derived_belief_ids`?"** No — `src/aelfrice/derivation_worker.py:318` is the only writer for `derived_belief_ids` on rows downstream of `_ingest_turn_ids`, and it always writes `derived_belief_ids=[actual_id]` (single-element list, with `actual_id` being the canonical/corroborated id from `get_or_insert_belief`). So `ids[0]`-only resolution is semantically equivalent to the prior "iterate all" loop in the actual call path, just expressed in a way that's easier to reason about per-sentence.

**(2) "Should the contract be documented?"** Agree. Adding an inline comment on the `bid: str | None = None` resolution explaining the single-id invariant would prevent the next reader from asking the same question. Will add it as a third commit on this branch if you want; happy to defer to a follow-up doc PR.

The multi-id test fixtures you found (`test_ingest_log.py:140`, `:227`) are exercising the storage layer's JSON round-trip, not the `_ingest_turn_ids` contract, so they're not regressed by this change. The schema staying multi-id supports a hypothetical future writer (a triple-extractor that emits multiple beliefs per ingest_log row) without a migration; that's defensible posture for a column shape.

**Nits acknowledged:**

- Numbered list / quote-marker patterns (`1.`, `>`): out of scope for this PR; not surfaced in the empirical corpus audit. If they show up later, the addition is mechanical.
- Spec doc reconciliation (`docs/feature-ingest-speaker-gate.md` §3): yes, that's worth a follow-up `docs(spec): reconcile §3 with shipped pattern+length-cap implementation` PR. Not blocking this one.

Awaiting full CI on the new SHA.

robotrocketscience:
[claim:review:bagheera:2026-05-14T21:36:09Z]

robotrocketscience:
[claim:review:clarke:2026-05-14T21:37:18Z]

robotrocketscience:
[release:review:clarke:2026-05-14T21:37:22Z]

robotrocketscience (commented):
## Independent review — rebase landed; multi-id audit answers owner's question

Picking this up after the prior owner-review pass (21:26Z + 21:29Z) flagged two blockers — rebase + the `derived_belief_ids` collapse semantics. Both are now addressable.

### 1. Rebase status — done

Branch tip `4eb7df4f` is now based directly on `fed650ca` (current `github/main` HEAD, post-#820). `git log --oneline github/feat/issue-809-svo-minlen-floor` shows the five PR commits sitting directly on `fed650ca` with no merge commit. The owner's prior review was against `3f57f029` (the pre-rebase tip); the rebase happened between then and now.

CI re-fired on the new HEAD and is green on the staging-gate matrix (`secrets-scan`, `pattern-scan`, `history-scan`, `release-docs-check`, `commit-msg-prefix`, `pr-title-prefix`, `pr-body-issue-link`), `deptry`, `vulture`, `bench-smoke`, `consecutive-green ≥ 7d`, `typos`, CodeQL (python + actions), `size-check`. `pytest (3.12)` and `pytest (3.13)` are still in-flight from runtime `25886927066`; `merge-train` is correctly SKIPPED while `ready-to-merge` is off.

### 2. `derived_belief_ids` multi-id audit — production is single-id by construction

The owner asked whether `run_worker` ever emits a log row with multiple `derived_belief_ids`, since the `ids[0]`-only collapse silently drops the tail if so. Grepping every production writer of that column in `src/`:

- `derivation_worker.py:257` — `store.update_ingest_derived_ids(log_id, derived_belief_ids=[])` (orphan / no-derive case, empty array).
- `derivation_worker.py:316-318` — `store.update_ingest_derived_ids(log_id, derived_belief_ids=[actual_id])` (the canonical "successfully derived one belief" case — explicit single-element list).
- `store.py:1124-1131` — legacy-log-synth migration (#263), inserts one row per orphan belief with `derived_belief_ids = [belief.id]` (single-element).

No production code path writes a multi-element `derived_belief_ids` array. The contract is "0 or 1 derived belief ids per ingest_log row." The previous `for bid in ids` loop in `_ingest_turn_ids` was defensive against a case that never arises in `run_worker` output. The schema column is over-typed (TEXT-encoded JSON array), but the empirical contract is singular — consistent with the `_ingest_turn_ids` docstring's "per-sentence derived belief id (in input order, with duplicates dropped)."

So the `ids[0]`-only collapse is **safe at current head**. The intra-turn edge wiring requires a positional single-id per sentence; the loop refactor is the natural shape for that.

### 3. Lock-the-assumption sub-ask is still unaddressed

The owner's #2 sub-ask — an explicit assertion or test locking the "multi-id rows are unreachable from `_ingest_turn_ids`" assumption — is **not** in the current diff. `tests/test_ingest_subfloor_noise.py` does not grep for `derived_belief_ids`. The over-typed schema means a future contributor could legitimately add a multi-id producer (e.g., a sentence that fans out to N beliefs through derivation) without realizing `_ingest_turn_ids` will silently swallow N-1.

Minimum: a one-line inline comment on `head = ids[0]` documenting "by contract `derivation_worker` emits at most one belief id per log row; if that contract ever broadens, this collapse drops the tail and `inserted` becomes incomplete" would be enough. Author's call whether to address inline here or open a follow-up. Not blocking on my read.

### 4. Logic spot-check on the edge-wiring loop

Read through `for i in range(1, len(log_belief_ids))` in `ingest.py`. The `prior_bid == curr_bid` short-circuit correctly suppresses self-edges when two adjacent sentences canonicalize to the same belief id (e.g., duplicate prose in one turn). One consequence: a sub-floor clause sandwiched between two occurrences of the *same* belief is silently dropped rather than attached as anchor_text — natural reading since there's no relational edge to wire, but worth noting if anyone later expects subfloor demotion to be a strict invariant. Tests cover the typical case (full→subfloor→full with distinct beliefs); they do not cover the same-belief sandwich. Not blocking.

### 5. Discretion

Diff vs `github/main` is clean — no public-boundary issues.

### Net

Rebase blocker is cleared by the new HEAD; the multi-id question is answered by the producer audit (single-id by construction in all three writers). Awaiting `pytest (3.12/3.13)` green on `4eb7df4f` and an owner formal approve on the multi-id finding; once both, re-add `ready-to-merge` and the merge-train should ship.

robotrocketscience:
[release:review:bagheera:2026-05-14T21:40:00Z]

PR #776, opened 2026-05-14
feat(bench): wire A4 continuation-fidelity bench gate for type-aware compression (#775)

robotrocketscience:
## Summary

Wires the A4 bench gate for `use_type_aware_compression` flip-default (precursor to #769). Mirrors A2's harness shape: schema validator + strict-band assertion against a lab-side corpus, runner in `tests/retrieve_uplift_runner.py`, skips cleanly on public CI when `AELFRICE_CORPUS_ROOT` is unset.

## Changes

1. **`tests/retrieve_uplift_runner.py`** — adds `CompressionA4Fidelity` + `run_compression_a4_fidelity`. Per row: build a transient `MemoryStore` from row beliefs, build `recent_turns` from `transcript_pre_clear`, call `rebuild_v14` under `AELFRICE_TYPE_AWARE_COMPRESSION=0` then `=1` at the row's `rebuilder_token_budget`. Score each arm via a deterministic token-coverage proxy against `expected_post_clear_answers` — fraction of normalized answer tokens present in the normalized rebuild block. Proxy documented inline so a later swap to captured-answer scoring (when corpus rows carry `captured_post_clear_answers_{off,on}`) is a drop-in.
2. **`tests/bench_gate/test_compression_a4_fidelity.py`** — bench-gate test. Schema validator checks row contract (`id`, `transcript_pre_clear`, `beliefs`, `expected_post_clear_answers`, optional `rebuilder_token_budget`). Strict-band test asserts `mean_fidelity_on >= mean_fidelity_off - 0.005` per spec § A4 (band mirrors BM25F #154 model).

## Spec source

`docs/feature-type-aware-compression.md` § A4:

> The continuation-fidelity scorer (#141 v1.4 deliverable) is run on the rebuild_logs corpus with `use_type_aware_compression={OFF, ON}`. Bench-gate: ON ≥ OFF on continuation-fidelity score at the same `[rebuilder] token_budget`. Tolerance band: `≥ baseline − 0.005`.

## Corpus contract

Corpus rows live in the private companion repo at `<corpus_root>/compression_a4_fidelity/*.jsonl` (directory-of-origin rule — public repo carries only the schema contract).

```
{
  "id": "row-id",
  "transcript_pre_clear": [
    {"role": "user"|"assistant", "text": "...",
     "session_id": "...", "ts": "..."}
  ],
  "beliefs": [
    {"id": "...", "content": "...",
     "retention_class": "fact"|"snapshot"|"transient"|"unknown",
[...]

Issue #775, opened 2026-05-14
Wire A4 bench gate harness for type-aware compression — #769 precursor


robotrocketscience:
## Summary

Wire the A4 bench-gate harness for `use_type_aware_compression` flip-default decision. Precursor to #769; the A4 wiring was tracked inline there but is now scoped concretely enough for its own sub-issue per the parent's "may want its own sub-issue once scoped" note.

A2 shipped its harness at `tests/bench_gate/test_compression_a2_recall.py` (runner: `tests/retrieve_uplift_runner.py::run_compression_a2_uplift`). A4 needs the same shape against the **rebuild_logs corpus** with the **continuation-fidelity scorer** (#138, already shipped).

## Spec source

Per `docs/feature-type-aware-compression.md` § A4:

> The continuation-fidelity scorer (#141 v1.4 deliverable) is run on the rebuild_logs corpus with `use_type_aware_compression={OFF, ON}`. Bench-gate: ON ≥ OFF on continuation-fidelity score at the same `[rebuilder] token_budget`. Tolerance band: `≥ baseline − 0.005`.

## Scope

**Test file:** `tests/bench_gate/test_compression_a4_fidelity.py`

Mirror A2's pattern:
- `@pytest.mark.bench_gated` — skips when `AELFRICE_CORPUS_ROOT` unset
- One `test_*_corpus_round_trip` — schema validation, asserts every row has the expected fields, expected_post_clear_answers reference valid turn indices
- One `test_*_strict_fidelity_band` — loads corpus, drives OFF/ON arms through `run_compression_a4_fidelity`, asserts `fidelity_on >= fidelity_off - 0.005`

**Runner:** `tests/retrieve_uplift_runner.py::run_compression_a4_fidelity`

Per-row: load the transcript prefix from the row, exercise the real rebuilder (`aelfrice.context_rebuilder.rebuild()`) twice — once with `use_type_aware_compression=False`, once `=True`, at fixed `[rebuilder] token_budget`. Score continuation fidelity for each arm via the `benchmarks.context_rebuilder.score.score_continuation_fidelity` exact-method pipeline using the row's `expected_post_clear_answers`. Aggregate `mean_fidelity_off`, `mean_fidelity_on`.

**Result dataclass:** `CompressionA4Fidelity(n_rows, mean_fidelity_off, mean_fidelity_on)` parallel to `CompressionA2Uplift`.

## Corpus contract (lab-side)

[...]

Issue #769, opened 2026-05-13
Flip default: use_type_aware_compression on by default — gated on A2 + A4 bench run


robotrocketscience:
## Summary

Flip `resolve_use_type_aware_compression()` default from `False` → `True` (`src/aelfrice/retrieval.py:995`). The type-aware compression transform shipped at v2.0.0 behind a default-OFF flag per the bench-gate / ship-or-defer policy in `docs/feature-type-aware-compression.md` §"Bench-gate / ship-or-defer policy" (line 191). #434 closed the implementation but the bench evidence to flip the default was never captured, so the flag is still default-OFF at v3.0.

This flip is bench-gated. The criteria are the A2 + A4 gates from the v2.0.0 spec.

## Why this is the right intent

1. The transform is implemented, deterministic (A3 verified), and integrated with the rebuilder (A5 tracker row present).
2. Default-OFF was a ship-or-defer pessimism, not a long-term design choice — the spec explicitly says "ship only on positive bench evidence per A2 and A4."
3. Sibling features that cleared their bench gates have already flipped: BM25F (#154), heat-kernel (#150), HRR structural (#152 / #437), intentional clustering (#436 R6). Type-aware compression is the last v2.0-era flag still default-OFF on a positive-evidence-pending basis.
4. `bfs_enabled` (#739) is separately bench-gated on latency; this one is bench-gated on recall/fidelity uplift. Independent decisions, independent gates.

## Precursor work — UPDATED 2026-05-14

Both bench-gate harnesses are wired, the rebuilder's structural compression-blindness (#798) is closed, and the lab corpora are populated. The remaining precursor is **executing the benches**; whether the results justify the flip-default is the open question, debated in the R7-R13 comment thread below.

### A2 — harness wired; lab corpus populated
- Harness: `tests/bench_gate/test_compression_a2_recall.py` (`@pytest.mark.bench_gated`; skips when `AELFRICE_CORPUS_ROOT` unset).
- **Lab corpus verified present** (2026-05-14T17:02Z): `~/projects/aelfrice-lab/tests/corpus/v2_0/compression_a2_recall/v0_1.jsonl` — 5 rows matching the A2 schema (`{id, query, k, token_budget, expected_top_k, beliefs}`).
- Statistical-power adequacy at N=5 is a separate question; see the R7-R13 comments.

### A4 — harness wired; structural rebuilder gap closed; lab corpus populated
[...]

Issue #288, opened 2026-04-29
Rebuild redesign: eval harness — per-rebuild log + fixed-corpus precision


robotrocketscience:
## Question

We cannot fix the rebuilder without measuring it. The v1.4 continuation-fidelity harness measures *downstream behavior* (does the agent continue similarly with vs without the rebuild block?) but does not measure **rebuild precision** — i.e. given a query, did the right beliefs get retrieved?

Without that, every proposed fix to ranking, gating, typing, or query understanding is unfalsifiable.

## What we actually need

Two layers:

### Layer 1: Per-rebuild diagnostic log (cheap, ships first)

On every rebuild invocation, write a side-file row containing:
- timestamp
- session_id
- recent-turns input (or hash thereof)
- entities/intents extracted (the query)
- top-K candidate beliefs with scores (BM25 raw + posterior + any reranker score)
- which candidates were packed (after dedup, after floor, after budget)
- which were dropped and why

One JSONL per session, kept under `<project>/.git/aelfrice/rebuild_logs/`. Operator-readable. Cheap. Ships before any redesign.

This alone — without any change to the ranker — would tell us within one operator-week whether the diagnoses in #281 are accurate, where the actual failures are, and what threshold values would have helped.

### Layer 2: Fixed-corpus precision harness (heavier, calibration-grade)

A captured corpus: ~50-100 hand-labeled (query, expected-belief-set) pairs covering the categories operators actually hit (debugging, code review, research, design). Run a candidate ranker against the corpus, compute precision@K + recall@K. Required for any quantitative claim that ranker variant A beats B.

## Sub-questions

- Where does the captured corpus come from? Hand-curated from operator sessions? Synthetic? Mixed?
- How is the "expected belief set" labeled? By the operator post-hoc? By a second-pass LLM judge? Both have failure modes.
- Does the corpus need to be project-specific (aelfrice's own beliefs) or generalizable (synthetic stores)?
- Cadence: is the harness run on every PR that touches retrieval code? Once per release?
- Storage: corpus + labels in the public repo, or lab-only? (Per the two-repo model — the corpus may contain operator-private context.)

## Why this is phase-1 work

Every other child issue (#TBD-1 floor, #TBD-2 typing, #TBD-3 query) needs this to make calibration decisions. Build the instrument first, then iterate. Skipping this means picking thresholds + heuristics by intuition — which is how the current state was reached.

## Out of scope

- The continuation-fidelity harness itself (already exists at v1.4 spec).
[...]

PR #815, opened 2026-05-14
fix(rebuilder): honour use_type_aware_compression in pack accounting (#798)

### LINKED ISSUE #798: A4 bench gate is structurally vacuous — rebuilder uses verbatim token cost, ignoring compression flag

robotrocketscience:
## Summary

Closes #798. `rebuild_v14` was re-packing `retrieve()`'s candidate set with verbatim token cost regardless of `use_type_aware_compression`, so any ON-arm extras `retrieve()` admitted at compressed cost got trimmed back to the OFF-arm count. The downstream A4 continuation-fidelity bench gate (#775 / PR #776) was therefore structurally vacuous — per-row fidelity delta = 0 by construction, regardless of corpus.

Operator decision on #798: **Option A — thread `compress_on` into rebuilder pack accounting.** Smallest diff; rebuild-block content stays verbatim; A4 gate becomes structurally measurable. Options B (rebuilder emits `compressed_beliefs[i].rendered`) and C (drop A4 from #769 acceptance) declined.

## What ships

1. `src/aelfrice/context_rebuilder.py`
   - New `use_type_aware_compression: bool | None = None` kwarg on `rebuild_v14`. Resolves once via `resolve_use_type_aware_compression(...)` at function entry; resolved boolean threaded to all sites.
   - `retrieve()` call inside `rebuild_v14` now passes `use_type_aware_compression=compress_on` so retrieve's L1 / L2.5 pack is compression-aware on the rebuilder path.
   - `_estimate_belief_tokens(b, *, compress_on=False)` gains a keyword-only flag. When True, it returns `compress_for_retrieval(b, locked=(b.lock_level == LOCK_USER)).rendered_tokens`; else the existing verbatim `(len + 3) // 4` math. Default-OFF preserves the byte-identical contract for the #139 / #288 regression tests and for the legacy `_retrieve_for_rebuild` (v1.2.0a0 alpha) path, which keeps its default-OFF call.
   - All three live rebuild_v14 pack sites (L0 init at :414, session tier at :463, L1 / L2.5 tier at :513) pass `compress_on=compress_on`.

2. `tests/test_context_rebuilder.py` — two new tests:
   - `test_rebuild_v14_pack_size_matches_compression_flag` — asserts ON packs strictly more beliefs than OFF at a budget that forces trim. Failing this means `_estimate_belief_tokens` is ignoring the flag and #798's verbatim-cost re-pack bug has regressed.
   - `test_rebuild_v14_compression_off_byte_identical_default` — asserts `rebuild_v14(...)` with no kwarg and no env override equals `rebuild_v14(..., use_type_aware_compression=False)`.

3. `CHANGELOG.md` — entry under `[Unreleased]` / Fixed.

4. `docs/feature-type-aware-compression.md` — §"vs. context rebuilder" rewritten (the old text described an Option-B-style `RetrievalResult` shape change that never shipped); §A4 gains a prerequisite note pointing at the pack-size property test.

## Observable behaviour change

- Default-OFF (env unset, no kwarg, no TOML override): **byte-identical**. The new test pins this.
- Default-ON (`AELFRICE_TYPE_AWARE_COMPRESSION=1` or explicit kwarg): rebuild block now contains *strictly more beliefs* at the same `token_budget` when the candidate set under verbatim cost would have trimmed. Each belief still renders verbatim.

## Tests

`uv run pytest tests/ --ignore=tests/bench_gate` — 4164 passed, 33 skipped, 75 xfailed (pre-existing).

The bench-gate tests at `tests/bench_gate/test_compression_a4_fidelity.py` remain `bench_gated` (skip without `AELFRICE_CORPUS_ROOT`); now that the rebuilder honours the flag, a lab-side run against a populated corpus will produce nonzero per-row deltas instead of 0-by-construction.

## What this unblocks

#769 (flip default `use_type_aware_compression` on). The A4 axis of that flip-default decision was load-bearing on a gate that could not measure compression; with this fix it can. The A2 axis was unaffected (#769 acceptance remains as-is for both bench runs).

## Risk

- `#139` (byte-identical v1.x rebuild) and `#288` (rebuild_log shape) regression tests both run green under default-OFF; the keyword-only `compress_on` parameter on `_estimate_belief_tokens` defaults to False.
- The legacy `_retrieve_for_rebuild` (v1.2.0a0 alpha contract, frozen docstring) is intentionally not modified — its two `_estimate_belief_tokens(b)` call sites use the default `compress_on=False`.
- Default-ON behaviour change: consumers reading the rebuild block under the flag now see a larger block (more belief entries) at the same budget. Content per entry unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@ installable release; see the roadmap in [README.md](README.md).
 
 ## [Unreleased]
 
+### Fixed
+
+- **Rebuilder pack accounting now honours `use_type_aware_compression`** ([#798](https://github.com/robotrocketscience/aelfrice/issues/798)). `rebuild_v14` was re-packing `retrieve()`'s candidate set with verbatim token cost regardless of the flag, so any ON-arm extras `retrieve()` admitted at compressed cost got trimmed back to the OFF-arm count. The downstream A4 continuation-fidelity bench gate ([#775](https://github.com/robotrocketscience/aelfrice/issues/775) / [PR #776](https://github.com/robotrocketscience/aelfrice/pull/776)) was therefore structurally vacuous — per-row fidelity delta = 0 by construction, regardless of corpus. Fix resolves the flag once at `rebuild_v14` entry (`resolve_use_type_aware_compression(use_type_aware_compression)`), threads it into the `retrieve()` call and into `_estimate_belief_tokens(b, *, compress_on=...)` at all three pack sites (L0 init, session tier, L1 / L2.5 tier). The rebuild block content itself stays verbatim — the change is in *how many* beliefs survive the budget, not what each surviving belief renders as. Default-OFF and the legacy `_retrieve_for_rebuild` (v1.2.0a0 alpha contract) are byte-identical. Unblocks the A4 axis of the [#769](https://github.com/robotrocketscience/aelfrice/issues/769) flip-default decision. Operator-decision history: Option A per [#798 thread](https://github.com/robotrocketscience/aelfrice/issues/798); Options B (rebuilder emits `compressed_beliefs[i].rendered`) and C (drop A4 from #769 acceptance) declined. Two new tests in `tests/test_context_rebuilder.py` (`test_rebuild_v14_pack_size_matches_compression_flag`, `test_rebuild_v14_compression_off_byte_identical_default`).
+
 ## [3.1.0] - 2026-05-14
 
 ### Added
--- a/docs/feature-type-aware-compression.md
+++ b/docs/feature-type-aware-compression.md
@@ -138,7 +138,7 @@ Per `docs/belief_retention_class.md` § "Recommendation summary": *"Soft down-we
 
 The rebuilder consumes retrieval output and writes a continuation-fidelity-scored block at `<git-common-dir>/aelfrice/rebuild_logs/`. With compression on, the rebuilder gets more total beliefs in the same `[rebuilder] token_budget` (default in `context_rebuilder.py:111`). Acceptance #4 makes this measurable: continuation-fidelity uplift at fixed budget.
 
-The rebuilder does not need a separate config knob; it inherits the retrieval-side flag. If `use_type_aware_compression` is on at retrieval call-time, the rebuilder reads compressed output transparently via the `RetrievalResult` shape change above.
+The rebuilder does not need a separate config knob; it inherits the retrieval-side flag. `rebuild_v14` resolves `use_type_aware_compression` once at function entry via `resolve_use_type_aware_compression(use_type_aware_compression)`, threads the resolved boolean into the `retrieve()` call so retrieve's L1/L2.5 pack accounting is compression-aware, and threads it into `_estimate_belief_tokens(b, *, compress_on=...)` at all three pack sites in rebuild_v14 (L0 init, session tier, L1 / L2.5 tier). The rebuild-block content itself remains verbatim — compression changes *how many* beliefs fit under the budget, not what each surviving belief renders as. This is the #798 fix; before it landed, rebuild_v14 used verbatim cost for its own re-pack and silently trimmed retrieve()'s ON-arm extras back to the OFF-arm count, leaving A4 structurally vacuous (per-row fidelity delta = 0 regardless of corpus).
 
 ### vs. type / source-tier axes
 
@@ -181,6 +181,8 @@ A second test compares against a fixture of `(belief_id, expected_rendered)` pai
 ### A4 — rebuilder fidelity
 
 The continuation-fidelity scorer (#141 v1.4 deliverable) is run on the rebuild_logs corpus with `use_type_aware_compression={OFF, ON}`. Bench-gate: ON ≥ OFF on continuation-fidelity score at the same `[rebuilder] token_budget`. Tolerance band: `≥ baseline − 0.005` (a half-point of the fidelity-score noise floor, mirroring the BM25F bench-gate band at #154).
+
+Prerequisite: rebuild_v14's pack accounting must honour the compression flag — otherwise per-row fidelity delta is 0 by construction, as documented in #798. That prerequisite landed alongside this doc revision (see `tests/test_context_rebuilder.py::test_rebuild_v14_pack_size_matches_compression_flag` — asserts ON packs strictly more beliefs than OFF at a budget that forces trim).
 
 ### A5 — composition tracker
 
--- a/src/aelfrice/context_rebuilder.py
+++ b/src/aelfrice/context_rebuilder.py
@@ -80,7 +80,8 @@ from aelfrice.query_understanding import (
     VALID_STRATEGIES,
     transform_query,
 )
-from aelfrice.retrieval import retrieve
+from aelfrice.compression import compress_for_retrieval
+from aelfrice.retrieval import resolve_use_type_aware_compression, retrieve
 from aelfrice.scoring import posterior_mean
 from aelfrice.store import MemoryStore
 from aelfrice.triple_extractor import extract_triples
@@ -316,6 +317,7 @@ def rebuild_v14(
     floor_l1: float = 0.0,
     query_strategy: str = DEFAULT_QUERY_STRATEGY,
     working_state: "WorkingState | None" = None,
+    use_type_aware_compression: bool | None = None,
 ) -> str:
     """v1.4 rebuild: L0 + session-scoped + L2.5/L1 via `retrieve()`.
 
@@ -358,6 +360,12 @@ def rebuild_v14(
     `aelfrice.query_understanding`. `legacy-bm25` is byte-identical
     to the v1.4 path and remains opt-in until PR-4 removes it.
     """
+    # #798: resolve compression flag once, thread through pack accounting
+    # so rebuild_v14's trim matches retrieve()'s trim under the same flag.
+    compress_on: bool = resolve_use_type_aware_compression(
+        use_type_aware_compression,
+    )
+
     locked: list[Belief] = store.list_locked_beliefs()
     locked_ids: set[str] = {b.id for b in locked}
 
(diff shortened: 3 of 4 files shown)


robotrocketscience:
[claim:review:bagheera:2026-05-14T20:32:17Z]

robotrocketscience:
Review (sister-session, cannot `--approve` — same GitHub author):

- `rebuild_v14` resolves `use_type_aware_compression` once at entry and threads the boolean to `retrieve()` and all three `_estimate_belief_tokens` sites (L0 init, session tier, L1 / L2.5 tier).
- `_estimate_belief_tokens(b, *, compress_on=False)` mirrors retrieval._cost: `compress_for_retrieval(b, locked=(b.lock_level == LOCK_USER)).rendered_tokens` when ON, verbatim cost otherwise. Locks render verbatim either way per the strategy table.
- Default-OFF byte-identical contract pinned by `test_rebuild_v14_compression_off_byte_identical_default` (pops `AELFRICE_TYPE_AWARE_COMPRESSION` env first).
- Pack-size delta pinned by `test_rebuild_v14_pack_size_matches_compression_flag` — guards against the verbatim-cost regression #798 documents.
- Legacy `_retrieve_for_rebuild` (v1.2.0a0 alpha contract) intentionally unmodified — both call sites keep default `compress_on=False`.
- Both commits SSH-signed (G). Discretion grep on the diff vs main is clean.
- A4 bench gate remains bench-gated (skip without `AELFRICE_CORPUS_ROOT`) — this PR unblocks measurability; lab-side run still required to score #769's flip-default decision.

Labeling `ready-to-merge`.

robotrocketscience:
[release:review:bagheera:2026-05-14T20:35:55Z]

PR #821, opened 2026-05-14
feat(retrieval): hot-path touch-state storage substrate (#816, closes #748)

robotrocketscience:
## Summary

Lands v1 of the #748 hot-path campaign per
`experiments/hot-path/DESIGN.md` (lab `6b40538`). v1 is the **storage
substrate only** — `belief_touches` accumulates per-(belief, session)
touch state; the rerank consumer is parked on R7c per DESIGN.md v1
ship list item 7.

Closes the storage axis of #748. Issue #816 stays in the loop for the
post-R7c consumer-flip PR.

## Scope (DESIGN.md v1)

| Item | Status |
|---|---|
| 1. Sidecar `belief_touches` table | ✅ |
| 2. Boolean "touched in last K fires" decay; no exponential τ | ✅ |
| 3. INJECTION-only event kind | ✅ |
| 4. Bitmask reserves `event_kinds_bitmask` bits 1–3 for v2 | ✅ |
| 5. #744 JSON-ring → sidecar migration | ✅ (one-shot per-session, idempotent) |
| 6. `aelf doctor --hot-path` inspection surface | ✅ |
| 7. **No** retrieval consumer in v1 | ✅ (parked post-R7c) |

## Divergence from #816's issue body

The issue body asked for a default-OFF rerank consumer wired into
`retrieve_v2` (`resolve_use_hot_path_touch_state`,
`hot_path_multiplier`, federation property-test for `multiplier=1.0`).
DESIGN.md v1 is explicit: **"No retrieval consumer. Touch state is
written but the rerank multiplier is dark — gated behind
`meta:retrieval.touch_temperature_enabled = false`. Flipping the gate
is a separate post-H3 PR."** The DESIGN was authored on the same day
as the issue and reflects the post-R7c verdict; the issue body
predates it. This PR follows DESIGN.md.

Concrete consequences: no `resolve_use_hot_path_touch_state` resolver,
no `hot_path_multiplier` helper, no `retrieve_v2` wiring changes. The
federation property is preserved structurally by the composite PK
`(belief_id, session_id)` — foreign federated beliefs come in cold by
construction rather than via a `multiplier=1.0` short-circuit. Doctor
surface, schema, migration, and lab evidence all match DESIGN.md
literally.

## Lab evidence reference (R0..R7c)

- **R1 / R1b** — sidecar wins +45.7% p50 / +42.2% p95 on rerank-stage
  reads at N=50k / 5k, no regression on unrelated reads; result holds
[...]

PR #831, opened 2026-05-15
fix(hook): pre-push scans commit messages, not just diff content (#830)

### LINKED ISSUE #830: fix(hook): pre-push discretion grep skips commit messages
## Bug

The user-managed pre-push discretion hook at `.git/hooks/pre-push` (the per-repo discretion enforcement chained from `.githooks/pre-push`, not the freshness-check hook) scans `git diff base..HEAD` content but does **not** scan commit messages. Sister-session names, model identifiers, or other banned vocabulary that lives in a commit subject or body slips past the hook entirely and lands on `github/main` (and is then pinned forever in `refs/pull/N/head`).

## Repro (2026-05-14)

PR #821 commit `b030a4f2` shipped with this body line:

> "PR #821 review ([five session names], 2026-05-14) flagged that..."

The body listed five session-routing names verbatim (paraphrased here to avoid re-amplifying the leak). The pre-push hook didn't catch it because the diff content was clean — only the message carried the leak. Caught after-the-fact during a rebase pass; the rewritten copy is scrubbed but the original SHA persists on `refs/pull/821/head`.

## Acceptance

In `.git/hooks/pre-push` (the discretion hook, **not** the in-repo `.githooks/pre-push` freshness check), extend the per-ref loop to also scan commit messages between the merge-base and the local SHA. One snippet near the existing `git diff` block:

```bash
# Scan commit messages too — diff-only check misses leaks in subjects/bodies.
msg_hits=$(git log --format=%B "$base..$local_sha" | \
  grep -niE "$BANNED_VOCAB" || true)
if [ -n "$msg_hits" ]; then
    echo "PRE-PUSH BLOCKED: banned vocab in commit message(s) for $local_ref"
    echo "$msg_hits" | head -10
    fail=1
fi
```

Also extend `BANNED_PHRASES` similarly. The override (`ALLOW_DISCRETION_OVERRIDE=1`) should still apply.

Because the file is user-managed (not in-repo), the actual fix needs to be propagated by updating each contributor's hook — consider adding a `scripts/install-discretion-hook.sh` installer that writes the canonical hook to `.git/hooks/pre-push` (idempotent, refuses to overwrite a divergent existing hook without `--force`), so this is a reproducible install rather than per-machine drift.

## Out of scope

- The in-repo `.githooks/pre-push` freshness-check hook — different file, different purpose.
- Rewriting the pre-existing pinned commit on `refs/pull/821/head` (impossible by design).

## Tier

`rook` — small, well-scoped change to a hook script + optional installer.

robotrocketscience:
Closes #830.

## Bug

The discretion pre-push hook at `.git/hooks/pre-push` (user-managed, not
in-tree) scans `git diff base..HEAD` content but never inspects commit
messages. Banned vocabulary in a commit subject or body slipped past the
hook and landed on `github/main` — pinned forever in `refs/pull/N/head`.

Repro: PR #821 commit `b030a4f2` shipped with a body line listing five
session-routing names verbatim. The diff was clean; only the message
carried the leak. Caught after-the-fact; the rewritten copy is scrubbed
but the original SHA persists.

## Change

One new tracked file: `scripts/install-discretion-hook.sh`. The installer's
heredoc is the single source of truth for the canonical hook. Relative to
the prior hook content, the canonical adds:

- **Check 4** — `BANNED_VOCAB` on `git log --format=%B "$msg_range"`
- **Check 5** — `BANNED_PHRASES` on the same

Both run inside the existing `while read` ref loop. When `base` is empty
(no common ancestor with main), `msg_range` is `$local_sha` so the full
reachable history is walked. Error message shape and the
`ALLOW_DISCRETION_OVERRIDE=1` bypass cover Checks 4/5 with no special-case.

Checks 2 and 3 (diff-content scans) gain
`':(exclude)scripts/install-discretion-hook.sh'` in their path-exclude
lists, because the installer's heredoc necessarily carries the literal
`BANNED_VOCAB` and `BANNED_PHRASES` regexes and would otherwise self-block
on every push that touches this file. Checks 4/5 have no path-excludes —
commit messages have no paths.

## Installer semantics

- `--help` prints usage from the header docblock.
- Idempotent: byte-for-byte match → exit 0 "already up to date".
- Refuses divergent overwrite without `--force`: prints diff and exits 1.
- `--force` overwrites cleanly. Sets the result executable.
- Resolves repo root via `git rev-parse --show-toplevel`; exits 1 if not in
  a git working tree.

## Migration

Each contributor whose `.git/hooks/pre-push` predates this PR will see a
divergent-hook warning on first install:

```
scripts/install-discretion-hook.sh --force
```

after pulling the change. The `--force` is required exactly once per
contributor; subsequent runs are no-ops.

## Verification

- `bash -n scripts/install-discretion-hook.sh` clean.
- Manual repro confirmed against this branch: a synthetic commit with
  banned vocab in the message but clean diff fires the Check 4 block as
  expected — exactly the PR #821 b030a4f2 scenario.
- The installer ran against the local repo and the new hook is now
  blocking commit-message leaks here.

### DIFF
--- a/scripts/install-discretion-hook.sh
+++ b/scripts/install-discretion-hook.sh
@@ -1,0 +1,216 @@
+#!/usr/bin/env bash
+# install-discretion-hook.sh — write the canonical pre-push discretion hook to
+# .git/hooks/pre-push in the aelfrice working tree.
+#
+# Usage:
+#   scripts/install-discretion-hook.sh [--force] [--help]
+#
+#   --force   Overwrite an existing hook even if it differs from the canonical
+#             content. Without --force the script prints a diff and exits 1.
+#
+# The hook blocks banned paths, banned vocabulary, and CLAUDE.md-derived phrases
+# from being pushed — both in diff content and in commit messages.
+#
+# This installer is idempotent: if .git/hooks/pre-push already matches the
+# canonical content byte-for-byte, it exits 0 with "already up to date".
+#
+# Must be run from inside the aelfrice git working tree.
+
+set -euo pipefail
+
+FORCE=0
+
+for arg in "$@"; do
+    case "$arg" in
+        --help|-h)
+            sed -n '2,/^$/p' "$0" | grep '^#' | sed 's/^# \?//'
+            exit 0
+            ;;
+        --force)
+            FORCE=1
+            ;;
+        *)
+            echo "Unknown argument: $arg" >&2
+            echo "Usage: $0 [--force] [--help]" >&2
+            exit 1
+            ;;
+    esac
+done
+
+# Resolve the repo root — works from any directory inside the working tree.
+if ! REPO_ROOT=$(git rev-parse --show-toplevel 2>/dev/null); then
+    echo "ERROR: not inside a git working tree." >&2
+    exit 1
+fi
+
+HOOK_PATH="$REPO_ROOT/.git/hooks/pre-push"
+
+# Canonical hook content lives here as a heredoc. This is the single source of
+# truth. If you update the hook logic, update it here and re-run the installer.
+CANONICAL=$(cat <<'HOOK_EOF'
+#!/bin/bash
+# Pre-push guard for ~/projects/aelfrice (public repo).
+# Blocks any push whose diff or commit messages contain:
+#   1. Files under tests/corpus/v2_0/**/*.jsonl  (corpus content lives in lab repo only)
+#   2. Files under .claude/, .planning/, ~/.claude/handoffs/  (session/handoff content)
+#   3. Banned vocabulary in diff content (codenames, model identifiers, pipeline vocab)
+#   4. Banned vocabulary in commit messages
+#   5. Banned phrases from ~/.claude/CLAUDE.md in commit messages
+#
+# To override for an emergency, set ALLOW_DISCRETION_OVERRIDE=1 in the env. The override
+# logs to ~/.aelfrice/discretion-override.log; investigate every entry afterward.
+#
+# DO NOT REMOVE OR WEAKEN THIS FILE WITHOUT EXPLICIT USER REQUEST.
+
+set -euo pipefail
+
+remote="${1:-}"
+url="${2:-}"
+
+z40=0000000000000000000000000000000000000000
+
+# Banned paths — extended regex
+BANNED_PATHS='^(tests/corpus/v2_0/.+\.jsonl|\.claude/|\.planning/|.*/handoffs/.*\.md)$'
+
+# Banned vocabulary in diff content — case-insensitive extended regex.
+# Word boundaries are intentional to avoid blocking on substrings like "claude.md".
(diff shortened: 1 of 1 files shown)


robotrocketscience:
[claim:review:clarke:2026-05-15T00:13:21Z]

robotrocketscience:
[claim:review:prince:2026-05-15T01:17:11Z]

robotrocketscience:
[claim:review:bagheera:2026-05-15T01:17:13Z]

robotrocketscience:
[release:review:clarke:2026-05-15T01:17:14Z]

robotrocketscience:
[release:review:bagheera:2026-05-15T01:17:17Z]

robotrocketscience:
Reviewed `9cbbc7a0` (rebased from `ce8f8280` onto current main `04a68ba5`).

**AC coverage vs #830:**
- Check 4 (BANNED_VOCAB on `git log --format=%B "$msg_range"`) — present, fires inside the existing `while read` ref loop.
- Check 5 (BANNED_PHRASES, same source) — present, same path.
- `msg_range` falls back to `$local_sha` when no common ancestor exists — walks full reachable history, matches the issue's prescription.
- `ALLOW_DISCRETION_OVERRIDE=1` covers Checks 4/5 with no special-case, as required.
- Path-excludes are correctly *not* applied to commit-message scans.

**Installer semantics:**
- Idempotent (byte-for-byte match → exit 0). Verified `printf '%s\n' "$CANONICAL"` vs `cat`-stripped trailing newline pairs cleanly.
- Divergent overwrite refused without `--force`; diff printed.
- `--force` overwrites, sets exec bit.
- Repo-root via `git rev-parse --show-toplevel`; clean exit if not in a working tree.

**Self-block avoidance:**
- Checks 2/3 exclude `scripts/install-discretion-hook.sh` from diff scans — required because the canonical heredoc necessarily contains the literal BANNED_VOCAB/BANNED_PHRASES regexes. This is the only structural location the regex can live; the exclude is the right call.

**CI:** all required checks green on the pre-rebase head (pytest 3.12/3.13, bench-smoke, CodeQL python+actions, secrets-scan, pattern-scan, history-scan, commit-msg-prefix, pr-body-issue-link, deptry, vulture, typos). Re-runs will fire on the rebased head.

**Rebase:** trivial — main only advanced by one unrelated commit (`scripts/sync-repo-settings.sh`). No conflicts. New head `9cbbc7a0` is signed and FF on `04a68ba5`.

**Minor non-blocking observations** (follow-up issue territory, not gating this PR):
1. `BANNED_VOCAB` lists `Setr|Kulili|Gylf` but not `Toug|Youg` — the issue body specifically notes #821 leaked five session names. Worth a follow-up to either complete the enumeration or move to an anonymous-token scheme (matches the long-term direction noted in the parallel-session protocol).
2. No automated test for the heredoc itself — if a future edit breaks a regex, the failure mode is silent. A small shellcheck/regex-validation test in `tests/scripts/` would catch it. Out of scope for this PR.
3. The `--help` parser relies on the first blank line ending the docblock. Tight coupling; fine for now.

Re-adding `ready-to-merge` per the merge-train comment.

robotrocketscience:
[release:review:prince:2026-05-15T01:24:10Z]

Issue #833, opened 2026-05-15
[bug] aelf v3.1.0 crashes on any DB touched by post-#814 code — demotion_pressure schema/reader skew


robotrocketscience:
## Summary

`aelf 3.1.0` (currently-published release) crashes with `IndexError: No item with that key` on any operation that goes through `get_belief()` (and therefore `_row_to_belief()` at `store.py:651`) whenever the local `~/.aelfrice/store.db` has had its `beliefs.demotion_pressure` column dropped by a newer build of `aelf`.

```
File ".../aelfrice/store.py", line 651, in _row_to_belief
    demotion_pressure=row["demotion_pressure"],
                      ~~~^^^^^^^^^^^^^^^^^^^^^
IndexError: No item with that key
```

Confirmed-affected commands: `aelf wonder`, and any command that walks the graph through `get_belief` (`aelf reason`, `aelf search` graph-walks, `aelf:reason` skill).

## Repro

1. Install published `aelfrice` v3.1.0 (`uv tool install aelfrice==3.1.0`). DB starts with the column intact.
2. From any worktree based on `main` post-`fed650ca` (#814), run `uv run aelf <anything>` — `_MIGRATION_STATEMENTS` at `store.py:606` executes `"ALTER TABLE beliefs DROP COLUMN demotion_pressure"` against `~/.aelfrice/store.db`. The column is gone.
3. Run the installed `aelf` again (`aelf wonder`). Crash above.

## Root cause

Two coupled defects:

1. **Forward-incompatible migration not paired with a release.** Tag `v3.1.0` is at commit `0b298728`. Commit `fed650ca` (`refactor(store): drop beliefs.demotion_pressure column (#814)`) landed on `main` *after* the tag. The migration in `_MIGRATION_STATEMENTS` permanently drops the column from any DB the new code touches; the published v3.1.0 reader cannot recover.
2. **`_row_to_belief` is non-defensive about the dropped column on v3.1.0.** On `main`, the read was removed entirely from `_row_to_belief`. On v3.1.0 the read at line 651 is unconditional, without the `"X" in keys else <default>` pattern that already protects `retention_class` (`store.py:660`), `valid_to` (`:672`), and `scope` (`:674`). The defensive pattern was added precisely so forward-incompatible schemas wouldn't break older readers — the drop-column migration broke that contract.

## Fix options

[...]

PR #836, opened 2026-05-15
fix(auto-install): gate cli-entry on uv-tool install context (#834)

### LINKED ISSUE #834: [bug] uv run aelf from a worktree silently mutates ~/.claude/settings.json and opts user out of hooks
## Summary

`uv run aelf <anything>` from a project worktree whose source tree pins a different aelfrice version than the user's installed `aelf` causes the in-worktree code path to rewrite `~/.claude/settings.json`, removing hooks the user had configured under the installed version.

Observed output from a `uv run aelf wonder` in a worktree pinned to a v3.0.1-era base, against an installed v3.1.0:

```
aelfrice: hooks updated to v3.0.1 (was v3.1.0) —
  added: user_prompt_submit_retrieval;
  opted out: transcript_ingest, commit_ingest, session_start,
             stop_lock_prompt, search_tool, search_tool_bash
```

The user did not invoke `aelf setup`. The mutation was a side effect of `uv run aelf` resolving its `.venv` from the worktree's older source tree, then a setup-reconciler running automatically.

## Why this is a bug

1. **Global side-effect from a per-worktree command.** A worktree's `uv run aelf` should not be able to reach out and rewrite `~/.claude/settings.json`. The point of having a per-worktree `.venv` and a per-worktree source tree is that operations stay scoped to that worktree.
2. **Silent downgrade.** Six hooks (`transcript_ingest`, `commit_ingest`, `session_start`, `stop_lock_prompt`, `search_tool`, `search_tool_bash`) were removed without any consent prompt. The user is currently on installed v3.1.0 with the v3.1.0 hook surface; the worktree's `uv run aelf` opted them OUT on the grounds that the worktree's source advertises a smaller hook surface.
3. **Multi-worktree workflow is the supported case.** This repo's workflow creates many worktrees per repo, often pinned to different bases (rebase branches, release branches, feature branches off old tags). Every one is a footgun under current behavior — touching any worktree's `uv run aelf` re-pins the global hook config to whatever that worktree's source advertises.

## Repro

1. Have `aelf` installed at version N (e.g. v3.1.0) with `~/.claude/settings.json` carrying the v3.1.0 hook set.
2. From a worktree whose source tree is pinned to version M < N (e.g. on a `release/v3.0.1` base, or a feature branch off a v3.0.1-era commit):
   ```
   uv run aelf status
   ```
3. Observe: `~/.claude/settings.json` is rewritten. Hooks that exist in N but not in M are silently removed.

## Fix options

- **A — Skip auto-setup when running from a non-uv-tool venv.** Detect that the running `aelf` is resolved from a `.venv/` inside a project tree (compare `sys.executable` parent against `uv tool dir`), and never auto-run setup in that case. Users who want worktree-pinned hooks run `aelf setup` explicitly.
- **B — Refuse to downgrade without consent.** If the in-source hook surface is a strict subset of what's already on disk, abort the reconciler with a clear message: "in-worktree aelf is older than installed; refusing to remove hooks A, B, C — re-run with `--force` to override". The reconciler may still ADD hooks the in-source version knows about.
- **C — Never touch `~/.claude/settings.json` from a worktree-resolved aelf.** Strongest variant of A.

I'd start with B (preserves the useful "in-worktree aelf adds new hooks" case) or A (simplest).

## Scope

- Detect uv-tool vs uv-run-from-worktree invocation context.
- Gate the auto-setup / hook-reconciler on that context.
- Regression test: from a worktree with an older `pyproject.toml` aelfrice version pin, run any `uv run aelf` command; assert `~/.claude/settings.json` is byte-identical before and after.

## Out of scope

- The `demotion_pressure` IndexError — separate issue.
- Redesigning the hook surface itself.

robotrocketscience:
Closes #834.

## Bug

`auto_install_at_cli_entry` ran on every CLI invocation, including `uv run aelf <cmd>` from a project worktree's local `.venv`. With `~/.aelfrice/installed-manifest-version` stamped at one version and a worktree pinned to a different one (the routine multi-worktree workflow), the merge ran against the worktree's bundled manifest and rewrote the user's global `~/.claude/settings.json` — re-pinning hook commands to the worktree's `.venv` paths and re-stamping the version backwards. The "opted out: …" line in the bug report is a separate symptom: the user already had those six hooks in `~/.aelfrice/opt-out-hooks.json` (from a prior `aelf setup --no-X`); the merge correctly skipped them, but the surrounding stamp downgrade and settings rewrite happened without consent.

## Fix

Option A from the issue body. New `is_running_from_uv_tool_install()` delegates to the existing `lifecycle._is_uv_tool_install` (single-source detection: `~/.local/share/uv/tools/aelfrice/` exists, or `sys.prefix` resolves under the uv tools root). `auto_install_at_cli_entry` gates on it after the existing `AELFRICE_NO_AUTO_INSTALL` env check.

After this PR:

| context | auto-install fires |
| --- | --- |
| `aelf <cmd>` from `uv tool install aelfrice` | yes |
| `uv run aelf <cmd>` from a worktree `.venv` | **no** (was: yes) |
| contributor `pytest` against checked-out source | **no** (was: yes) |
| `python -m aelfrice.cli` from a system Python | **no** (was: yes) |
| `aelf setup` (explicit) | yes — `setup` calls `maybe_install_manifest` directly, gate doesn't apply |

Symmetric with the existing `AELFRICE_NO_AUTO_INSTALL=1` env override; the new gate is the implicit version of the same escape hatch for non-uv-tool contexts.

## Tests

Three new tests in `tests/test_auto_install.py`:

- `test_auto_install_at_cli_entry_skips_when_not_uv_tool` — gate False → no stamp written, no settings.json created. Direct regression reproducer.
- `test_is_running_from_uv_tool_install_delegates_to_lifecycle` — confirms both branches of the gate flow through `lifecycle._is_uv_tool_install` so detection stays in one place.
- `test_auto_install_at_cli_entry_runs_when_uv_tool` — gate True path still invokes `maybe_install_manifest`. Uses a stub rather than the real merge: `maybe_install_manifest` binds `STAMP_PATH` / `OPT_OUT_PATH` at function-definition time, so an end-to-end test would leak to the user's actual `~/.aelfrice/`.

`test_cli_auto_install.py::test_auto_install_at_cli_entry_does_not_propagate_exceptions` updated to monkeypatch the new gate to True so the post-gate exception-handling path is still exercised.

Full suite: 4271 passed, 62 skipped, 75 xfailed (`pytest -x -q --timeout=120`, run during pre-PR gate).

## Out of scope

- The `demotion_pressure` IndexError (#833 — being worked separately).
- Refactoring `maybe_install_manifest` so monkeypatching `auto_install.STAMP_PATH` / `OPT_OUT_PATH` propagates into the merge path. The default-arg binding is a real test-pollution hazard but not the user-facing bug; flagging here as a follow-up.
- Fix B from the issue body (refuse to downgrade without consent). Fix A is the strictly stronger guarantee — Fix A skips entirely; Fix B would still permit a uv-tool downgrade. The Out-of-Worktree contract is the load-bearing property.

## Summary by CodeRabbit

* **Bug Fixes**
  * Auto-install functionality now executes exclusively for `uv tool` installations, preventing unintended modifications to global configuration and settings files when running from project worktrees with incompatible or mismatched manifest versions. The existing `AELFRICE_NO_AUTO_INSTALL` environment variable override remains fully functional, and test coverage has been expanded to validate the new behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/836)


robotrocketscience:
[claim:review:prince:2026-05-15T01:49:10Z]

robotrocketscience:
Reviewed `51d1d2a8` → `0f78c23d` → `c40550a0`. Three atomic signed commits; CI green; `mergeState=CLEAN`.

**Fix is correct.** `is_running_from_uv_tool_install()` delegates to `lifecycle._is_uv_tool_install`, which is the existing single-source detection (`~/.local/share/uv/tools/aelfrice/` exists or `sys.prefix` resolves under the uv tools root). The local import inside the function avoids the early-CLI-bootstrap circular import. Gate fires after `AELFRICE_NO_AUTO_INSTALL` so the env override still short-circuits first; it's the implicit version of the same escape hatch for non-uv-tool contexts.

**Tests cover the right axes:**
- `test_auto_install_at_cli_entry_skips_when_not_uv_tool` — gate False → no stamp, no `settings.json`. Direct regression reproducer for #834. Monkeypatches `STAMP_PATH`/`OPT_OUT_PATH`/`USER_SETTINGS_PATH` to `tmp_path` so leakage to `~/.aelfrice/` is impossible.
- `test_is_running_from_uv_tool_install_delegates_to_lifecycle` — pinpoints the lifecycle seam, exercises both branches. Catches the failure mode where someone adds a second uv-tool-detection path that drifts from the single-source.
- `test_auto_install_at_cli_entry_runs_when_uv_tool` — gate True still invokes `maybe_install_manifest`. Stub-based, with a comment explaining why end-to-end isn't run (the `STAMP_PATH`/`OPT_OUT_PATH` are bound at function-definition time → real-merge leaks to user's `~/.aelfrice/`). That's a known sharp edge in the surrounding test architecture; the test takes the cautious path.
- Existing `test_auto_install_at_cli_entry_does_not_propagate_exceptions` was updated to monkeypatch the new gate to True — keeps the post-gate exception path exercised.

**Changelog entry** is accurate: notes that the "opted out: …" line in the bug report was a pre-existing `~/.aelfrice/opt-out-hooks.json` condition, not caused by the bug — the bug was the surrounding stamp downgrade and settings rewrite. Lands under `[Unreleased] → Fixed` (correct — main release queue will migrate when the next release dates this section).

**Behavior matrix from the PR body** matches my read of the diff:
- `aelf <cmd>` from `uv tool install` → fires (gate True)
- `uv run aelf` from worktree `.venv` → skipped (gate False, was: ran)
- contributor `pytest` → skipped (gate False, was: ran)
- `python -m aelfrice.cli` from system Python → skipped (gate False, was: ran)
- `aelf setup` → still works (calls `maybe_install_manifest` directly, bypasses the entry helper)

**Minor — out of scope, worth noting:**
- The function-definition-time binding of `STAMP_PATH` / `OPT_OUT_PATH` that forced the stub-based test pattern is a real testability foot-gun. Hoisting them to function kwargs (or reading at call time) would let future tests exercise the real merge path with `tmp_path` overrides. Not gating this PR; worth a small follow-up.
- Option B from the issue ("refuse to downgrade without consent") still has the standalone value the issue noted — but it's the right call to ship A first (smaller blast radius, plus B is additive to A). If we want to revisit, a separate issue for "warn-on-downgrade in `aelf setup` itself" would catch the explicit-invocation case too.

Adding `ready-to-merge`.

robotrocketscience:
[release:review:prince:2026-05-15T01:50:56Z]

PR #835, opened 2026-05-15
release: v3.2.0 — fix #833 + ship Unreleased (was: v3.1.1 release)

robotrocketscience:
Closes #833.

Replaces the prior `release: v3.1.1` shape on this branch after the review thread surfaced that tagging current `main` at v3.1.1 would be a SemVer violation: `[Unreleased]` includes `#814`'s Removed entry with four explicit breaking-API changes (`apply_feedback(propagate=)` kwarg → `TypeError`, `FeedbackResult.{pressured_locks,demoted_locks}` → `AttributeError`, `aelf locked --pressured` CLI flag, `aelf:feedback`/`aelf:confirm` MCP payload keys → `KeyError`). Patch-version bumps cannot ship breaking changes; the SemVer-correct shape is **v3.2.0**.

## What changed vs the prior PR shape

- `pyproject.toml` + `uv.lock`: `3.1.0` → **`3.2.0`** (was: `3.1.0` → `3.1.1`).
- `CHANGELOG/v3.md`:
  - Renames `## [Unreleased]` → `## [3.2.0] - 2026-05-15` and opens a new empty `## [Unreleased]` above (per top-level `CHANGELOG.md` policy: "On release, move `[Unreleased]` content into a dated `## [X.Y.Z]` section").
  - Prepends the `#833` Fixed entry to the new `[3.2.0]` `### Fixed` section.
  - Updates compare-link footers: `[Unreleased]: …v3.2.0…HEAD`, new `[3.2.0]: …v3.1.0…v3.2.0`. (Fixed a pre-existing drift in the same edit — the previous `[Unreleased]: …v3.0.1…HEAD` should have been `…v3.1.0…HEAD` after the v3.1.0 ship.)
- Regression test commit (`6c8963ef test: regression for #833 demotion_pressure post-DROP read`) — unchanged from `3d1aecbb`.

## What ships in v3.2.0

| Section | Issue | Summary |
| --- | --- | --- |
| Added | #816 | Hot-path touch state v1 storage substrate |
| Added | #817 | ζ posterior-rerank surface (bounded sigmoid contribution) |
| Added | #758 | Seventh meta-belief consumer: adaptive `posterior_temperature` |
| Fixed | **#833** | **`aelf` v3.1.0 crash on any DB previously touched by post-#814 code** |
| Fixed | #834 | Auto-install gated to uv-tool installs only |
| Fixed | #798 | Rebuilder pack accounting honours `use_type_aware_compression` |
[...]

Issue #814, opened 2026-05-14
refactor(feedback): remove unreachable CONTRADICTS-edge auto-demote path (per #605)


robotrocketscience:
## Summary

The CONTRADICTS-edge-driven auto-demotion path in `apply_feedback`
is **unreachable in practice** and **philosophically inconsistent
with the locked PHILOSOPHY decision #605**. Proposing removal.

## Direct evidence (production DB inventory)

Run against `~/projects/aelfrice/.git/aelfrice/memory.db`,
2026-05-14T20:25Z:

| metric | value |
|---|---|
| user locks | 14 |
| locks with `demotion_pressure > 0` | **0** |
| `CONTRADICTS` edges (any) | **0** |
| `CONTRADICTS` edges → a user-lock | 0 |
| positive feedback rows | 63,861 |

Edge-type histogram: `DERIVED_FROM=1915`, `RELATES_TO=84`,
`TESTS=9`, `CITES=1`. Zero `CONTRADICTS`. Across 63,861 positive
feedback events (each calling `_pressure_and_maybe_demote`), the
walk has terminated immediately every time because
`edges_from(source_id)` yields no `CONTRADICTS` rows.

## Mechanism (`src/aelfrice/feedback.py:65-103`)

Demotion fires only when:
1. A belief X has positive feedback applied with `propagate=True`.
2. X has an outbound `CONTRADICTS` edge to a `LOCK_USER` belief Y.
3. Y's `demotion_pressure` accumulates; at `DEMOTION_THRESHOLD=5`
   the lock drops to `LOCK_NONE`.

Step 2 is the failure mode. The only writers of `CONTRADICTS`
edges on `github/main` are:

- `src/aelfrice/triple_extractor.py:130-131` — regex on literal
  strings `"contradicts"` / `"disagrees with"` only. Too narrow
  to catch real cross-belief contradictions in ordinary onboarded
  content.
- `src/aelfrice/wonder/lifecycle.py:153` and
  `src/aelfrice/wonder/simulator.py:234` — fire only when
  `aelf wonder` explicitly flags contradiction. Wonder runs
  produce relatively little graph compared to ingest.

Result: across the project's full lifetime to date, no
`CONTRADICTS` edge has been produced that points at a locked
belief, so the demotion path is dead.

## Why this is the right call to delete, not to fix

Per locked belief `c06f8d575fad71fb`:

> v3.0 PHILOSOPHY decision (#605, ratified 2026-05-10):
[...]

PR #838, opened 2026-05-15
ci(release-docs-check): assert [Unreleased] drained on release PR (#837)

### LINKED ISSUE #837: ci(release-docs-check): assert [Unreleased] drained into dated section on release PR
## Summary

The `release-docs-check` job in `.github/workflows/staging-gate.yml` already detects release PRs (head version != base version) and asserts that `CHANGELOG/v${major}.md` contains a `## [${head_ver}]` header and a `[${head_ver}]: ...` compare-link footnote. It does **not** assert that `[Unreleased]` was actually drained into that new section.

This gap let PR #835 (`release: v3.1.1`) pass `release-docs-check` while leaving five `[Unreleased]` entries — `#816`, `#817`, `#798`, `#809`, `#814` — in `[Unreleased]` rather than migrating them to `[3.1.1]`. The PR surfaced as a blocked review (see #835 thread) for two compounding reasons:

1. **Project release policy violation** — top-level `CHANGELOG.md` is explicit: *"On release, move `[Unreleased]` content into a dated `## [X.Y.Z]` section and add the compare-link footnote at the bottom of the same file. CI checks both."* — but CI doesn't actually check the migration.
2. **SemVer break** — `#814` documents four explicit breaking-API removals (`apply_feedback(propagate=)` kwarg → `TypeError`, `FeedbackResult.{pressured_locks,demoted_locks}` → `AttributeError`, `aelf locked --pressured` CLI flag, `aelf:feedback`/`aelf:confirm` MCP payload keys). Tagging a *patch* version while those sit in `[Unreleased]` and ship in the diff between v3.1.0 and v3.1.1 is a SemVer break.

## Fix

Extend the `release-docs-check` job to assert the migration: between the `## [Unreleased]` header and the next `## [` header, only headers (`### Added` / `### Fixed` / `### Removed` etc.) and whitespace may exist. Any bullet content under `[Unreleased]` at release time = fail.

Suggested implementation (extends `.github/workflows/staging-gate.yml` `release-docs-check`):

```bash
# After the existing "## [${head_ver}]" assertion:
unreleased_body=$(awk '
  /^## \[Unreleased\]/ { in_unreleased=1; next }
  /^## \[/ && in_unreleased { exit }
  in_unreleased && NF > 0 && !/^###/ { print }
' "$changelog_file")
if [ -n "$unreleased_body" ]; then
    echo "::error file=${changelog_file}::Release PR detected but [Unreleased] still contains entries that should have moved into [${head_ver}]. Migrate them per the policy in top-level CHANGELOG.md."
    echo "Offending content:"
    printf '%s\n' "$unreleased_body" | head -20 | sed 's/^/  /'
    fail=1
fi
```

## Out of scope

- Detecting SemVer-mismatch at the version level (e.g. "patch bump but Removed entries with breaking-change language exist") — that's a harder text-classification problem. The Unreleased-empty check catches the symptom mechanically: if you're shipping breaking changes in `[Unreleased]`, they ride the next release regardless of bump shape, and the operator decides whether that's patch / minor / major.
- The independent `release-drafter / publish.yml` draft-promotion gap (covered by PR #805 / related work).

## Tier

`rook` — small CI-job patch, contained to one workflow file plus its regression coverage. The full fix would also add a self-test (e.g. a fixture file with non-empty `[Unreleased]` and a synthetic release PR that the job rejects), but the workflow itself is a few lines of awk.

## Refs

- PR #835 — the release PR that surfaced the gap.
- Top-level `CHANGELOG.md` — the project policy this gap fails to enforce.
- `.github/workflows/staging-gate.yml` `release-docs-check` job (lines ~145-175) — the surface to extend.

robotrocketscience:
Closes #837.

## Bug

`release-docs-check` in `.github/workflows/staging-gate.yml` detects release PRs and asserts the `## [X.Y.Z]` header and compare-link footnote exist in `CHANGELOG/v{major}.md`, but does **not** assert that `[Unreleased]` was actually drained into the new dated section. The project policy (top-level `CHANGELOG.md`) explicitly says CI checks both; only the dated-section half was implemented.

The gap surfaced on PR #835 (`release: v3.1.1`): the PR reached `attn:review` while five `[Unreleased]` entries — including `#814`'s breaking-change `Removed` block — sat undrained, which would have shipped under a SemVer-patch tag.

## Fix

Extend the `release-docs-check` job with one awk pass over the changelog file:

```awk
/^## \[Unreleased\]/ { in_unreleased=1; next }
/^## \[/ && in_unreleased { exit }
in_unreleased && NF > 0 && !/^###/ { print }
```

- Sets `in_unreleased=1` on the `## [Unreleased]` header
- Exits at the next `## [` header (typically `## [X.Y.Z]`)
- Inside the window: prints any line that is non-whitespace AND not a `### ...` subsection header

Output non-empty → bullet / paragraph content remains in `[Unreleased]` → fail with an actionable error message that points at the policy and previews the offending lines (head -20).

Empty `### Added` / `### Fixed` / `### Removed` scaffolding is preserved (common pattern for clarity); only entries-under-headers count as un-drained.

## Verification

Smoke-tested the awk logic locally against three cases:

| case | input | awk output | verdict |
| --- | --- | --- | --- |
| current `CHANGELOG/v3.md` | non-drained `[Unreleased]` | first 3 bullets surface | would fail a release PR (correctly) |
| synthetic drained | `## [Unreleased]\n\n### Added\n\n### Fixed\n\n## [9.9.9]...` | empty | passes (correctly) |
| synthetic non-drained | as above + `- leftover entry` under `### Added` | `- leftover entry` | fails (correctly) |

The PR itself does not bump `pyproject.toml`, so the existing `head_ver == base_ver` short-circuit at the top of the job means this PR's own `release-docs-check` run will exit 0 (correctly — this is a CI-config PR, not a release PR).

## Out of scope

- SemVer-level guard ("patch bump shipping `Removed` entries with breaking-change language"). Detecting that mechanically requires text classification; the Unreleased-drain check catches the symptom at the release boundary instead.
- Lifting the awk to a standalone `scripts/check-changelog-drained.sh` with its own unit tests. Worth doing if this check grows more clauses; currently 4 lines of awk is small enough to keep inline.
- The independent `release-drafter / publish.yml` draft-promotion gap (separate work).

## Tier

`rook` — single-workflow patch, 21 lines added, no production code touched.

### DIFF
--- a/.github/workflows/staging-gate.yml
+++ b/.github/workflows/staging-gate.yml
@@ -165,6 +165,27 @@ jobs:
               echo "::error file=${changelog_file}::Missing compare-link footnote '[${head_ver}]: https://github.com/...' near bottom of this file."
               fail=1
             fi
+            # #837 — [Unreleased] must be drained on release. Between the
+            # `## [Unreleased]` header and the next `## [` header, only
+            # `### ...` subsection headers and whitespace may remain.
+            # Any bullet / paragraph content = entries that should have
+            # moved into the dated section (per top-level CHANGELOG.md
+            # policy: "On release, move [Unreleased] content into a
+            # dated ## [X.Y.Z] section"). Catches SemVer breaks like a
+            # patch release shipping breaking changes left in
+            # [Unreleased] (the original failure mode that motivated this
+            # check: PR #835 review thread).
+            unreleased_body=$(awk '
+              /^## \[Unreleased\]/ { in_unreleased=1; next }
+              /^## \[/ && in_unreleased { exit }
+              in_unreleased && NF > 0 && !/^###/ { print }
+            ' "$changelog_file")
+            if [ -n "$unreleased_body" ]; then
+              echo "::error file=${changelog_file}::Release PR detected (${base_ver} -> ${head_ver}) but [Unreleased] still contains content. Migrate it into [${head_ver}] (or split the unmigrated entries into a follow-up release) before merging."
+              echo "Offending [Unreleased] content (first 20 lines):"
+              printf '%s\n' "$unreleased_body" | head -20 | sed 's/^/  /'
+              fail=1
+            fi
           fi
           if grep -qE "v?${head_ver}\s*\|\s*(next|planned)\b" README.md; then
             echo "::error file=README.md::Roadmap row for v${head_ver} still says 'next' or 'planned'. Flip status to 'shipped' before cutting the release."


robotrocketscience:
[claim:review:oppenheimer:2026-05-15T02:58:17Z]

robotrocketscience:
[claim:review:bagheera:2026-05-15T02:58:28Z]

robotrocketscience:
[release:review:bagheera:2026-05-15T02:58:32Z]

robotrocketscience:
## Approve — clean CI guard, semantically correct

21-line patch to `release-docs-check`, single signed commit (`75d27b38` G), FF on current main, `mergeable=CLEAN`, discretion grep on full diff = 0 hits. Gated by the existing `head_ver != base_ver` short-circuit so non-release PRs (this one included) are unaffected.

**Awk smoke-tested locally against four cases — all correct:**

| input | awk output | verdict |
| --- | --- | --- |
| current `CHANGELOG/v3.md` | bullets from `[Unreleased]` § Fixed / Added | would correctly fail a release PR today |
| `[Unreleased]` with empty `### Added` / `### Fixed` scaffolding only | empty | passes |
| `[Unreleased]` with `- leftover entry` under `### Added` | `- leftover entry` | fails (correct) |
| changelog with no `[Unreleased]` header at all | empty | passes (correct) |

Edge cases I worked through and are handled:

- `[Unreleased]` is the last section in the file (no following `## [`): awk's exit clause never fires but printing reaches EOF correctly. No false positive.
- Bullet lines containing `### foo` mid-line (e.g., `- the ### foo case`): line starts with `-`, not `###`, so `!/^###/` correctly keeps it as content. No false negative.
- The `next` after setting `in_unreleased=1` correctly prevents the `## [Unreleased]` header itself from being printed.

One forward-looking note worth flagging on the next iteration (not a blocker): if anyone ever adds a `## Notes` or other top-level non-version `## ` section *inside* `[Unreleased]`, awk would treat its body as undrained content. Not realistic in keepachangelog format; calling out so reviewers know it's known.

Adding `ready-to-merge`. Releasing claim.

robotrocketscience:
[release:review:oppenheimer:2026-05-15T03:00:23Z]

PR #862, opened 2026-05-18
feat(hook): session-scoped retrieval exclusion via aelf scope-out (#856)

robotrocketscience:
Closes #856. Defect 1 of #855.

## Summary

The UserPromptSubmit memory hook now honors a per-session retrieval-exclusion list. `aelf scope-out <pattern>` (slash form `/aelf:scope-out <pattern>`) lets a user say "stop bringing this up for the rest of the session" without hard-deleting the beliefs.

## Mechanism

1. **Storage** — `<git-common-dir>/aelfrice/session_exclusions.json`, sibling of `session_first_prompt.json`. Format: `{"session_id": "<sid>", "patterns": ["substr1", ...]}`.
2. **CLI** — `aelf scope-out <pattern>` resolves the active `session_id` from `session_first_prompt.json` and appends to the list (idempotent). Sibling flags `--list` and `--clear`.
3. **Hook filter** — after `_retrieve()` and before any telemetry / dedup / formatting, `_filter_session_exclusions(hits, session_id)` drops every belief whose content contains a listed pattern (case-insensitive literal substring).
4. **Auto-clear on new session** — `load_exclusions` returns `[]` when stored `session_id` ≠ current. A fresh session starts clean.

## Design choices

- **Substring, not regex.** Aligned with the v3.0 PHILOSOPHY (#605) — deterministic stdlib-only retrieval. A future `--regex` flag can extend if a concrete need surfaces.
- **L0 (locked) beliefs are filtered too** on the per-turn retrieval path. When the user explicitly says "stop bringing this up", that overrides ground-truth re-injection for the session. The lock stays in the DB; only injection is suppressed.
- **SessionStart sub-block is NOT filtered.** It dumps L0 on the very first prompt of a session, which runs before the user could have typed `/aelf:scope-out`. Captured in the test docstring.

## Files

- New module: `src/aelfrice/session_exclusions.py`
- New slash file: `src/aelfrice/slash_commands/scope-out.md`
- CLI subparser + handler: `src/aelfrice/cli.py`
- Hook filter: `src/aelfrice/hook.py`
- `EXPECTED_COMMANDS` updated: `tests/test_slash_commands.py`
- CHANGELOG entry: `CHANGELOG/v3.md`

## Tests

[...]

Issue #857, opened 2026-05-18
feat(hook): report retrieval coverage to surface hook/index recall asymmetry (#855 defect 2)


robotrocketscience:
## Summary

Sub-issue of #855 — defect 2 (retrieval vs index recall asymmetry).

The per-turn memory hook surfaces a small biased subset of relevant beliefs. Users reasonably assume "what I have seen in chat is what is in there" and clean based on visibility. The keyword index has much broader recall. Cleanup driven by chat-surface visibility is structurally incomplete.

## Proposed mechanism

After every retrieval, the hook emits a coverage line:

    retrieved N of M matching beliefs for this query;
    run `aelf search <topic>` to see the rest.

Where N = beliefs included in injection, M = total beliefs matching the same keyword scoring at any non-zero relevance. Surfacing the delta makes the asymmetry visible to the user before they assume chat is exhaustive.

## Acceptance items

- [ ] Hook computes M = total candidate matches alongside N = surfaced.
- [ ] Hook output includes the coverage line when M > N.
- [ ] Tests: M == N (line omitted), M > N (line shown), M == 0 (no-op edge case).
- [ ] Coverage line text is privacy-safe — no belief content, just counts and the topic the user already typed.

## Tier

`rook` — small, localized to hook output formatter + candidate-count computation.

## Parent

#855 — three-defect bundle. This issue handles defect 2 only.

robotrocketscience:
[claim:issue:Kulili:2026-05-18T16:38:46Z]

robotrocketscience:
[claim:issue:idnn:2026-05-18T16:48:42Z]

robotrocketscience:
[release:issue:idnn:2026-05-18T16:48:52Z]

robotrocketscience:
[claim:issue:Gylf:2026-05-18T16:56:41Z]

robotrocketscience:
[release:issue:Gylf:2026-05-18T16:56:50Z]

robotrocketscience:
[release:issue:Kulili:2026-05-18T17:08:58Z]

Issue #856, opened 2026-05-18
feat(hook): session-scoped belief exclusion via aelf:scope-out (#855 defect 1)


robotrocketscience:
## Summary

Sub-issue of #855 — defect 1 (no conversational suppression channel).

Today the UserPromptSubmit memory hook has no way for a user to say "stop bringing up topic X for the rest of this session" short of hard-deleting the matching beliefs. The hook re-injects on every turn, pattern-matching the conversation back to X even after the user has told the agent to drop it.

## Proposed mechanism

A new slash command `aelf:scope-out <pattern>` writes a session-local file (e.g. under the active session dir) holding one exclusion pattern per line. The memory hook reads this file on every fire and excludes any belief whose text or topic matches a listed pattern from the retrieval result.

## Acceptance items

- [ ] `aelf:scope-out <pattern>` slash command — appends pattern to session exclusion file. Idempotent (no duplicates).
- [ ] `aelf:scope-out --list` shows current exclusions; `aelf:scope-out --clear` empties the file.
- [ ] Memory hook reads the file and filters retrieval before formatting injection.
- [ ] Session-end cleanup (or natural cleanup via session-dir TTL).
- [ ] Tests: empty file (no-op), single pattern, multiple patterns. Pick literal-vs-regex semantics and document.

## Tier

`queen` — new CLI surface + hook integration + session lifecycle.

## Parent

#855 — three-defect bundle. This issue handles defect 1 only.

robotrocketscience:
[claim:issue:Setr:2026-05-18T16:41:48Z]

robotrocketscience:
[claim:issue:Gylf:2026-05-18T16:56:25Z]

robotrocketscience:
[release:issue:Gylf:2026-05-18T16:56:34Z]

robotrocketscience:
[release:issue:Setr:2026-05-18T17:02:52Z]

Issue #855, opened 2026-05-15
Memory retrieval cannot be scoped-out by user intent, and silently under-samples relative to the keyword index


robotrocketscience:
## Summary

The UserPromptSubmit memory hook re-injects beliefs about a topic on every turn, even after the user has explicitly told the agent in conversation to stop bringing up that topic. There is no mechanism for in-session suppression. Additionally, the per-turn hook surfaces a much smaller subset of relevant beliefs than `aelf search` returns for the same keyword, so users who clean up based on what they see in chat leave most of the target population behind.

## Repro

1. Have a non-trivial number of beliefs tagged with topic X (any common keyword).
2. Start a new session. The hook retrieves N beliefs per turn, biased to some subset (relevance? recency? salience? — opaque from the user side).
3. Tell the agent: "stop bringing up X."
4. Send a follow-up unrelated prompt. The hook fires, re-injects beliefs about X, and the agent pattern-matches the conversation back to X.
5. Run `aelf search X`. Observe that the result set is materially larger than the union of what's been surfaced in the conversation.

## Three distinct defects bundled here

1. **No conversational suppression channel.** There is no way for the user (or the agent on the user's behalf) to add "exclude beliefs matching pattern P" to the active session's retrieval, short of hard-deleting the beliefs. The UserPromptSubmit hook cannot react to prior assistant or user turns.

2. **Retrieval vs. index recall asymmetry.** The per-turn hook surfaces a small biased subset of relevant beliefs. The user reasonably assumes "what I've seen in chat is what's in there" and cleans based on that. The keyword index has much broader recall. Cleanup driven by chat-surface visibility is structurally incomplete.

3. **No project-scoped or tag-scoped retrieval.** Beliefs from one project context keep surfacing when the user is working in a different worktree on a re-scoped version of the same project. The hook has no notion of active scope.

## Impact

The user-visible failure mode is an agent that appears unable to follow direct instructions. From the user's seat it looks like the agent is being stubborn. The actual mechanism is that the agent is being structurally re-poisoned with the topic on every turn and cannot stop it. Worst possible failure mode for a memory system that is supposed to be the agent's collaborator.

## Proposed mitigations, ranked smallest to largest

1. **Session-scoped exclusion list.** A command like `aelf:scope-out <pattern>` writes a session-local file the hook reads and uses to filter retrieval for the rest of the session. Cheap; addresses defect 1 directly. Cleared on session end so it does not accumulate.

[...]

PR #864, opened 2026-05-18
feat(hook): project-context retrieval filter (#858)

### LINKED ISSUE #858: feat(hook): project-scoped retrieval filtering — surface beliefs only from active scope (#855 defect 3)
## Summary

Sub-issue of #855 — defect 3 (no project-scoped or tag-scoped retrieval).

Beliefs from one project context keep surfacing when the user is working in a different worktree on a re-scoped version of the same project. The memory hook has no notion of active scope.

## Existing surface

`beliefs.scope TEXT NOT NULL DEFAULT 'project'` already exists on the schema (added in `src/aelfrice/store.py` via commit 988bf19e). Today it is used for:

- user/project promotion lifecycle (`scope='user'` vs `'project'`)
- federation routing (`scope='global'` for peer-shared beliefs, `scope=<peer name>` for peer-imported beliefs)

It is **not** used by the memory hook to filter retrieval against an "active worktree" or "active project context".

## Open question — overlap with existing scope field

The current `scope` column is overloaded. Adding worktree-context filtering may conflict with the existing user/project/federation semantics. Possible approaches:

1. Reuse `scope` and add a hook-side filter: only return beliefs where `scope IN ('user', '<current-project-scope>', 'global')`. Requires defining what `<current-project-scope>` is and how it is discovered (cwd? env var? explicit slash command?).
2. Add a separate `project_context` or `tag` column for per-worktree namespacing, leaving `scope` to mean what it already does.
3. Add a tag/labels table for many-to-many tagging, with hook filter by active-tag set.

Operator triage requested before claiming.

## Acceptance items

(Sketch — pending operator decision on mechanism above.)

- [ ] Hook filters retrieved beliefs by active project context.
- [ ] Active-context discovery mechanism chosen and documented.
- [ ] Tests for cross-worktree isolation.
- [ ] No regression on user/project promotion or federation paths.

## Tier

`queen` — design decision required before implementation.

## Parent

#855 — three-defect bundle. This issue handles defect 3 only.

robotrocketscience:
Closes #858. Defect 3 of #855.

## Summary

The UserPromptSubmit hook now filters retrieved beliefs by the active within-repo project context. Cross-repo isolation already exists via `db_path()`'s `--git-common-dir` routing (two repos → two DBs → no contamination); this PR catches the within-repo re-scope case (same DB, same repo, but the project's conceptual identity has shifted and old-scope beliefs should no longer surface).

## Mechanism

1. **Schema** — new `beliefs.project_context TEXT NOT NULL DEFAULT ''` column + `idx_beliefs_project_context`. Default `''` preserves pre-#858 retrieval semantics: legacy rows always pass the filter.
2. **Resolver** — `db_paths.active_project_context()` reads `$AELFRICE_PROJECT_CONTEXT`, strips whitespace, returns `''` when unset / blank. Public constant `PROJECT_CONTEXT_ENV` exported so callers don't depend on a magic string.
3. **Belief dataclass** — `project_context: str = ""` field; `_row_to_belief`, `insert_belief`, `update_belief` round-trip it.
4. **Hook filter** — `_filter_by_project_context(hits)` applied post-`_retrieve()`. Rule:
   - Empty active context → no-op (preserves pre-#858 behaviour for sessions that have not opted in).
   - `scope != 'project'` (federation `global` / `shared:*` or user-promoted `user`) bypasses the filter.
   - Project-scope rows: keep iff `project_context == '' OR project_context == active`.

## Why post-`_retrieve()` filter and not a SQL WHERE clause

The retrieval surface is layered (L0 locks, L2.5 entity-index, L1 BM25, L3 BFS). Filtering once after the orchestrator collapses everything keeps the "which tier sees what" matrix trivial and includes federation peer hits (`search_peer_beliefs`) under the same scope check.

## Why `project_context` is distinct from `scope`

`scope` (#688) controls federation visibility — whether peer DBs may surface the row. `project_context` controls within-repo retrieval namespacing — whether the hook surfaces it in the active in-process context. Orthogonal questions; read-only federation (#661) does not interact with `project_context`.

## Design choices

- **Env var only at this commit.** State-file fallback (`.aelfrice/context` at git-common-dir set by an `aelf context set <name>` CLI subcommand) deferred to a follow-up issue. Env var alone is sufficient for Claude Code sessions and CI pipelines that set context at startup; the resolver location and the public env-var name fix the contract so a later commit can add state-file fallback inside the same function without changing callers.
- **Write-side stamping is opt-in.** Callers that want a belief tagged set `b.project_context` before `insert_belief` / `update_belief`. Legacy writers fall through to the `''` default unchanged. Auto-stamping from the resolver at write time was considered and deferred — it conflates "what context was active when this was ingested" with "what context should this row be visible in", and the bench / federation paths shouldn't be silently re-tagged by an env var being set elsewhere in the process.
- **Empty-input fast path** in `_filter_by_project_context` skips the env-var read when retrieval returned nothing. The resolver is itself cheap, but skipping it removes the only side effect from the hot path.

## Files

- Schema + writes: `src/aelfrice/store.py`
- Resolver: `src/aelfrice/db_paths.py`
- Dataclass: `src/aelfrice/models.py`
- Hook filter: `src/aelfrice/hook.py`
- CHANGELOG entry: `CHANGELOG/v3.md`

## Tests

- `tests/test_project_context_field.py` — 9 tests: column on fresh DB, index present, default `''`, migration idempotent on reopen, migration adds column to legacy pre-#858 DB, Belief default `''`, insert round-trip, update round-trip, search-result round-trip.
- `tests/test_active_project_context.py` — 6 tests: unset / empty / whitespace returns `''`, simple value, surrounding-whitespace strip, public-constant pin.
- `tests/test_hook_project_context_filter.py` — 10 tests: empty input fast path, no-active-context no-op, legacy `''` always passes, matching active kept, mismatching dropped, `user` / `global` / `shared:peer` bypass, whitespace-active no-op, input-order preserved.
- Full pytest suite: 4385 passed, 34 skipped, 75 xfailed. No regressions.

## Acceptance items (from #858)

- [x] Hook filters retrieved beliefs by active project context.
- [x] Active-context discovery mechanism chosen and documented (env var, with state-file fallback explicitly deferred).
- [x] Tests for cross-worktree isolation (full filter matrix + schema migration on legacy DB).
- [x] No regression on user/project promotion or federation paths (full suite green; `scope != 'project'` bypass test covers `user` / `global` / `shared:*` explicitly).

## Out of scope (other defects of #855)

- Defect 1 (session-scoped suppression via `aelf scope-out`) → #856 / PR #862.
- Defect 2 (retrieval-vs-index recall coverage line) → #857.

## Out of scope (follow-up for this defect)

- `aelf context set/get/clear` CLI subcommand + `.aelfrice/context` state-file fallback. Tracked as a follow-up.
- Auto-stamping `project_context` on write from the resolver. Deferred pending a concrete use case that disambiguates "active-at-ingest" from "visible-in".

## Test plan

- [x] Schema migration unit tests pass (fresh + legacy DB)
- [x] Resolver unit tests pass (env-var matrix)
- [x] Hook filter unit tests pass (full filter matrix)
- [x] Full pytest suite passes (4385 passed)
- [x] Discretion grep clean on diff + commit messages

## Summary by Sourcery

Scope belief retrieval within a repository by introducing a project-context tag and filtering hook results against an env-configured active context, while preserving existing cross-repo and cross-context behavior by default.

New Features:
- Add a project_context field to beliefs and index it to support within-repo project scoping of retrieval results.
- Introduce an active_project_context resolver that reads the AELFRICE_PROJECT_CONTEXT environment variable to determine the current project context.
- Filter UserPromptSubmit hook retrieval results by project_context for project-scoped beliefs, bypassing non-project scopes and maintaining order.

Enhancements:
- Extend Belief persistence and migrations to round-trip the project_context field without impacting legacy data semantics.

Documentation:
- Document the new project-context retrieval filter and configuration in the v3 changelog.

Tests:
- Add tests covering schema migration and indexing for project_context, env-based active context resolution, and the hook-level project-context filter matrix including edge cases.


robotrocketscience:
[claim:review:Idnn:2026-05-18T18:33:53Z]

robotrocketscience:
[claim:review:Kulili:2026-05-18T18:36:14Z]

robotrocketscience:
[release:review:Kulili:2026-05-18T18:36:19Z]

robotrocketscience:
Review pass on PR #864 (#858 defect 3, project-context retrieval filter).

**Verdict: ✅ ready-to-merge.**

### Verified
- **Migration safety.** `ALTER TABLE beliefs ADD COLUMN project_context TEXT NOT NULL DEFAULT ''` runs through the existing `_MIGRATIONS` loop (`store.py:850-865`) whose `except sqlite3.OperationalError` swallows `duplicate column name` for idempotent re-runs. `_maybe_apply_content_hash_unique` (#219 one-shot table swap) reads columns via `PRAGMA table_info`, so it will preserve the new column on legacy DBs that hit it after the migration adds it. `migration-policy-check` CI green.
- **Round-trip integrity.** `insert_belief` and `update_belief` both updated; `_row_to_belief` defaults missing column to `''` (handles projected column lists). `get_belief` / `search_beliefs` both round-trip project_context (tests `test_insert_belief_round_trip`, `test_search_beliefs_returns_project_context_field`).
- **Filter contract.** Empty-active no-op (preserves pre-#858 retrieval); `scope != 'project'` bypass (user / global / shared:* surface unchanged — `scope` and `project_context` stay orthogonal); legacy `project_context=''` always passes; matching active kept, mismatching dropped. Order-preserving. Empty-input fast path avoids the `os.environ.get` call. Full matrix in `test_hook_project_context_filter.py` (10 tests).
- **Discretion grep on the full PR diff (`origin/main...HEAD`): clean.**
- **All 5 commits SSH-signed (`G`) with conventional-commit prefixes.**
- **CI:** pytest (3.12 + 3.13), CodeQL (python + actions), Staging Gate (all subchecks), bench-smoke, migration-policy-check, deptry, vulture — all SUCCESS. Sourcery commented PASS.
- **Full pytest suite: 4385 passed, 34 skipped, 75 xfailed — no regressions** (per PR body, matches local recent run).

### Notes (non-blocking)
- `_filter_by_project_context` docstring closes with "a peer's local-only row is not visible to us in any context" — that property is actually enforced by the federation layer, not by this filter (which only sees rows the federation already surfaced). Not a bug; the prose just over-attributes. Leave or tidy later.
- 603 / 8 over advisory size cap; 470 of the 597 added lines are tests. Acceptable.
- State-file fallback (`.aelfrice/context` + `aelf context set`) and auto-stamping at write deliberately deferred — call out in the PR body, contract pinned via `PROJECT_CONTEXT_ENV` constant.

Labelling `ready-to-merge`.

robotrocketscience:
[release:review:Idnn:2026-05-18T18:36:46Z]

robotrocketscience:
[claim:review:Kulili:2026-05-18T18:36:55Z]

robotrocketscience:
Concur with prior verdict (`ready-to-merge`).

One non-blocking docstring nit, follow-up-only:

`_filter_by_project_context` docstring (`src/aelfrice/hook.py:1456-1457`):

> `scope != 'project'` (federation 'global' / 'shared:*' or user-promoted 'user') bypasses the filter.

There is no `scope='user'` value. `BELIEF_SCOPE_RE` (`src/aelfrice/models.py:260`) admits only `^project$|^global$|^shared:[a-z0-9_-]+$`. The user-vs-agent axis is `lock_level` (`LOCK_NONE` / `LOCK_USER`). The runtime predicate `b.scope != BELIEF_SCOPE_PROJECT` is correct as-is; only the parenthetical prose is wrong and inherited the misconception from #858's body.

Will file a follow-up to fix this docstring line and post a body-correction comment on the issue. Not gating this PR.

Verified independently:
- migration (ALTER + index) is additive via existing `_MIGRATIONS` loop with `OperationalError` swallow for re-runs — safe on legacy DBs.
- `_filter_by_project_context` predicate matches the rule documented in the PR body.
- `scope` semantics (federation visibility, #688) untouched; `project_context` is purely additive on the within-repo retrieval axis.
- Discretion grep clean.

Labelling `ready-to-merge`.

robotrocketscience:
[release:review:Kulili:2026-05-18T18:38:38Z]

robotrocketscience:
**Review verdict: changes requested.** (Posting as comment because the GitHub user is shared across sessions.)

Requesting changes on the discovery layer before merge.

Operator decision (just made on the umbrella triage): the active-context discovery mechanism for #858 is **a `.aelfrice/context` file at the repo root**, not the `AELFRICE_PROJECT_CONTEXT` env var. The rationale was specifically that the operator wants this to be *automatic on first onboard* — env vars don't survive shell restarts and can't be persisted by `aelf onboard` without writing into `~/.zshrc` (footgun). The schema, dataclass plumbing, hook filter, and 25-test matrix in this PR are all sound — the changes below are scoped to the resolver and the onboard side.

**Requested changes:**

1. **Replace `active_project_context()` resolver.**
   - Walk upward from `os.getcwd()` looking for `.aelfrice/context` (single-line tag, whitespace-stripped). Stop at the git-common-dir if found, else at filesystem root. Return `""` if no file found (preserves existing "empty = no filter" semantic, full backward compat).
   - Drop the `PROJECT_CONTEXT_ENV` constant and the env-var read. Rename the resolver if helpful, but the existing call sites (`_filter_by_project_context`, the write-side stampers) keep their signature.
   - Cache the resolution per-process (cheap, but per-hook-call cost is real — one file `open()` + a few `stat()`s per UserPromptSubmit).

2. **Wire `aelf onboard` to write a default `.aelfrice/context`.**
   - If the file does not exist at the target dir's git-common-dir, write a default tag. Proposal: `<basename(repo_root)>-<short-hash-of-repo-root-abs-path>` so two worktrees of the same repo get the same tag (shared identity), but two unrelated repos can't collide. Print the chosen tag so the operator sees it.
   - If the file already exists, leave it alone.
   - Document in `aelf onboard --help` and the onboard slash file.

3. **Tests to add / move:**
   - `tests/test_active_project_context.py` — rename to drop env-var cases; add file-walk cases (file at cwd, file at parent dir, no file present, file with whitespace, file empty).
   - One end-to-end test that `aelf onboard <tmp>` writes `.aelfrice/context` and the hook filter then honours it.

4. **CHANGELOG block update.**
   - The current line says active context is read from `$AELFRICE_PROJECT_CONTEXT`. Update to "read from `<repo>/.aelfrice/context` (written automatically by `aelf onboard`)."
   - Drop the "State-file fallback deferred" sentence — file is now primary, not fallback.

**Out of scope (don't expand here):**
- `aelf context set <tag>` CLI subcommand — the file-walk + onboard-auto-write covers the workflow without it. File one as a follow-up if needed.
- Env-var as an optional override — the operator picked file-based; adding env-var alongside re-introduces the discoverability split that motivated the decision.

The rest of the PR (schema, index, `_row_to_belief`, insert/update roundtrip, hook integration point, test matrix structure) is good as-is.

Issue #871, opened 2026-05-19
feat: P2 cadence (ctx-threshold + phase-boundary) — Stop nudge + UPS-resume injection (#749 P2)


robotrocketscience:
## Summary

Implements **#749 P2 (token-budget watermark)** plus two operator-facing
consumers: a Stop-hook nudge at phase-boundary + ctx-threshold moments,
and a UserPromptSubmit-side cadence-resume injection that re-seeds a
fresh session with the rebuilder synthesis from the just-cleared one.

Sibling of PR #869 (P1 every-K-turns) and issue #870 (UPS-side P1
sibling). P2 differs from P1 on **trigger surface**: P1 fires every K
turns deterministically; P2 fires when the transcript byte-count
crosses a watermark *and* the recent user turn looks like a phase
boundary.

## Why phase-boundary AND threshold (not OR)

#749's body asks: "can we make manual `/clear` unnecessary?" The honest
read on Claude Code's hook surface (researched 2026-05-19) is: **no
hook can clear the conversation window**. No `clearHistory` /
`compactNow` / `truncateWindow` output field exists. `continue: false`
suspends with full history intact — not a clear.

Given that constraint, P2's goal narrows to: **identify the moment when
a manual `/clear` will lose the least state**. Two conditions:

1. **Ctx is high enough that clearing is worthwhile.** Below ~50% of the
   watermark, the operator isn't paying a meaningful tax to keep the
   session running.
2. **Recent user prompt is a phase boundary.** Mid-task `/clear` loses
   the active thread; phase-boundary `/clear` does not. Boundary
   indicators: short ack messages ("done", "thanks", "next", "ship
   it", "perfect", etc.) and explicit transition phrases ("switch to
   X", "let's move on").

AND logic — both required. OR would over-fire on every short
acknowledgment regardless of ctx, or under-fire across long mid-task
stretches.

## Trigger details

**Token estimation.** Hooks have no access to Claude Code's ctx%
(researched 2026-05-19; no payload field, no statusline hook). Aelfrice
estimates from `transcript_path` size in bytes — deterministic per
#605 (same transcript bytes → same fire decision), independent of
harness internals.

**Watermark default.** `ctx_byte_window = 600_000` ≈ 200K tokens at
~3 bytes/token. Configurable per-project via `.aelfrice.toml`.
Calibration ratio is approximate; the operator-side ctx% reading
serves as ground truth and the threshold gets re-tuned after
empirical use.

[...]

PR #869, opened 2026-05-19
feat(hook): P1 every-K-turns Stop-hook cadence (#749)

robotrocketscience:
## Summary

Ships P1 every-K-turns cadence policy (#749), the smallest of the three pre-registered policies in the issue body. The maintainer-self-report on issue comment 4491022396 tripped the body's OR-clause ("manual `/clear` still required at meaningful rate even with #746 shipped"); the empirical fire-rate audit (6 PreCompact fires across 2 sessions over 6 days post-#746) confirms P0 PreCompact-only is insufficient at the maintainer's workload shape.

## Design ratifications

Surfaced under #749 comment 4491022396 + answered in-session:

1. **Policy order:** P1 every-K-turns first. Cheapest to ship, easiest to falsify, deterministic via fire_idx.
2. **Cadence parameter encoding:** static config + `[cadence]` TOML knob, no #480 meta-belief integration yet. Resolver shape leaves a slot for `meta:hook.cadence_k` to land later without call-site changes.
3. **K default:** 15 turns (body-suggested literature-typical starting point).
4. **Default-OFF:** `[cadence] enabled = false` by default; operator opts in.

P2 (token-budget watermark) and P3 (turn-density-aware) per the issue body remain unimplemented. The policy enum is forward-compatible so those land as new values without rewriting the resolver or predicate.

## Commits (atomic by category)

1. `feat(cadence): cadence-policy module — P1 every-K-turns scaffold (#749)` — new `src/aelfrice/cadence.py` with `CadenceConfig` dataclass, env > kwarg > TOML > default resolvers (`resolve_cadence_enabled` / `resolve_cadence_policy` / `resolve_cadence_k`), and the pure `should_fire(fire_idx, config)` predicate (`fire_idx > 0 AND fire_idx % k == 0`, gated on `enabled AND policy == "p1_every_k_turns"`).
2. `test(cadence): unit tests for resolvers + firing predicate (#749)` — 30 tests in `tests/test_cadence.py` covering predicate cross-product (enabled / policy / k including k=0/k=1/negative/zero-fire_idx cold-start guard), TOML fail-soft on missing-file/section/malformed/wrong-type/unknown-policy/non-positive-k/bool-as-k, and resolver precedence including unparseable-env falls-through.
[...]

Issue #749, opened 2026-05-13
Tracker: cadence-policy R&D campaign — deferred pending PreCompact fire-rate data after #746 ships


robotrocketscience:
## Summary

Placeholder tracker for a checkpointing **cadence** R&D campaign,
deferred pending real-world PreCompact fire-rate data with
`[rebuilder] trigger_mode = "threshold"` enabled in user
`.aelfrice.toml`.

Filed as part of the diagnostic chain that produced #746 (flip
rebuilder default to threshold). This issue is itself work-not-yet-
scoped — exists so the question doesn't drop on the floor.

## The phase question

> What cadence policy maximizes session-state recovery quality
> while staying under M ms/turn overhead on the hook hot-path?

Equivalent operator-facing question: even with PreCompact rebuilder
enabled (#746 once that flips), does the harness fire it often
enough to make manual `/clear` unnecessary, or does the project
need a Stop-hook auto-flush cadence (cadence-checkpoint every-K-turns
checkpoint)?

## Pre-registered policies to evaluate

- **P0 PreCompact-only** (status quo + #746 once it flips) —
  harness decides when; rebuilder fills the post-compact context.
  Baseline.
- **P1 Every-K-turns Stop-hook** — periodic-checkpoint pattern. Counter
  monotonic; fire_idx-based; deterministic. Hypothesis: K=15 is a
  reasonable starting point (literature-typical value).
- **P2 Token-budget watermark** — flush at 50% / 80% of harness
  context-window watermark, expressed in transcript-byte-offset
  terms so it stays deterministic for replay.
- **P3 Turn-density-aware** — flush rate scales with substantive-
  turn rate (no flush on quiet windows; aggressive flush on dense
  bursts).

## Why this is parked, not actionable today

The campaign needs **a week or so of real fire-rate data** from
the maintainer's sessions with `trigger_mode = "threshold"`
enabled (set 2026-05-13 in `~/.aelfrice.toml` per the diagnostic
chain in #746). Without that data, R0 baseline can't be measured
against actual workload — only synthetic. Synthetic R0 risks
designing for the wrong distribution of session shapes.

The audit trail today shows **zero PreCompact fires** across 99
session-starts and 563 UPS events because `trigger_mode` defaulted
to `manual`. With the local config flip live, fire-rate data will
start accumulating. The campaign opens once data is in.

## Trigger criteria for opening the campaign

- [ ] At least 10 PreCompact fires logged in
[...]

PR #874, opened 2026-05-20
feat(hook): UPS-side cadence — inject rebuilder output via additionalContext (#870)

### LINKED ISSUE #870: feat: UPS-side cadence — inject rebuilder output via additionalContext (#749 P1 follow-up to #869)
## Summary

PR #869 shipped P1 every-K-turns cadence as a **Stop-hook** firing where the
rebuilder body is **discarded** (`src/aelfrice/cadence.py`, `_maybe_fire_cadence_checkpoint`
in `hook.py`). Stop has no `additionalContext` channel, so the cadence
checkpoint produces only side-effect bookkeeping: a `rebuild_log` row and
touch-state refresh. The open conversation window is untouched.

This issue proposes a sibling implementation on the **UserPromptSubmit
hook**, where `additionalContext` IS available, so a cadence-triggered
rebuilder pass can surface its synthesized output back into the next turn
— actually closing the loop #749's body framed.

## Why this is a separate issue, not a #869 follow-up commit

The #749 body asks: "does the harness fire PreCompact often enough to make
manual `/clear` unnecessary, or does the project need a **Stop-hook
auto-flush cadence**?" PR #869 answered the literal "Stop-hook" part by
shipping the deterministic firing predicate and the rebuilder-pass side
effects. But "auto-flush" — in the sense of actually moving rebuilder
output into the live conversation — needs the hook that has an injection
channel. That hook is UPS, not Stop.

PR #869 is internally consistent (the cadence module + Stop wiring + tests
ship a clean, falsifiable mechanism). It just doesn't satisfy the
operator-facing question on its own. A second consumer of the same
`cadence` module — this one — is what closes the gap.

## Proposed mechanism

Inside the UPS hook flow (`hook.user_prompt_submit` /
`hook._handle_user_prompt_submit`), after the existing per-prompt retrieval
runs but before the `additionalContext` envelope is finalized:

1. Resolve `[cadence]` via the same `resolve_cadence_enabled` /
   `resolve_cadence_policy` / `resolve_cadence_k` resolvers shipped by
   #869.
2. Read `next_fire_idx` from `session_ring.read_ring_state(session_id)`
   — same monotonic counter the Stop-hook P1 uses, so the two consumers
   stay perfectly synchronized.
3. Call `cadence.should_fire(fire_idx, cfg)`. On True:
   - Read recent N turns via the existing `_read_recent_for_pre_compact`
     helper (or its UPS equivalent if a transcript-reader already exists
     pre-UPS).
   - Call `_rebuild_and_format(recent, ...)` with the same
     `rebuilder_cfg` path PreCompact and Stop-hook cadence use.
   - **Inject the returned body** via `additionalContext` — either
     replacing the per-prompt retrieval payload at cadence turns, or
     concatenated alongside it (TBD — see open questions).
4. Emit the stderr observability line (mirror of #869's):
   `aelfrice: ups cadence checkpoint fired @ fire_idx=N (policy=..., k=...)`.

## Out of scope

- **Window truncation / `/clear`-equivalent.** UPS cannot truncate the
  harness window any more than Stop can. The harness exposes no API for
  it. This issue does **not** make `/clear` unnecessary in the literal
  sense; it makes the conversation that builds back up *after* a `/clear`
  (or that runs continuously without one) carry fresher rebuilder
  synthesis at K-turn boundaries.
- **P2 (token-budget watermark) and P3 (turn-density-aware).** Same
  forward-compatible pattern as #869: those land as new policy enum
  values without changing the UPS-side consumer's shape.
- **Cadence counter independent from Stop.** Both consumers share the
  one `session_ring.next_fire_idx` counter on purpose — a single
  cadence event-stream with two effect surfaces, not two independent
  cadences.

## Open questions for the implementer

- **Replace vs. augment.** At a cadence-turn UPS, does the rebuilder
  synthesis *replace* the per-prompt retrieval (treat it as a richer
  superset), or get *appended* (per-prompt retrieval + cadence
  synthesis = two `additionalContext` blocks)? Replacement is simpler;
  augment risks duplication and exceeds the token budget more often.
- **Determinism interaction with per-prompt retrieval.** UPS retrieval
  is already deterministic per #605. A cadence-turn UPS injection
  shares the same fire_idx with the next Stop's cadence firing (they
  both read `next_fire_idx`). Decide whether *both* fire at the same
  fire_idx (UPS first, Stop second on the same turn boundary) or only
  one — the cleanest is "both fire; Stop is bookkeeping-only as #869
  ships, UPS adds the injection." That's the natural reading; flag it
  in the implementation PR's design ratifications.
- **Default-OFF policy.** Inherits from #869's `[cadence] enabled =
  false` default. Operator opt-in. Same TOML knob enables both
  consumers — that's by design; one switch, two side effects.

## Trigger criteria

- [ ] PR #869 is merged (✓ — merge commit `9fa6fb7a1b68b1858d1c7457755cec25fe0a5722`,
      2026-05-19T20:39:22Z)
- [ ] Operator has accumulated ≥10 cadence fires under the Stop-hook P1
      consumer to confirm the firing predicate behaves as designed on
      real workload (i.e. log this gates work on the UPS-side consumer
      shipping with a known-good firing baseline).

## Determinism (#605)

Same predicate as #869: pure function of `(fire_idx, k)`, no wall-clock,
no random sampling. Replay reproduces the same UPS-cadence injections.

## Discretion

No `~/.claude/`-derived content. The rebuilder pass operates on the same
public surfaces PreCompact already drives.

## Test plan sketch

- Unit: extend `tests/test_cadence.py` only if a new helper is needed;
  the firing predicate is unchanged.
- Integration: new `tests/test_hook_ups_cadence.py` mirroring
  `tests/test_hook_stop_cadence.py`'s shape — multiples-of-K fire,
  cold-start suppression, missing/session-mismatched ring sentinel,
  empty recent-turns slice, env-override beats TOML, the
  `additionalContext` envelope carries the rebuilder body at cadence
  turns and only the per-prompt retrieval otherwise, env-default-OFF
  preserves pre-#749 UPS behaviour byte-for-byte.
- Regression: existing `tests/test_hook_user_prompt_submit.py` continues
  to pass unmodified — cadence is purely additive when enabled, no-op
  when disabled.

## Relation to #749

This is the second of the four pre-registered policies #749 mentions —
or rather, the **second consumer of P1**, since the same monotonic-
counter / fire_idx-keyed predicate drives both. P2/P3 follow the same
two-consumer pattern. #749 stays the umbrella tracker.

robotrocketscience:
Closes #870.

## Summary

Adds an in-session UPS-side cadence consumer that injects the rebuilder body via `additionalContext` at cadence boundaries — the surface #870 calls out as missing between #869 (P1 Stop-hook firing, body discarded) and #871 (resume cache, only read on next session start).

## Mechanism

- New `_maybe_run_ups_cadence_checkpoint(payload, session_id, serr)` in `src/aelfrice/hook.py` mirrors the Stop-side `_maybe_fire_cadence_checkpoint`, but **returns the formatted rebuilder body** instead of writing the resume cache. Default-OFF, fail-soft. Resolves `[cadence]` config via the same env > kwarg > TOML > default resolvers.
- Counter sharing: reads `session_ring.read_ring_state(session_id)["next_fire_idx"]` **before** this turn's `_ring_append_ids`, so UPS sees the same `fire_idx` Stop saw at end of the prior turn. The two consumers fire on the same monotonic-counter boundary by construction.
- Wiring in `user_prompt_submit`: after `_extract_session_id`, runs the dispatch; on fire, wraps the body in `<cadence-checkpoint>...</cadence-checkpoint>` (distinct tag from #871's `<cadence-resume>`) and writes it to stdout **ahead** of any retrieval body so the broader synthesis primes the model before query-specific BM25 hits.

## Open-question dispositions from the #870 body

- **Replace vs. augment** → **augment**. Both blocks emit when both apply. Replacement was rejected for the loss of query-specific signal; the token-budget concern is acceptable because cadence fires only at K boundaries (1× per K turns) — well below worst-case per-turn injection.
- **Counter sharing with Stop** → **shared**. UPS reads pre-append, Stop reads post-append-of-prior-turn → same value, same boundary. Stop continues to own the resume cache write; UPS does not write the cache, so the cache stays single-sourced. The cost is one duplicate rebuilder pass per K turns — acceptable.
- **Surface separation from `<cadence-resume>`** → distinct tag, distinct firing site. `<cadence-resume>` fires on first prompt of a new session (#871); `<cadence-checkpoint>` fires on in-session UPS K boundaries. They never collide because P1 `should_fire` requires `fire_idx > 0 AND fire_idx % k == 0` (which a fresh ring's `next_fire_idx=0` can't satisfy), and P2 needs accumulated transcript bytes which a fresh session lacks.

## Tests

- 11 unit tests (`tests/test_hook_ups_cadence.py`) target the dispatch in isolation: empty-session-id short-circuit, disabled-by-default, policy=off, P1 multiples-of-K cross-product, fire_idx=0 cold-start guard, missing ring state, empty recent-turns no-rebuild, missing DB, P2 boundary+threshold fire, P2 without-boundary no-fire, env override.
- 5 integration tests (`tests/test_hook_ups_cadence_wiring.py`) exercise `user_prompt_submit` end-to-end: disabled emits no block; P1 boundary emits `<cadence-checkpoint>` on stdout; P1 non-boundary emits no block; augment-mode ordering check; missing session_id skips.
- Full hook + cadence suite: 590 passed, 3 skipped.

## Determinism + discretion

- Determinism (#605): P1 predicate pure over `(fire_idx, k)`; P2 predicate pure over `(transcript_bytes, last_user_prompt, config)`. No wall-clock, no random sampling.
- Discretion (`ab96e9d3501b1c14`): full diff against `main` is clean against the pre-push deny-list.

## Default behaviour

Unchanged for anyone not opted into cadence. `[cadence] enabled` defaults to `false`; an unset section is byte-identical to a pre-#870 UPS hook.

## Out of scope (per issue body)

- Window truncation / `/clear`-equivalent — UPS cannot truncate the harness window any more than Stop can; this PR makes the conversation that builds up *after* a `/clear` (or runs continuously without one) carry fresher rebuilder synthesis at K boundaries.
- P3 (turn-density-aware) per #749 remains unimplemented; the policy enum is forward-compatible.

## Summary by Sourcery

Add UPS-side cadence support that injects rebuilder output into user prompt handling at cadence boundaries, aligned with existing Stop-side cadence policies and preserving default-off behaviour.

New Features:
- Introduce an in-session UPS cadence dispatcher that returns rebuilder synthesis for injection via additional context when cadence policies fire.
- Emit a `<cadence-checkpoint>` block on user prompt submission stdout at configured cadence boundaries, ahead of retrieval output, to prime the model mid-session.

Enhancements:
- Align UPS cadence firing with Stop-side cadence by sharing the session ring counter while keeping the resume cache single-sourced on the Stop hook.

Documentation:
- Document UPS-side cadence checkpoint injection and its behaviour in the v3 changelog.

Tests:
- Add unit tests for the UPS cadence dispatcher covering policy selection, ring state handling, firing predicates, and failure modes.
- Add integration tests for `user_prompt_submit` to verify checkpoint emission, ordering relative to retrieval, and behaviour when disabled or missing session IDs.

## Release Notes

* **New Features**
  * Added in-session cadence checkpoint injection with configurable policies (turn-based and transcript context-aware).
  * Configuration support via environment variables and TOML settings.
  * Checkpoints emit to stdout before other session output for improved observability.

* **Documentation**
  * Updated changelog documenting the new cadence checkpoint feature, configuration options, and behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/874?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


robotrocketscience:
[claim:review:garsecg:2026-05-20T22:26:36Z]

robotrocketscience (commented):
LGTM. Approving.

### Verified

- **CI** — all required checks green (pytest 3.12 + 3.13, bench-smoke, secrets-scan, pattern-scan x3, history-scan, codeql, deptry, migration-policy-check, pr-title/body/commit-msg prefix).
- **Signatures** — all 7 commits signed (`G`).
- **Counter sharing claim** — `_maybe_run_ups_cadence_checkpoint` reads `next_fire_idx` via `read_ring_state(session_id)` at `hook.py:852`, before this turn's `_ring_append_ids` runs at `hook.py:1015`. UPS therefore sees the same `fire_idx` Stop saw at end of the prior turn — boundary parity by construction, as the body claims.
- **Tag separation from #872's `<cadence-resume>`** — checks out. `<cadence-resume>` only fires on `is_session_first_prompt(session_id)` (gate from #578) with a fresh ring (`next_fire_idx=0`), so P1's `fire_idx > 0 AND fire_idx % k == 0` predicate cannot also be true on that same prompt; P2 requires accumulated transcript bytes a fresh session lacks. No collision path.
- **Augment-mode ordering** — `cadence_checkpoint_block` is written directly to `sout` at `hook.py:862`. The retrieval `body` (which carries #872's `<cadence-resume>` via `_format_hits_with_session_start` at `hook.py:964 / 1052`) is written later in the same function. Test `test_checkpoint_block_precedes_retrieval_body` exercises the ordering against a seeded hit.
- **Determinism (#605)** — P1 predicate pure over `(fire_idx, k)`; P2 predicate pure over `(transcript_bytes, last_user_prompt, config)`. No wall-clock, no random sampling.
- **Discretion** — diff vs `main` clean against the pre-push deny-list. (One hit on `Claude API format` is in the existing #872 changelog text already on main, and it names a public Anthropic API surface, not a session-internal reference.)
- **Bot feedback** — both already self-applied:
  - CodeRabbit's `[#871](pull/872)` label/URL mismatch fix in `360eabe1`.
  - Silent-except critique addressed in `84dda57d` (`traceback.print_exc(file=serr)`, fail-soft preserved).
- **Tests** — `tests/test_hook_ups_cadence.py` + `tests/test_hook_ups_cadence_wiring.py` + Stop-side P1/P2 + UPS resume regressions: 51 passed locally on PR HEAD `84dda57d`.

### Minor — not a blocker

Sourcery's DRY observation has merit — `_maybe_fire_cadence_checkpoint` (Stop) and `_maybe_run_ups_cadence_checkpoint` (UPS) share the policy-dispatch / config-resolve / ring-read scaffolding; they diverge only in (a) Stop side-effects (cache write + `rebuild_log` discard) vs UPS pure return, and (b) UPS's empty-`session_id` short-circuit. A future refactor extracting a `_resolve_cadence_dispatch_inputs(payload, session_id, cwd) -> tuple[CadenceConfig, fire_idx_or_None]` helper would compress both sites; defensible to leave as-is given the divergent terminals.

### Net

Surface split is the right call for the constraint (Stop has no `additionalContext` channel → cache-only; UPS has stdout → can inject). Duplicate rebuild pass per K turns is the explicit cost — accepted in the PR body. Defaults stay OFF, contract is fail-soft + observable, and the test surface is thorough across both policies.

Adding `ready-to-merge`.

robotrocketscience:
[release:review:garsecg:2026-05-20T22:30:12Z]

Issue #899, opened 2026-05-21
bench(mab/structmemeval): back-fill TBD re-measurements per v3.0.1 _calibration_notes


robotrocketscience:
## Background

`benchmarks/results/v3.0.1.json` `_calibration_notes` declares four MAB splits and four StructMemEval sub-tasks as TBD re-measurements:

> "the v2.0.0 canonical four MAB splits (Conflict_Resolution, Test_Time_Learning, Long_Range_Understanding, Accurate_Retrieval) and four StructMemEval sub-tasks (location, accounting, recommendations, tree) were NOT re-measured at v3.0.1 and are TBD."

This tracker exists so the back-fill work has a paper-trail target. Each back-fill is a single PR adding a `benchmarks/results/v3.0.1-<surface>-backfill.json` capture.

## Status

| Surface | Status | PR / capture |
|---|---|---|
| MAB Long_Range_Understanding | done | PR follows; `v3.0.1-mab-lru-backfill.json` |
| MAB Conflict_Resolution | TBD | — |
| MAB Test_Time_Learning | TBD | — |
| MAB Accurate_Retrieval | TBD | — |
| StructMemEval per-sub-task (4) | TBD | (existing 14-row aggregate at SH=100% has the variance-probe asterisk in v3.0.1.json; per-sub-task break-down would refine that) |

## Methodology (established by the LRU PR)

Three-step pipeline for each surface:

1. `uv run python benchmarks/mab_adapter.py --split <Split> --retrieve-only /tmp/<surface>/retrieval.json` — produces 100-2000 `(question, context)` pairs + companion `_gt.json` (ISOLATION-guaranteed: no answers in retrieval file).

2. Reader pass: in-process LLM helpers dispatched in parallel via the `Agent` tool, each handling a chunk of the retrieval JSON. No `anthropic` SDK / API-key path (locked rule: bench LLM via in-process helpers). Each helper produces `predictions_<N>.json` with `{row_idx, id, prediction}` items.

3. Score via `mab_adapter.score_multi_answer` joined on `(row_idx, id)` composite key.

Single-run capture; variance band TBD on a future multi-run pass.

## Acceptance

- [x] MAB Long_Range_Understanding back-fill landed.
- [ ] MAB Conflict_Resolution back-fill (800q, the largest non-factconsolidation cut).
- [ ] MAB Test_Time_Learning back-fill (700q).
- [ ] MAB Accurate_Retrieval back-fill (2000q — heaviest reader load).
[...]

Issue #891, opened 2026-05-21
bench(locomo): v3.0.1 overall F1 regression vs agentmemory writeup (66.1%→40.88%, Adversarial cliff)


robotrocketscience:
## Summary

v3.0.1 pinned bench results (2026-05-13, commit `0e91fd1d`) show LoCoMo scoring 25 percentage points below the published agentmemory writeup:

| Cut | agentmemory writeup | aelfrice v3.0.1 | Delta |
|---|---|---|---|
| **LoCoMo overall F1** | **66.1%** | **40.88%** | **−25.2 pp** |

Per-category breakdown is more diagnostic:

| Category | writeup F1 | aelfrice v3.0.1 F1 | Delta |
|---|---|---|---|
| Single-hop | 69.4% | 38.5% | −30.9 |
| Temporal | 45.4% | 30.6% | −14.8 |
| Multi-hop | 42.2% | 24.8% | −17.4 |
| Open-ended | 30.5% | 48.8% | **+18.3** |
| Adversarial | **97.5%** | 38.3% | **−59.2** |

## Where the regression sits

**Adversarial-category cliff (−59 pp)** is the most diagnosable signal. The writeup says aelfrice's predecessor scored 97.5% on Adversarial. v3.0.1 scores 38.3%. That's not a tuning regression; it's a behavioral change — something the predecessor did to recognize and reject adversarial queries is no longer happening.

The other categories are also down:
- Single-hop −31 pp suggests basic retrieval-quality regression
- Temporal −15 pp suggests temporal_sort isn't doing what it did, OR the narrative-timestamp ingestion is missing
- Multi-hop −17 pp tracks the MAB MH regression (separate issue)

**Open-ended +18 pp** is the one bright spot. v3 ships better generative-answer formatting. Doesn't compensate.

## Protocol differences to verify

Writeup spec for LoCoMo:
- Dataset: `locomo10.json`, 10 conversations, 5,882 turns, 1,986 QA pairs, 5 categories
- Ingestion: all 10 through standard onboarding, session boundaries preserved
- Retrieval: FTS5 + HRR + BFS, 2,000-token budget, batch size 1
- Reader: Claude Opus 4.6
- Prompts: exact LoCoMo protocol prompts (categories 1/3/4 standard, category 2 appends DATE, category 5 forced-choice)

[...]

PR #908, opened 2026-05-22
exp(bench): MAB Test_Time_Learning back-fill — n=700, not measurable

robotrocketscience:
## What

Back-fill measurement of the **MAB Test_Time_Learning** surface (700q) from the v3.0.1 `_calibration_notes` TBD list, following the LRU-established retrieve-then-reader methodology. Adds one static capture: `benchmarks/results/v3.0.1-mab-ttl-backfill.json`.

## Result: 0.0% — a *measurement-validity* zero, not a model-quality zero

Full pipeline ran clean (700 non-empty retrieval contexts, 7 parallel in-process reader helpers, `score_multi_answer` joined on `(row_idx, id)`), and every one of the 700 items scored 0 on EM/SEM/F1. The capture's `_status` is `not_measurable` (not `ok`), so `badge.py` does not count it as a passing bench.

There are **two distinct failure modes** across the 6 sources (verified empirically):

### ICL sources (banking77, clinic150, NLU, TREC-coarse, TREC-fine)
The demonstration labels are dropped **at ingestion, not retrieval**. The raw context is `(utterance, label)` pairs (`utterance label: N`); `mab_adapter.ingest_context` chunks it and `ingest_turn`'s belief extraction atomizes each chunk into per-utterance beliefs, discarding the `label: N` annotation. **Verified: ingesting one banking77 row produced 4293 beliefs, of which 0 retain any `label:` token.** `retrieve_v2` can only return label-free utterances because the labels were never stored.

The opaque integer GT is **not** an independent blocker here: if the `(utterance, label)` demonstrations survived ingestion and co-retrieved, the reader would do in-context learning — see `similar-utterance label: 28` and output `28` = GT. Integer GT is just why the missing-label failure yields exactly 0 with no partial credit.

### recsys_redial (distinct cause)
GT is a ReDial movie-DB id (`7008`); the context references movies by title+year (`From Dusk till Dawn (1996)`), never by id. Even a correct recommendation emits a title that can't match the id without a title→id decode. This is a scoring/decode gap, independent of ingestion.

## Relationship to #891

Distinct finding. #891 was "the dispatcher runs no reader at all." Here the reader pipeline runs correctly end-to-end — the ICL failure is at ingestion (labels not persisted), the recsys failure is at scoring (GT id absent from context). The reader pipeline that rescued LRU does **not** rescue TTL. For contrast, the Conflict_Resolution back-fill (factconsolidation, textual answers) scored SEM 37.1% on the same pipeline.

## Recommendation (captured in `_not_measurable`)

The two modes need **different** fixes:
- **ICL:** ingest each demonstration as an atomic `(utterance, label)` unit (or attach the label as belief metadata) so it co-retrieves. No GT-format change needed.
- **recsys:** decode the predicted title to a ReDial movie id before scoring.

Until both land, TTL sources should be **excluded from the string-match headline** rather than reported as a 0% model score. Flagging for the operator to decide whether to reframe the #899 TTL acceptance checkbox.

## Verification

- 42-item stratified probe → 0.0%, then full 700-item run reproduced 0.0% on all 6 sources (0 items with any nonzero metric).
- Ingest mechanism verified directly: 0 of 4293 beliefs from a banking77 row retained any label token.
- Measured at `3af82bc3`; the later CR back-fill `64d56bd6` is result-only and did not change `retrieve_v2`.
- `tests/test_benchmarks_badge.py` + `tests/test_bench_tolerance_skip.py` green.

Refs #899.

### DIFF
--- a/benchmarks/results/v3.0.1-mab-ttl-backfill.json
+++ b/benchmarks/results/v3.0.1-mab-ttl-backfill.json
@@ -1,0 +1,63 @@
+{
+  "_calibration_notes": "Back-fill measurement of MAB Test_Time_Learning (700q) — one of the TBDs noted in v3.0.1.json _calibration_notes. Measured 2026-05-22 against main HEAD 3af82bc3 (post-PR #887/#888/#891; the subsequent CR back-fill at 64d56bd6 is result-only and did not change retrieve_v2 ranking behaviour). Methodology: full retrieve-then-reader pipeline, identical to the LRU back-fill. Step 1 — `uv run --extra benchmarks python benchmarks/mab_adapter.py --split Test_Time_Learning --retrieve-only` produced 700 (question, context) pairs across 6 source rows; budget=2000 per retrieve call, retrieve_v2 via the v1.0.x lab-compat shim with use_bfs=True. All 700 items had non-empty retrieval context (median 2348 chars, max 6121). Step 2 — seven in-process LLM helpers dispatched in parallel (one per 100-item chunk), each writing predictions_<N>.json. Locked rule honored: bench LLM calls go through in-process helpers, never via a direct SDK / API-key path. Step 3 — score via mab_adapter.score_multi_answer (exact_match + substring_exact_match + token-level F1) joined on (row_idx, id). HEADLINE: 0.0% on every metric and every source (700/700 items scored 0; zero items had any nonzero metric). THIS IS A MEASUREMENT-VALIDITY ZERO, NOT A MODEL-QUALITY ZERO — see _not_measurable below. n=1 single run; no variance probe was taken because the score is structurally pinned at 0 (re-running cannot move it). Bench fixtures downloaded fresh. Not pinned by checksum.",
+  "_not_measurable": {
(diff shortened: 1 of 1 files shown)


yoshi280:
[claim:review:Kulili:2026-05-22T19:00:59Z]

yoshi280:
Review: LGTM, adding ready-to-merge.

Strong analysis. Independent verification on my end: my closed PR #910 ran the same pipeline and saw the same 0.0% across all 6 sources, with multiple reader subagents independently flagging the missing label / id metadata. My PR attributed the cause to retrieval, but your empirical check (`0/4293 beliefs from a banking77 row retain a 'label:' token`) proves it's actually at ingest — `ingest_turn`'s belief extraction atomizes the chunks and drops the `label: N` annotation. `retrieve_v2` is faithfully returning what was stored. Right call to correct that framing.

The two-failure-mode split is also load-bearing — collapsing ICL and recsys into one finding (as my PR did) hides that they need different fixes:
- ICL: atomic (utterance, label) ingest unit so labels co-retrieve
- recsys: title→id decode at scoring time, or score in title space

The `_status: not_measurable` + `items_with_any_nonzero_metric: 0` shape is also a better contract than the `_status: ok` shape I had — badge.py won't count this as a passing bench, which matches the actual semantics.

CI green (26 SUCCESS, 4 SKIPPED); JSON parses; discretion clean against the diff.

yoshi280:
[release:review:Kulili:2026-05-22T19:01:47Z]

PR #911, opened 2026-05-22
fix(hook): conversation-aware UserPromptSubmit retrieval

### LINKED ISSUE #909: hook: UserPromptSubmit ranking surfaces irrelevant beliefs while load-bearing beliefs in the same index are skipped
## The broken contract

The whole point of the auto-memory system is that the user can stop repeating themselves across sessions. The `UserPromptSubmit` hook is the only path by which prior-session content reaches the agent without an explicit `aelf search`. When the hook surfaces low-relevance beliefs while load-bearing ones sit unused in the same index, the contract is broken end-to-end: the agent has working amnesia about content the user knows is stored, and the user has to do retrieval by hand.

That's the failure mode in this report. It is not a duplicate of #218 (dedup quality) or #855 (coverage / scope-out). The surfaced beliefs in this session were unique and the index had the right content. The ranker just picked the wrong slice.

## What happened

Long-running session. Topic the user had developed with the agent across roughly four prior sessions, stored as ~30 unlocked beliefs plus a small number of locked beliefs about an adjacent topic. User asked a follow-up question that any reasonable retrieval system should have routed to that thread.

Hook fired and surfaced:

- The locked beliefs from the adjacent topic. Correct, as designed.
- ~10–15 unlocked beliefs from **completely unrelated** topics: stale presentation-slide notes, off-topic technical jargon from other projects, ambient chatter. Zero overlap with the user's actual question.

The load-bearing unlocked thread the user was referring to — the one containing the specific numeric recommendations the user expected the agent to recall — was **not surfaced**.

User pushed back: \"we already went through this 4 times why arent you remembering.\" Ran `aelf search` with a three-keyword query any user would phrase. The load-bearing thread came back in the top 10 hits immediately, including multiple beliefs containing the exact recommendation the user expected the agent to already have.

The hook's own coverage footer (shipped via #855) reported \"retrieved 13 of 59 matching beliefs.\" The 13 surfaced were noise. The 46 not surfaced contained the answer.

## What I expected

The hook should rank unlocked beliefs such that the top-K it injects is at least overlapping with what `aelf search <reasonable-query>` would return for the same prompt. Today the two retrievals can be near-disjoint, and the agent has no way to know it's working off a bad slice. That's the asymmetry that turns the system into a footgun: the user trusts the agent, the agent trusts the hook, the hook silently mis-ranks, the user concludes the agent isn't listening.

## Hypothesis

BM25 over raw prompt tokens fails when the user's vocabulary in the prompt diverges from the vocabulary in the load-bearing beliefs (paraphrase, numeric references, synonym drift, pronouns). When a user runs `aelf search` they pick keywords that match storage; the hook can't.

Candidate directions, not prescriptive:

1. **Query rewriting at hook fire.** Extract entities/topics from the prompt, expand to synonyms, BM25 the rewritten query.
2. **Conversation-aware query construction.** Condition on the last N turns, not just the latest prompt — the topic continuity is in the dialog history, not always in the literal current message.
3. **Scope-biased retrieval.** When a `scope`/`project` field is populated (per #855 follow-on), weight beliefs in the same scope above out-of-scope ones, even at the cost of raw BM25 score.
4. **Link-graph expansion.** After top-K BM25, expand to 1-hop neighbors via `[[name]]` links so directly-cited related content gets a chance.

None of these are *the* fix. Flagging them so the issue is actionable.

## Severity

High. This is not a quality nit. It is the agent appearing to ignore stored user context, which is the worst possible failure mode for a memory product. #855 made the asymmetry visible with the coverage footer; this is the next step — making the top-K worth reading.

## Minimal reproduction (sketch)

```
AELFRICE_DB=/tmp/scratch.db
aelf:lock \"Topic A: specific recommendation X with numeric detail Y\"
# add 20 unlocked beliefs on Topic A using related-but-not-identical vocabulary
# add 80 unlocked beliefs on unrelated Topics B, C, D
# trigger UserPromptSubmit with a prompt about Topic A using a *paraphrase* of stored vocabulary
# observe: surfaced unlocked beliefs are mostly B/C/D
# aelf search \"Topic A\" returns the load-bearing belief in the top 5
```

I have not constructed the scratch repro yet — the live observation is the artifact. Happy to build one if useful.

## aelfrice version
3.3.0

## Python version
3.12+ (exact pending)

## Operating system
macOS, Darwin 25.3.0

## Which surface
- MCP server (Claude Code) — UserPromptSubmit hook

## Related (not duplicates)

- #218 (closed, Apr) — hook dedup/quality. Not this. Surfaced beliefs here were unique.
- #855 (closed, May 19) — under-sampling + `aelf:scope-out`. Made this bug observable from the user side via the coverage footer. Ranking quality is the next step.

robotrocketscience:
## What

Fixes #909 — the UserPromptSubmit hook surfacing irrelevant beliefs while the load-bearing thread sits unsurfaced in the same index.

## Root cause (reproduced + pinned)

The live per-prompt path (`user_prompt_submit` → `_retrieve` → `retrieve`) BM25s the **literal prompt only**. When the topic vocabulary lives in the dialog history — paraphrase, pronoun, or numeric-reference follow-ups — and not in the current prompt, the load-bearing thread scores ~0 lexically and ranks near the bottom, while noise that shares the prompt's generic words ("numbers", "settings", "decided") out-ranks it. Under the hook's token-budget trim that buried thread falls below the cut — the reported "13 of 59" symptom.

I built a deterministic repro (the issue author left it unbuilt). Result, on a corpus of pure-jargon load-bearing beliefs + noise sharing the prompt's generic words:

| query | load-bearing rank |
|---|---|
| prompt-only (current behavior) | **46 of 47** (cut by budget) |
| + `stack-r1-r3` `transform_query` | **still buried** — *not* the fix |
| + recent-turn context | **rank 0** |

`transform_query` is confirmed **not** the fix: the domain vocab isn't in the prompt to expand. The query rewrite is also only wired into the rebuild path, not the per-prompt path — but wiring it in wouldn't help here anyway.

## Fix

Fold a **small window of recent dialog turns** into the BM25 query, with the current prompt repeated (`prompt_weight`) so its terms stay dominant. The topic continuity that the prompt lacks lives in the conversation history. Pure, deterministic, **no embeddings** (honors the determinism lock). The recent-turn reader (`_read_recent_for_pre_compact`) already existed for the rebuild path; this wires it into the per-prompt path.

**Design note — deliberately *not* the rebuilder's `turn_window_n` (default 50).** I verified empirically that a large / topic-drifted window *re-buries* the thread (20 off-topic turns → load-bearing absent again). The fix uses a small dedicated window (default 4) plus prompt-weighting as a precision guard.

- Default-on, opt-out via `[user_prompt_submit_hook] conversation_aware_query_enabled`.
- `conversation_aware_turn_window` (default 4) and `conversation_aware_prompt_weight` (default 3) tunable.
- Fail-soft: any error reading turns falls back to the prompt-only query. The prompt-shape gate and all telemetry/audit still key on the raw prompt.

## Tests

`tests/test_hook_conversation_aware_909.py` (14 tests):
- Builder unit tests (weighting, window slice, fail-soft clamps).
- Config parsing of the three new keys (valid + bad-type fallback).
- **Retrieval-seam regression** — prompt-only ranks the thread last of 47; conversation-aware lifts it to rank 0 (the exact #909 mechanism).
- **Precision guard** — a good-overlap prompt stays top-ranked after augmentation (no dilution).
- End-to-end wiring through `user_prompt_submit` (ON surfaces the thread ahead of the noise it was buried under; flag-off reverts to prompt-only ranking).

All 120 hook-suite + new tests green.

## Commits

1. `fix(hook):` — config + query builder + wiring.
2. `test(hook):` — regression + guard.
3. `docs(changelog):` — CHANGELOG/v3.md + docs/user/CONFIG.md.

Closes #909.

## Summary by CodeRabbit

* **New Features**
  * UserPromptSubmit retrieval now supports conversation-aware queries, enhancing search results by incorporating context from recent dialog turns while maintaining prompt dominance. Enabled by default with tunable window and weighting parameters.

* **Documentation**
  * Added configuration reference for new conversation-aware retrieval options, including enable/disable toggle and tuning parameters.

[Review Change Stack](https://app.coderabbit.ai/change-stack/robotrocketscience/aelfrice/pull/911?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/CHANGELOG/v3.md
+++ b/CHANGELOG/v3.md
@@ -8,6 +8,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **UserPromptSubmit retrieval is now conversation-aware (#909).** The
+  per-prompt hook previously BM25'd the literal prompt only, so
+  paraphrase / pronoun / numeric-reference follow-ups (where the topic
+  vocabulary lives in the dialog history, not the current prompt) ranked
+  the load-bearing thread near the bottom — below the token-budget cut,
+  never surfaced — while noise sharing the prompt's generic words won.
+  The hook now folds a small window of recent turns into the query, with
+  the current prompt weighted to stay dominant. Default-on, opt-out via
+  `[user_prompt_submit_hook] conversation_aware_query_enabled`; window
+  and weight are tunable. Deterministic, no embeddings.
+
 ## [3.3.0] - 2026-05-21
 
 ### Added
--- a/docs/user/CONFIG.md
+++ b/docs/user/CONFIG.md
@@ -12,7 +12,7 @@ A single optional TOML file at the root of a project (or any ancestor). It expos
 - `[retrieval]` (v1.3+) — retrieval-time tier toggles + ranking. Knobs: `entity_index_enabled` (L2.5), `bfs_enabled` (L3), `posterior_weight` (partial Bayesian-weighted L1 ranking), `use_bm25f_anchors` (BM25F-with-anchor-text since v1.7), `use_heat_kernel` (authority scoring lane, default-on since v2.1), `use_hrr_structural` (HRR structural-query lane, default-on since v2.1), `hrr_persist` (HRR structural-index on-disk persistence, default-on since v3.0), `use_type_aware_compression` (per-belief retention-class compression, default-on since #769), `use_intentional_clustering` (co-locating related beliefs, default-on since v3.0). Two placeholder flags (`use_signed_laplacian`, `use_posterior_ranking`) are recognised but emit a deprecation warning if set — their lanes have not yet shipped.
 - `[rebuilder]` (v1.7+) — context-rebuilder knobs. Selects the query-understanding stack (`query_strategy`) and sets token-budget floors for the session-scoped and L1 belief lanes (`[rebuild_floor] session` and `[rebuild_floor] l1`).
 - `[feedback]` (v3.0+) — feedback-lane opt-ins. `sentiment_from_prose` (default `false`) wires the sentiment-feedback detector into `UserPromptSubmit` (#606).
-- `[user_prompt_submit_hook]` (v3.0+) — UPS hook knobs. `prompt_shape_gate_enabled` (default `true`) gates trivial-prompt and system-envelope short-circuits before BM25 retrieval runs (#674).
+- `[user_prompt_submit_hook]` (v3.0+) — UPS hook knobs. `prompt_shape_gate_enabled` (default `true`) gates trivial-prompt and system-envelope short-circuits before BM25 retrieval runs (#674). `conversation_aware_query_enabled` (default `true`, v3.x #909) folds a small window of recent dialog turns into the BM25 query so paraphrase / pronoun / numeric-reference follow-ups still surface the load-bearing thread; tuned by `conversation_aware_turn_window` (default `4`) and `conversation_aware_prompt_weight` (default `3`).
 
 Locks, hooks, MCP tools, and the Bayesian feedback math are not affected.
 
@@ -166,6 +166,22 @@ sentiment_from_prose = false
 # "system-tag:<task-notification>", etc.). The session-start
 # sub-block is preserved unaffected. Set to false to disable.
 prompt_shape_gate_enabled = true
+# v3.x (#909). Default `true`. Conditions the per-prompt BM25 query on
+# a small window of recent dialog turns, so paraphrase / pronoun /
+# numeric-reference follow-ups still surface the load-bearing thread
+# (the topic vocabulary the prompt lacks lives in the conversation
+# history). The current prompt is repeated `conversation_aware_prompt_weight`
+# times so its terms stay dominant; the last `conversation_aware_turn_window`
+# turns are appended once. Fail-soft: any error reading turns falls back
+# to the prompt-only query. Set enabled to false for v3.2-and-earlier
+# prompt-only behaviour.
+conversation_aware_query_enabled = true
+# Number of trailing turns folded into the query (default 4). Kept small
+# on purpose: a large window re-buries the thread on topic-drift.
+conversation_aware_turn_window = 4
+# Prompt repeat count for BM25 term-frequency weighting (default 3,
+# minimum 1). Higher = the current prompt dominates the appended turns.
+conversation_aware_prompt_weight = 3
 
 [onboard.llm]
 # v1.3.0+; default flipped to true in v1.5.0 (#238). Host-driven
--- a/src/aelfrice/hook.py
+++ b/src/aelfrice/hook.py
@@ -963,7 +963,11 @@ def user_prompt_submit(
             if token_budget is not None
             else DEFAULT_HOOK_TOKEN_BUDGET
         )
-        config = load_user_prompt_submit_config(stderr=serr)
+        # #909/#887: resolve config from the payload's cwd, not the hook
+        # process's incidental cwd — same project-relative reasoning as the
+        # <recent-work> builder above. Falls back to process cwd when the
+        # payload carries no cwd (start=None → Path.cwd()).
+        config = load_user_prompt_submit_config(start=payload_cwd, stderr=serr)
         # #606: sentiment-feedback lane — apply correction signals from
         # this prompt to the prior UPS turn's retrieved beliefs BEFORE
         # this turn's retrieval, so demoted posteriors are reflected in
--- a/src/aelfrice/hook.py
+++ b/src/aelfrice/hook.py
@@ -176,6 +176,24 @@ _UPS_SECTION: Final[str] = "user_prompt_submit_hook"
 _COLLAPSE_KEY: Final[str] = "collapse_duplicate_hashes"
 _PROMPT_SHAPE_GATE_KEY: Final[str] = "prompt_shape_gate_enabled"
 _CONFIG_FILENAME: Final[str] = ".aelfrice.toml"
+# #909: conversation-aware retrieval. The live per-prompt UPS retrieval
+# BM25s the literal prompt only; when the topic vocabulary lives in the
+# dialog history (paraphrase / pronoun / numeric reference) and not in
+# the current prompt, the load-bearing thread scores ~0 lexically and is
+# never surfaced. Folding a SMALL window of recent turns into the query
+# restores it. Deliberately NOT the rebuilder's `turn_window_n` (default
+# 50): a large window re-buries the thread on topic-drift (empirically
+# verified). Small window + prompt-weighting keeps the current prompt
+# dominant and avoids dragging in stale topics.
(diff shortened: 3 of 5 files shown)


robotrocketscience:
[claim:review:Kulili:2026-05-22T19:35:42Z]

robotrocketscience:
Review: LGTM, adding ready-to-merge.

Clean fix and well-motivated. Verifying the work:

**Root-cause repro is the right move.** Building the deterministic harness (46/47 → rank 0) and empirically ruling out `transform_query` (the obvious wrong answer — query rewrite can't expand vocab the prompt doesn't have, and the rewrite path isn't even wired to per-prompt retrieval) is what makes the diagnosis trustworthy. The issue author left this unbuilt; you closed the gap.

**Mechanism is consistent with #605.** This is a hook fix, not a retrieval substrate change — the BM25 ranker stays byte-identical; only the query input changes. Deterministic, no embeddings. Locked properties hold.

**Defaults are right.** Default-on with TOML opt-out (`conversation_aware_query_enabled = false` for legacy v3.2-and-earlier behaviour). The small window (4) + prompt repeat (3) is a sensible starting point — too-large window would re-bury on topic-drift, which the body notes.

**Belt and suspenders that I checked:**
- bool-as-int trap closed (explicit `isinstance(x, bool)` before `isinstance(x, int)` for both window and weight)
- Telemetry/audit still key on raw prompt, not augmented query (so logs stay legible)
- Fail-soft via try/except → falls back to prompt-only on any error
- `prompt_weight < 1` clamped to 1 (prompt always appears)
- New helper is pure and tested independently from the wiring

**CI green** (27 SUCCESS / 7 SKIPPED / 2 CANCELLED from earlier runs). Mergeable. Discretion grep clean against the diff.

robotrocketscience:
[release:review:Kulili:2026-05-22T19:36:48Z]

robotrocketscience:
[claim:review:Kulili:2026-05-22T19:51:39Z]

robotrocketscience:
Reviewed (Kulili).

**Gates**
- 3 atomic signed commits (`37a0f134` fix, `d3c93b64` test, `754dcfd5` docs), FF on `main`.
- Discretion grep on diff: clean.
- All required checks SUCCESS / SKIPPED (size-check cancellations are the duplicate-job artifact, not a failure).
- Bot reviews: sourcery rate-limited, github-advanced-security empty, CodeRabbit posted 2 actionable nits (see below).

**Mechanism vs the locked design (memory `ups_retrieval_conversation_aware`)**
1. ✓ Per-prompt UPS path is prompt-only — fixed by folding recent turns into the BM25 query at `hook.py:_build_conversation_aware_query`.
2. ✓ `transform_query` is explicitly not the fix and the PR body documents this empirically (matches the locked belief).
3. ✓ Window is **not** the rebuilder's `turn_window_n=50`: dedicated `_CONV_AWARE_WINDOW_KEY` default 4, with the PR body's empirical note that a large window re-buries the thread on topic drift.
4. ✓ Prompt-weighting (default 3) keeps the current prompt's BM25 term-frequency dominant — `parts = [prompt] * weight` then append turn text. The legacy-equivalence corner (`prompt_weight=1`, empty `recent_turns` → byte-identical to raw prompt) is correct.
5. ✓ Fail-soft: any read/decode error on turns falls back to prompt-only with traceback to stderr; prompt-shape gate + audit/telemetry still key on the raw prompt, not the augmented query.
6. ✓ Determinism lock honored — no embeddings, pure string composition.
7. ✓ Default-on, opt-out via `[user_prompt_submit_hook] conversation_aware_query_enabled` (matches the project pattern of "fix lives on, escape hatch exists").
8. ✓ `bool`-rejection in the int validators is the right call (Python's `bool ⊂ int` would otherwise let a stray `true` become `window=1`).

**Test coverage**
The regression test is the exact #909 mechanism: load-bearing espresso jargon vs noise that shares the prompt's generic recall words ("numbers", "settings", "decided"). Prompt-only buries the thread (locked at 46/47); conversation-aware with a 1-turn domain context lifts it to rank 0. Precision guard inverts the setup so a high-overlap prompt isn't degraded by augmentation. End-to-end wiring through `user_prompt_submit` is exercised both flag-on and flag-off.

**CodeRabbit nits — non-blocking**
1. *"Add `MAX_CONV_AWARE_WEIGHT` clamp on `_CONV_AWARE_WEIGHT_KEY`."* — The value is sourced from the user's own `.aelfrice.toml`; it's not an attack surface. `[prompt] * 1000` is a config-file footgun the user explicitly opted into. Worth adding as a follow-up nicety, not a blocker.
2. *"Test should write `.aelfrice.toml` into the payload cwd, not via `monkeypatch.chdir(tmp_path)`."* — Test-hygiene preference; the production code path resolves config via cwd walk-up, so `chdir + write` exercises the real path. Non-blocking.

LGTM — adding `ready-to-merge`.

robotrocketscience:
[release:review:Kulili:2026-05-22T19:53:19Z]

robotrocketscience:
Merge-train blocked on two items now (author-side cleanup):

1. **Two unresolved CodeRabbit threads** on `src/aelfrice/hook.py` and `tests/test_hook_conversation_aware_909.py`. I considered both non-blocking in my review (see above) — author/sister can `Resolve conversation` as wontfix to clear the block, or push minor follow-ups.
2. **Stale base** — `#913` merged in front, so this branch is no longer FF on `main`. Needs `git rebase github/main` + force-push.

After both: re-add `ready-to-merge`. (I'm not the author, so I'm not touching the branch; flagging for the next session that picks this up.)

robotrocketscience:
[claim:review:garsecg:2026-05-22T20:02:51Z]

robotrocketscience:
Addressed the three review threads and rebased onto current `main`:

- **`hook.py` — cap `conversation_aware_prompt_weight`** (55da7da6): added `MAX_CONV_AWARE_WEIGHT = 8`; values above the ceiling fall back to the default, mirroring the existing `< 1` floor handling. Prevents `[prompt] * weight` from ballooning the FTS query on the hot path. Unit test added.
- **`hook.py` — config resolution** (7552b484): the handler now loads `[user_prompt_submit_hook]` config from the payload cwd (`start=payload_cwd`), consistent with the `<recent-work>` builder, instead of the process's incidental cwd. Falls back to process cwd when the payload carries none. The flag-off e2e test drops its `chdir` so it genuinely exercises payload-cwd resolution.
- **test — implicit string concat** (dce96e0e): collapsed the two adjacent literals into one string. The concatenation was intentional (a single corpus belief); a comma would have wrongly split it and broken the `_TARGET` assertion.
- **test follow-on** (8eb4d370): `_payload` gained an optional `cwd`; the gate-disable test in `test_hook_user_prompt_submit.py` now passes config via the payload cwd rather than `chdir` + process cwd.

Full suite green locally (4833 passed, 68 skipped, 75 xfailed).

robotrocketscience:
[release:review:garsecg:2026-05-22T20:17:09Z]

PR #914, opened 2026-05-22
fix(bench): task-aware StructMemEval prediction scorers

robotrocketscience:
## What

Adds **task-aware prediction scorers** for StructMemEval. The shipped `check_state_correctness` is a single word-overlap heuristic over the whole reference and is invalid for scoring answers, in a different way per task. Part of #899.

## Why the current scorer is invalid

`check_state_correctness(text, reference)` builds a significant-word set from the **entire** reference and returns `(matches / |ref_words|) >= 0.5`. Per task:

- **location** — the reference embeds an exclusion note: *"…should NOT mention activities from your previous cities (Lisbon …, Tokyo …, London …, Barcelona …)."* Those distractor tokens enter the word-set, so text that contains **every stale city** matches them and scores high, while a concise **correct** answer that omits them falls below 0.5 and scores 0. Measured: raw retrieval (which carries all stale states) scored **92.9%**; a correct reader answer scored **0%** — the metric is inverted.
- **tree** — the answer is a binary *Yes/No* wrapped in boilerplate (*"…indirect colleagues according to their graph relations"*). A correct *"No"* shares almost none of the boilerplate word-set, so polarity is never actually tested.

## Fix

New functions in `structmemeval_adapter.py`:
- `score_location_prediction` — correct iff the expected-state terms are present **AND** no excluded-distractor term leaks in.
- `score_tree_prediction` — Yes/No **polarity** match.
- `score_prediction(prediction, reference, task)` — dispatches by task; `accounting` / other fall back to the legacy heuristic (accounting needs a settlement-tuple scorer, tracked separately).

The legacy `check_state_correctness` and the default raw-retrieval scoring path (adapter line ~404) are **left unchanged** — properly fixing the default path requires wiring an in-process reader (the #891 no-reader-in-dispatcher issue), out of scope here.

**Validation:** under the fixed scorers the raw-retrieval baseline correctly drops to **0%** on both location and tree (vs the old 92.9% location recall). 10 unit tests in `tests/test_structmemeval_subtask_scorer.py`.

## Relationship to the merged StructMemEval back-fill (`61219415`)

The recently-merged `v3.0.1-structmemeval-backfill.json` (location 90.5% / accounting 0% / recommendations 15.4% / tree 0%) was produced by the **default raw-retrieval path** scored with the broken `check_state_correctness`. Those numbers are therefore raw-retrieval *recall* artifacts, not answer correctness — the location 90.5% is inflated by the exclusion-distractor matching described above, and the tree 0% reflects polarity never being tested. With the reader pipeline + these fixed scorers, location measures **~57%** and tree **~7%** (the latter dominated by retrieval trimming the connecting edges — all references are "Yes"). A correction note is posted on #899; re-measuring the merged capture with the fixed scorers is recommended as follow-up.

Refs #899.

### DIFF
--- a/benchmarks/structmemeval_adapter.py
+++ b/benchmarks/structmemeval_adapter.py
@@ -312,6 +312,136 @@ def check_state_correctness(retrieved: str, reference: ReferenceAnswer) -> bool:
 
 
 # ---------------------------------------------------------------------------
+# Task-aware prediction scoring (#899)
+#
+# `check_state_correctness` above is a single word-overlap heuristic over
+# the *whole* reference text. It is invalid for scoring a reader's
+# prediction for three independent reasons, one per task:
+#
+#   * location: the reference embeds an exclusion note ("should NOT
+#     mention <previous cities>"). Those distractor tokens land in the
+#     word-set, so raw retrieval (which contains every stale city)
+#     matches them and scores high, while a concise *correct* answer
+#     that omits them falls below the 0.5 threshold and scores 0.
+#   * tree: the answer is binary (Yes/No) wrapped in boilerplate
+#     ("...indirect colleagues according to their graph relations"). A
+#     correct "No" prediction shares almost none of the boilerplate
+#     word-set, so polarity is never actually tested.
+#   * accounting: the reference is a set of settlement tuples
+#     ("Charlie -> Bob : 218.33 EUR | ..."); word overlap on names and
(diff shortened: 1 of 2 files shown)


robotrocketscience:
[claim:review:garsecg:2026-05-22T20:17:31Z]

robotrocketscience:
[release:review:garsecg:2026-05-22T20:23:39Z]

PR #917, opened 2026-05-22
fix(bench): settlement-tuple scorer for StructMemEval accounting

robotrocketscience:
## What

Extends the task-aware StructMemEval scorer module (from #914, merged) with a **settlement-tuple scorer** for the accounting sub-task. Part of #899.

## Why

The legacy `check_state_correctness` word-overlap never checks settlement **amounts** or **directions** — so the merged capture's accounting 0% is uninformative. `score_accounting_prediction`:
- parses the reference (a `" | "`-joined list of acceptable alternative settlements) and the prediction into directed `(payer, payee, amount)` transaction sets,
- accepts both the reference arrow form (`Alice -> Bob : 91.00 EUR`) and the requested answer form (`Alice pays 91 to Bob`),
- returns correct iff the prediction matches **any** alternative within a 0.5 EUR tolerance (same cardinality; no extra/missing transactions).

Wired into `score_prediction` dispatch. 12 unit tests.

## Scope note — this lands the scoring *tool*, not an accounting number

A valid accounting **number** is not yet admissible. On the live data the reference settlements' net-balance **direction is inconsistent with the parsed expense log**: e.g. for `accounting_10%_1`, top-payer Alice nets **+188 EUR** (paid 1780 vs Bob 1377 / Charlie 1401), so Alice should be the creditor — but all three reference alternatives make **Bob** the creditor. Retrieval is **complete** (45/45 expense lines present in context), so this is *not* a retrieval gap. Either the benchmark applies a split/refund/settlement convention not yet reverse-engineered, and/or the three `|`-alternatives encode specific resolutions of the injected contradictory splits.

A reader-based reader pass scored 0/15 against these references, but I'm **not reporting that as a substrate/reader result** until the settlement convention is pinned — the discrepancy is in the reference↔expense-log mapping, not (demonstrably) in the reader. Tracked on #899.

This complements the merged #914 (location + tree scorers). Together they replace the invalid word-overlap for 3 of the 4 StructMemEval sub-tasks.

Refs #899.

### DIFF
--- a/benchmarks/structmemeval_adapter.py
+++ b/benchmarks/structmemeval_adapter.py
@@ -15,6 +15,7 @@ from __future__ import annotations
 
 import argparse
 import json
+import re
 import sys
 import tempfile
 import time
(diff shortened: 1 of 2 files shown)


robotrocketscience:
[claim:review:garsecg:2026-05-22T21:08:40Z]

robotrocketscience:
[release:review:garsecg:2026-05-22T21:10:41Z]

Issue #915, opened 2026-05-22
bench(structmemeval): re-measure per-sub-task back-fill with task-aware scorers (#914) + reader wiring


robotrocketscience:
## Problem

The merged StructMemEval per-sub-task back-fill (`benchmarks/results/v3.0.1-structmemeval-backfill.json`, commit `61219415`, PR #912) reports:

| Task | Score |
|---|---:|
| location | 90.5% |
| recommendations | 15.4% |
| accounting | 0.0% |
| tree | 0.0% |

These were produced by the **default raw-retrieval path** scored with `check_state_correctness`, which PR #914 established is invalid for these tasks:

- **location 90.5%** is inflated — `check_state_correctness` builds its word-set from the *entire* reference, including the exclusion note ("should NOT mention <previous cities>"). Raw retrieval carrying every stale city matches those distractor tokens and scores high, while a concise correct answer scores 0. The metric is inverted.
- **tree 0%** reflects polarity (Yes/No) never being tested — it is drowned by boilerplate in the word-overlap.

A correction note is on the now-closed #899, and the JSON `_calibration_notes` carries a correction banner (annotation PR, refs this issue). The numbers themselves remain unrevised on main.

## What this issue tracks

Re-measure the StructMemEval per-sub-task back-fill using the task-aware scorers shipped in #914 (`score_location_prediction`, `score_tree_prediction`, `score_prediction`), driven by an actual reader prediction rather than raw retrieval.

## Blocked on

The #914 scorers are not wired into the default adapter path (`structmemeval_adapter.py` ~line 404) — that path still scores raw retrieval with the legacy heuristic. Properly re-measuring requires an in-process reader to produce predictions for the scorers to grade. PR #914 references this as the "no-reader-in-dispatcher" gap (out of scope there).

## Acceptance

- [ ] Reader prediction wired into the StructMemEval adapter scoring path (or a documented bench harness that produces reader predictions).
- [ ] location and tree re-scored with `score_prediction`; accounting scorer (settlement-tuple) is separate follow-up.
- [ ] `v3.0.1-structmemeval-backfill.json` updated (or a new dated capture) with the corrected numbers and a methodology note distinguishing answer-correctness from raw-retrieval recall.

Refs #899, #912, #914.

robotrocketscience:
[...]

PR #912, opened 2026-05-22
exp(bench): StructMemEval per-sub-task back-fill — 90.5%/15.4%/0%/0% spread (#899)

robotrocketscience:
Closes #899 (StructMemEval per-sub-task row of the back-fill matrix — fourth and final non-AR back-fill alongside LRU PR #900, CR PR #907, TTL PR #908; Accurate_Retrieval tracked separately, in progress).

## Headline — striking spread, not the mean

| Task | Score | Queries | Cases (perfect/total) |
|---|---:|---:|---:|
| **location** | **90.5%** | 38/42 | 38/42 |
| **recommendations** | **15.4%** | 170/1104 | 0/84 |
| **accounting** | **0.0%** | 0/15 | 0/15 |
| **tree** | **0.0%** | 0/43 | 0/22 |

The spread is the load-bearing signal — not the mean.

## What the pattern means

Post-#605 deterministic-substrate signature:

- **location (90.5%)**: substrate excels at single-value state tracking when the answer is a noun-phrase grounded in retrieved prose. The 4 failures are state-transition cases where the most-recent state requires picking between conflicting prose statements; substrate has no recency-or-supersedes signal beyond ingest order.
- **recommendations (15.4% per-query, 0/84 perfect cases)**: signature of partial-retrieval-no-aggregation. Substrate surfaces SOME preference-relevant chunks, but no case has every relevant chunk surfaced together — recommendation tasks need preference *aggregation* across the full session history, which the chunk-projection lane lacks.
- **accounting (0.0%)**: requires arithmetic over a ledger history (sum / delta / running balance) that no chunk carries verbatim — the answer must be *derived* from the chunks, not retrieved from them. Substrate has no derivation lane.
- **tree (0.0%)**: requires hierarchical-knowledge traversal (parent→child→grandchild). Substrate has no graph-traversal primitive in retrieval.

## Alignment with 2026-05-22 #897 disposition

Lands the post-#605 honest baseline. The 0% / 0% on accounting + tree, and the 15.4% partial credit on recommendations, are the visible costs of the deterministic-narrow-surface ratification you reaffirmed earlier today (closing #895 as wontfix-via-philosophy). No re-open trigger under the current ratification — accounting / tree are structurally pinned at 0% by substrate design.

## Methodology

Per `benchmarks/run.py` CANONICAL_INVOCATIONS:

- `uv run --extra benchmarks python benchmarks/structmemeval_adapter.py --task <T> --bench big` for each of the 4 tasks
[...]

PR #918, opened 2026-05-22
feat(bench): reader-prediction scoring seam for StructMemEval

robotrocketscience:
## What

Adds the reader-prediction scoring seam for StructMemEval — the entry point that lets the task-aware scorers (#914 location/tree, #917 accounting) grade an actual reader's answers instead of raw retrieval.

- `score_predictions(predictions, ground_truth, default_task)` — matches reader predictions to the `--retrieve-only` ground-truth file on `(case_id, question)` and grades each with `score_prediction`. Returns an aggregate (`total` / `correct` / `score_pct` / `unmatched` / `per_query`).
- `--score-predictions PATH` CLI mode (pairs with `--ground-truth`, defaults to the `_gt` sibling). Skips retrieval/ingest.

## Why

The adapter scored raw retrieval with `check_state_correctness` (the invalid heuristic — see the #912 correction banner). The new scorers existed but had no way to consume reader output. This completes the existing `--retrieve-only` isolation workflow: retrieve context (no answers) → external reader → score predictions with the task-aware scorer.

**No reader runs in the adapter.** Predictions come from a sub-agent fed by an earlier `--retrieve-only` run, preserving the deterministic, no-API-key contract for bench adapters.

## Test plan

- 6 unit tests in `tests/test_structmemeval_reader_scoring.py`, incl. the load-bearing one: a raw-retrieval-style answer (names every stale city) now scores **wrong** because it routes through `score_location_prediction`, not `check_state_correctness`.
- CLI smoke verified locally (`--score-predictions` + `_gt` sibling inference).

## Scope — what this does NOT do (tracked on #915)

- Does **not** run the full multi-query re-measure (15–1104 queries/task via sub-agent reader) — that's a separate execution.
- Does **not** admit an accounting number — the settlement-convention discrepancy from #917 (reference net-balance vs expense log) is still unresolved.

Refs #915.

### DIFF
--- a/benchmarks/structmemeval_adapter.py
+++ b/benchmarks/structmemeval_adapter.py
@@ -523,6 +523,71 @@ def score_prediction(
     )
 
 
+def score_predictions(
+    predictions: list[dict[str, object]],
+    ground_truth: list[dict[str, object]],
+    default_task: str | None = None,
+) -> dict[str, object]:
+    """Score reader predictions against ground truth with `score_prediction`.
+
+    This is the reader seam that completes the `--retrieve-only` workflow
+    (#915): `--retrieve-only` writes context (NO answers) for an external
+    reader (a sub-agent — bench LLM calls do not run in this adapter), and
+    this function grades the reader's predictions with the task-aware
+    scorer instead of the raw-retrieval `check_state_correctness` path.
+
+    `predictions` items must carry `case_id`, `question`, and `prediction`;
+    `task` is read per-item when present, else `default_task`. `ground_truth`
+    items carry `case_id`, `question`, and `reference_answer` (the GT file
+    written alongside the retrieval file). Predictions are matched to GT on
+    the `(case_id, question)` key.
+
+    Returns an aggregate dict mirroring the `--output` schema: `task`,
+    `total`, `correct`, `score_pct`, `unmatched` (predictions with no GT
+    row), and `per_query` rows.
+    """
+    gt_index: dict[tuple[str, str], str] = {}
+    for row in ground_truth:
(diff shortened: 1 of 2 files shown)


yoshi280:
[claim:review:Setr:2026-05-22T21:37:37Z]

yoshi280:
Reviewed (have deep context here — authored #914/#917, the canonical-judge investigation on #915).

**Verdict: approve / ready-to-merge.** Correct and well-scoped: `score_predictions` completes the retrieve-only → external-reader → score loop using the task-aware `score_prediction` instead of the raw-retrieval `check_state_correctness` — the reader seam #915 needs and the #891-style fix for StructMemEval. Pure addition (+219/-0), all CI green, discretion clean.

Tests are strong — the load-bearing `test_raw_retrieval_style_answer_scores_wrong` locks exactly the inversion #914 fixed. Ran them locally: 6/6 pass.

Two non-blocking notes:
1. Matching on `(case_id, question)`: each StructMemEval case has 1 query, so no collision; but if a case ever had two identical question strings, `gt_index` would keep only the last. Fine for current data.
2. Accounting routes task-aware only once #917 merges; until then accounting predictions fall back to the legacy heuristic. Not a defect in this PR (it uses whatever `score_prediction` exposes) — just sequencing.

Independently corroborated by my apples-to-apples run on #915: the StructMemEval repo's own judge (`judge/prompt.txt`, lenient binary key-fact) gives the same per-task numbers this seam's scorers produce. Good to merge.

yoshi280:
[release:review:Setr:2026-05-22T21:39:17Z]

Issue #935, opened 2026-06-04
feat(audit): aelf:scope-out --vs-claude-memory — cross-memory dedup audit


yoshi280:
## Motivation

User concern (dadvideoessay):
> *\"there's also a chance that in my Claude memory for this repo there's some bad usage shit.\"*

Claude Code's auto-memory (\`~/.claude/projects/<project>/memory/MEMORY.md\` + per-memory \`.md\` files) and \`.claude/CLAUDE.md\` per-project rules both store assertions outside aelfrice. There is no surface that compares the two stores, so duplicates and contradictions across them stay invisible.

## Surface

Extend `aelf scope-out` (or new `aelf audit-claude-memory` if scope-out semantics don't fit). Read-only audit.

```
aelf scope-out --vs-claude-memory [--project PATH]
```

Steps:
1. Parse \`~/.claude/projects/<project>/memory/MEMORY.md\` bullets into pseudo-belief rows.
2. Run exact-slot equality check against locked aelfrice beliefs (same primitive as proposed contradiction column).
3. Emit a report with four buckets:
   - **Potential duplicates** — same slot, same value in both stores
   - **Potential contradictions** — same slot, different values
   - **aelfrice-only** — not present in Claude memory
   - **Claude-memory-only** — not present in aelfrice

No automatic sync. Pure report.

## #605 + #606 fit

Parser + exact-slot match. No fuzzy / embedding / LLM judge. Subcommand only — no background sync.

## Acceptance criteria

- [ ] Parses standard Claude Code MEMORY.md bullet format
- [ ] Slot-equality check uses same code path as the `[!]` contradiction column (proposal in coverage_extension)
- [ ] Four-bucket report in human + JSON modes
- [ ] Read-only — no writes to either store

## Dependency

This needs the exact-slot-equality primitive that the contradiction-column issue also needs. File whichever lands first as a prerequisite.

## Files touched

- `src/aelfrice/audit_claude_memory.py` (new — parser + comparator)
- `src/aelfrice/cli.py` (extend `_cmd_scope_out` or add `_cmd_audit_claude_memory`)
- `tests/test_audit_claude_memory.py` (new)

robotrocketscience:
[claim:issue:garsecg:2026-06-04T18:47:54Z]

robotrocketscience:
[release:issue:garsecg:2026-06-04T18:48:37Z]

robotrocketscience:
[...]

PR #946, opened 2026-06-04
feat(search): AELF_SHOW_CONFLICTS — [!] slot-conflict column on aelf search (#938)

### LINKED ISSUE #938: feat(search): [!] contradiction-flag column on aelf:search via exact-slot equality
## Motivation

Visibility into contradictions across the belief store — surface when a search hit conflicts with a locked belief. The #422 \"contradiction-detector\" wall was paraphrase / synonymy detection (an embedding-shaped problem). This proposal sidesteps that by restricting to **exact (subject, predicate) slot equality** — regex-tractable when the entity-extractor has already typed-slotted both rows.

## Surface

When `aelf search QUERY` returns a hit B, check:

> Does B share a (subject, predicate) tuple with any locked belief whose object differs from B's?

If yes, render `[!]` marker next to B in the listing. In `--json` mode, add `slot_conflict_with: <locked_id>` field.

Gate behind `AELF_SHOW_CONFLICTS=1` env var (default off — feature is informational, surfaces unaudited slot pairs).

## Reuse

The same `_slot_conflict(belief, locked_set)` primitive powers the cross-Claude-memory audit. Whichever ships first carries the primitive in `src/aelfrice/contradiction.py` (module already exists).

## #605 fit

Exact-slot equality. Regex / dict-lookup tractable. Bounded false-negatives (different slots that *mean* the same thing won't match — that's the embedding lane, closed by #605). **Zero false-positives by construction** — slot-equality is a hard match.

## #606 fit

Default off, env-var opt-in — the surface adds visual noise to every search result, so it earns the opt-in posture even though it's not posterior-mutating.

## Acceptance criteria

- [ ] `_slot_conflict(belief, locked_set) -> Optional[str]` pure function in `contradiction.py`, returns locked-belief ID or None
- [ ] `aelf search` renders `[!]` when env var set
- [ ] `--json` mode adds `slot_conflict_with` field when present
- [ ] Latency: ≤2ms p95 added per query (locked set is small; in-memory dict)
- [ ] Test fixture covers: matching slot+different object → flag; matching slot+same object → no flag; different slot → no flag

## Files touched

- `src/aelfrice/contradiction.py` (add `_slot_conflict`)
- `src/aelfrice/cli.py` (`_cmd_search` rendering — ~10 LOC)
- `tests/test_slot_conflict.py` (new)

robotrocketscience:
Closes #938.

## What

Adds `_slot_conflict(belief, locked_set) -> Optional[str]` to `contradiction.py` and wires it into `_cmd_search` behind the `AELF_SHOW_CONFLICTS=1` env var. When set, hits whose value-slots collide with a locked belief's get a `[!]` marker; in `--json` mode, the row gains a `slot_conflict_with: <locked_id>` field.

```
$ AELF_SHOW_CONFLICTS=1 aelf search "rebuild p99"
[locked]  abc123: rebuild p99 is 4.5 ms
          [!] def456: rebuild p99 is 12 ms
```

## How

The exact-slot-equality primitive already shipped in `src/aelfrice/value_compare.py`: `ValueSlots` dataclass, `extract_values(text) -> ValueSlots`, `find_conflicts(slots_a, slots_b) -> tuple[SlotConflict, ...]`. `_slot_conflict` is a thin wrapper — extract the belief's slots, loop over the locked set, return the first id whose `find_conflicts` is non-empty.

`_cmd_search` pre-extracts the locked-set slots once per invocation and calls a sister helper `_slot_conflict_preextracted(belief, locked_pairs)` per hit (hot path). Both functions live in `contradiction.py` and are tested separately. When `AELF_SHOW_CONFLICTS` is unset, neither runs — zero overhead on the default search path.

## Spec deviations

1. **`aelf search` gained `--json`.** The issue body says "In `--json` mode, add `slot_conflict_with` field" but the search subparser didn't actually have `--json` on `main`. Added as a precondition. Output shape is `print(json.dumps(rows, indent=2))` — an array, not JSONL. This matches the closest existing payload-style emitters in `cli.py` but is a different convention from `_cmd_speculative` / `_cmd_locked`'s per-line JSONL. Worth a future harmonisation pass; not blocking here.
2. **Latency test asserts ≤5 ms, spec asks ≤2 ms.** Measured p95 on 10 hits × 5 locked, 100 iterations: ~0.05-0.1 ms. The 5 ms threshold is a CI-hardware safety margin; the real measurement is comfortably under the spec. The comment in the test calls this out — happy to tighten if reviewer prefers.
3. **A second helper `_slot_conflict_preextracted` was added** so the call site doesn't re-run `extract_values` per locked belief per hit. `_slot_conflict` (spec signature) remains the public, tested function; the hot-path variant is for callers who already extracted the locked-set slots.

## Verification

- `pytest tests/test_slot_conflict.py tests/test_cli.py -x` → 70 passed.
- Discretion grep on diff: clean.
- 12 new unit tests cover: same slot/same object (no flag), same slot/different object (flag), different slots (no flag), empty locked set (no flag), first-match-wins ordering, latency p95 budget, env-var on/off rendering, JSON `slot_conflict_with` field present/absent.

## Out of scope

- Cross-Claude-memory dedup audit (#935 — same primitive will compose into that issue's surface).
- Promoting `SlotConflict` details into the JSON row (only the locked-id is exposed per spec).
- Per-PR JSONL/array harmonisation across all `--json` outputs (separate concern).

## Summary by Sourcery

Add optional slot-conflict detection to `aelf search` and expose conflicts in both human and JSON output, gated by an environment flag.

New Features:
- Introduce `_slot_conflict` and `_slot_conflict_preextracted` helpers to detect value-slot conflicts between beliefs and a locked set.
- Extend `aelf search` with a `--json` option to emit results as a JSON array including optional conflict metadata.
- Add an `AELF_SHOW_CONFLICTS`-controlled `[!]` marker and `slot_conflict_with` field to highlight hits that conflict with locked beliefs in search results.

Tests:
- Add unit tests for slot-conflict helpers covering conflict/no-conflict scenarios, ordering semantics, empty locked set, and latency bounds.
- Add CLI tests to verify the conflict marker in text output and `slot_conflict_with` behavior in JSON output with `AELF_SHOW_CONFLICTS` on and off.

### DIFF
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -753,7 +753,8 @@ def _cmd_search(args: argparse.Namespace, out: object) -> int:
         else:
             n_beliefs = -1  # not consulted on the success path
         # #938: pre-extract locked-set slots once per invocation (hot path).
-        # Zero cost when feature is off.
+        # Zero cost when feature is off — name only bound under the gate;
+        # all reads of locked_pairs below are inside `if show_conflicts:`.
         if show_conflicts:
             from aelfrice.contradiction import _slot_conflict_preextracted
             from aelfrice.value_compare import extract_values
@@ -761,8 +762,6 @@ def _cmd_search(args: argparse.Namespace, out: object) -> int:
             locked_pairs = [
                 (b, extract_values(b.content)) for b in locked_beliefs
             ]
-        else:
-            locked_pairs = []
     finally:
         store.close()
     if not hits and not peer_hits:
--- a/src/aelfrice/cli.py
+++ b/src/aelfrice/cli.py
@@ -738,6 +738,7 @@ def _run_llm_onboard(
 
 
 def _cmd_search(args: argparse.Namespace, out: object) -> int:
+    show_conflicts = os.environ.get("AELF_SHOW_CONFLICTS", "0") == "1"
     store = _open_store()
     try:
         hits = retrieve(store, args.query, token_budget=args.budget)
@@ -751,6 +752,17 @@ def _cmd_search(args: argparse.Namespace, out: object) -> int:
             n_beliefs = store.count_beliefs()
         else:
             n_beliefs = -1  # not consulted on the success path
+        # #938: pre-extract locked-set slots once per invocation (hot path).
+        # Zero cost when feature is off.
+        if show_conflicts:
+            from aelfrice.contradiction import _slot_conflict_preextracted
+            from aelfrice.value_compare import extract_values
+            locked_beliefs = store.list_locked_beliefs()
+            locked_pairs = [
+                (b, extract_values(b.content)) for b in locked_beliefs
+            ]
+        else:
+            locked_pairs = []
     finally:
         store.close()
     if not hits and not peer_hits:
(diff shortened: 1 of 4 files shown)


robotrocketscience:
[claim:review:garsecg:2026-06-04T19:56:38Z]

robotrocketscience (commented):
## Verdict

**Approve.** Correct reuse of the existing `value_compare` primitive; new code lives where you'd want it (`contradiction.py`); env-var gate keeps the default path zero-overhead. Four CodeQL nits below are non-blocking and trivially fixable in a follow-up if you don't want to bother on this PR.

### Why it works

- `_slot_conflict(belief, locked_set)` is the pure-function the issue specified, and `_slot_conflict_preextracted(belief, locked_pairs)` is the right hot-path variant — `extract_values` runs once per locked belief per query, not per hit. Good call.
- `--json` addition to `aelf search` is a sane spec deviation; the issue body referenced a `--json` mode that didn't exist on `main`, so adding it as a precondition is the only way to satisfy "add `slot_conflict_with` field" without inventing JSONL semantics.
- Tests cover all three states (env on → marker, env off → no marker, JSON → field present).
- Discretion grep clean.

### CodeQL nits (follow-up acceptable)

1. `src/aelfrice/cli.py:765` — `locked_pairs = []` in the `else` branch is dead. All consumers are inside `if show_conflicts:`. Drop the else branch and only define `locked_pairs` when `show_conflicts`.
2. `tests/test_slot_conflict.py:15` — unused import `Sequence`.
3. `tests/test_slot_conflict.py:17` — unused import `pytest`.
4. `tests/test_slot_conflict.py:26` — unused import `ValueSlots`.

### Note for cross-PR awareness

PR #947 introduces a separate slot-extraction primitive in `claude_memory.py` (subject/predicate token-prefix) for cross-store dedup. The two primitives have non-overlapping domains — `value_compare` is typed numeric/enum, `claude_memory.extract_slot` is text-prefix — so they coexist, but if a future PR wants to unify them, the typed-slot one in `value_compare.py` is the more conservative substrate (per #605 PHILOSOPHY).

Tagging `ready-to-merge`.

robotrocketscience:
[release:review:garsecg:2026-06-04T19:58:23Z]

Issue #980, opened 2026-06-23
Phantom belief lifecycle audit + trigger-driven generation in normal turns


robotrocketscience:
Tracking issue: audit the phantom (`speculative`-origin) belief lifecycle end to end, and address under-exploitation.

## Motivation
Operator observation: phantoms don't seem to show up recently. Two questions:
1. **Are phantoms being generated, ranked, and either promoted to real beliefs or discarded** when appropriate? (full lifecycle audit)
2. **Generation is too narrow.** Phantoms should be generated by a **trigger during normal conversation turns with the LLM**, not only via the explicit `/aelf:wonder` command. `/aelf:wonder` should generate a LOT of phantoms; separately, there should be an automatic trigger mechanism. We are likely not fully exploiting phantom beliefs.

## Lifecycle stages under audit
- [ ] **Generation** — `/aelf:wonder` -> `wonder_ingest` -> `origin='speculative'`; initial (α,β)/scope; is there ANY normal-turn / automatic trigger today?
- [ ] **Ranking / retrieval** — are speculative beliefs retrieval-eligible, discounted, or filtered out of the hook-injected block (so the user never sees them)?
- [ ] **Promotion** — `aelf promote` and lock-auto-promote (#550, Jaccard ≥ 0.9); wired end-to-end?
- [ ] **Discard / GC** — `wonder_gc` / `aelf stale`: automatic or manual-only? default-on or opt-in? do stale phantoms accumulate forever?

## Empirical check (operator's stores)
- [ ] Count speculative beliefs per DB, latest `created_at`, any promoted (origin transitions), any GC'd. Confirm whether generation is actually happening in practice.

## Design gap to evaluate
- [ ] Specify a **trigger-driven phantom generation** mechanism for normal turns (what fires it: gaps detected at retrieval time? uncertainty? new entities? contradiction?), bounded so it does not spam the store. Determinism/cost constraints (LLM dispatch runs under the host agent's credentials, not in aelfrice's deterministic retrieval path).

## Method
Read-only lifecycle map via subagent (running) + direct SQLite queries across the operator's aelfrice DBs (running). Results land here.

## Related
`/aelf:wonder`, `/aelf:reason`, #550 (lock auto-promote), origin tiers in `models.py`.

_Status: investigation running; will update with the lifecycle map, the data, and a trigger-generation proposal._

robotrocketscience:
## Findings: phantom lifecycle is wired but barely exercised, and generation is too narrow

[...]

Issue #936, opened 2026-06-04
feat(review): weekly .aelfrice/review.md checkbox file + aelf:review --apply


yoshi280:
## Motivation

dadvideoessay's *\"curious how out of date it is\"* — `aelf:stale` (#933) answers the read side. This is the write-back side: a way to confirm / remove / lock the stale candidates *at the user's cadence*, without aelfrice hijacking the prompt loop with mid-session \"still true?\" interruptions.

## Surface

Pattern borrowed from `crontab -e` / dotfile-config tools.

**Generate:**
```
aelf review --generate    # writes .aelfrice/review.md
```

The file contains the 10 oldest-retrieved-but-unconfirmed beliefs as a checkbox-list:

```markdown
## aelfrice review — generated 2026-06-04
For each belief: leave checkboxes empty to no-op; check exactly one verdict per row.

- [ ] keep   [ ] remove   [ ] lock   | abc123 (47d old, 21d cold) — pipx is no longer a supported install path...
- [ ] keep   [ ] remove   [ ] lock   | def456 (62d old, 35d cold) — atomic commits beat batched
```

User edits the file at their leisure.

**Apply:**
```
aelf review --apply [path]    # default: .aelfrice/review.md
```

Reads checkboxes, dispatches:
- `keep` → `last_confirmed_at = now()` (new column, scoped to this feature)
- `remove` → soft-delete with audit row
- `lock` → promote to L0
- No box checked → no-op (skip row, leave for next review)

## #605 + #606 fit

File I/O is deterministic. No model inference. File is not written unless `aelf:review --generate` or `AELFRICE_WEEKLY_REVIEW=1` (cron-driven generate). Pure subcommand surface — no ambient hook.

## Dependency

Cleanly composes with `aelf:stale` (#933) — the candidate list is the same query, just rendered as a markdown checkbox file instead of a CLI table.

## Acceptance criteria

- [ ] `aelf review --generate` writes `.aelfrice/review.md` with current 10 candidates
- [ ] `aelf review --apply` parses checkboxes; rejects ambiguous rows (multiple boxes checked) with a clear error
- [ ] Partial edits OK — unchecked rows are skipped, no rewrite needed
- [ ] Adds `last_confirmed_at` column to belief table (migration)
- [ ] `aelf:stale` excludes beliefs with recent `last_confirmed_at` from default output

## Files touched

- `src/aelfrice/review.py` (new — generator + applier)
[...]

PR #987, opened 2026-06-23
fix(store): exclude soft-deleted beliefs from all retrieval lanes (#980)

robotrocketscience:
## Summary

Soft-deleted beliefs (`valid_to NOT NULL`) were retrievable through **every**
candidate-generation lane. This is the correctness precursor in #980's
recommended-work list (item 1) — the fix that must land before auto-GC is
turned on.

A belief is soft-deleted by two production paths, both via
`MemoryStore.soft_delete_belief`:
- `wonder_gc` — stale-phantom garbage collection (`aelf wonder --gc`)
- `aelf review` with a **`remove`** verdict (shipped v3.5, #936) — a
  user-facing path, so this is not phantom-only.

`retrieval.py` applies **no** `valid_to` filter anywhere, so a soft-deleted
belief surfaced in results and in the hook-injected block.

## Scope correction vs the #980 findings comment

The findings comment scoped the leak to `search_beliefs` / `search_beliefs_scored`
(the FTS lane). Investigation found it is **systemic** — the soft-deleted
belief leaks through all four candidate lanes, including the **default-on**
BM25F lane:

| Lane | Source | Leaked? |
|---|---|---|
| L1 FTS | `search_beliefs`, `search_beliefs_scored` | yes |
| L1 BM25F (default-on) | `list_beliefs_for_indexing` → `BM25Index` | yes |
| L2.5 entity index | `lookup_entities` | yes |
| L0 locked | `list_locked_beliefs` | yes |

`list_beliefs_for_indexing` also feeds `dedup` and `relationship_detector`,
which likewise should not see soft-deleted rows.

## Fix

Add `valid_to IS NULL` at each read boundary, and prune the derived
`beliefs_fts` row in `soft_delete_belief` (mirrors `delete_belief`).

**Audit trail preserved:** the `beliefs`, `edges`, `belief_entities`, and
corroboration rows are all retained — only the derived FTS index is pruned and
reads are filtered. `get_belief(id)` still returns the soft-deleted row.

**Determinism preserved:** `list_beliefs_for_indexing` still returns an
id-ASC set; it is just filtered. Empirically a **no-op on all current data** —
0 soft-deleted beliefs exist across local stores — so no existing test or
bench corpus changes. Latent today, but real: the first `aelf review` remove
or `aelf wonder --gc` would otherwise leave the removed belief retrievable.

## Tests

`tests/test_soft_deleted_retrieval_exclusion_980.py` pins the invariant at
every lane (FTS ×2, `list_beliefs_for_indexing` + `BM25Index`, `lookup_entities`,
`list_locked_beliefs`) plus an end-to-end `retrieve()`, and asserts
`soft_delete` prunes the FTS row while the `beliefs` / `belief_entities` audit
rows survive.

## Scope

Part of #980 (recommended-work **item 1**). Does **not** close the umbrella —
items 2–4 (make GC run, trigger-driven generation, observability) remain.

## Summary by Sourcery

Exclude soft-deleted beliefs from all retrieval and indexing paths while preserving their audit trail.

Bug Fixes:
- Ensure soft-deleted beliefs are filtered out from FTS search, scored search, entity lookup, locked-belief listing, and BM25/BM25F indexing.
- Prune the corresponding FTS index row when a belief is soft-deleted so it no longer participates in keyword search.

Enhancements:
- Clarify store and indexing docstrings around active vs soft-deleted beliefs and retrieval behavior.

Tests:
- Add regression tests to assert soft-deleted beliefs are excluded from every retrieval lane and BM25 index while their underlying audit rows remain accessible.

### DIFF
--- a/src/aelfrice/store.py
+++ b/src/aelfrice/store.py
@@ -2040,14 +2040,22 @@ class MemoryStore:
 
         Only moves beliefs from active (valid_to IS NULL) to soft-deleted.
         Calling this on an already-GC'd belief is a no-op (idempotent via
-        the WHERE clause). Does not remove edges or corroboration rows —
-        the phantom's evidence trail is preserved for audit.
+        the WHERE clause). Does not remove edges, entity-index, or
+        corroboration rows — the phantom's evidence trail is preserved for
+        audit. The derived FTS index row *is* pruned (mirrors
+        ``delete_belief``) so the soft-deleted belief stops matching
+        keyword search; read-side queries also filter ``valid_to IS NULL``,
+        so this is index hygiene plus defense-in-depth (#980).
         """
         now = ts if ts is not None else datetime.now(timezone.utc).isoformat()
-        self._conn.execute(
+        cur = self._conn.execute(
             "UPDATE beliefs SET valid_to = ? WHERE id = ? AND valid_to IS NULL",
             (now, belief_id),
         )
+        if cur.rowcount > 0:
+            self._conn.execute(
+                "DELETE FROM beliefs_fts WHERE id = ?", (belief_id,)
+            )
         self._bump_belief_version(belief_id)
         self._conn.commit()
         self._fire_invalidation()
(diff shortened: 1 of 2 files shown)


robotrocketscience:
[claim:review:Gylf:2026-06-23T03:49:25Z]

robotrocketscience:
## Review

Verified the correctness fix and the rebase. **Local hot path: fully covered.**

- All four candidate-generation lanes now filter `valid_to IS NULL`: FTS (`search_beliefs`/`search_beliefs_scored`), BM25F (`list_beliefs_for_indexing` → `BM25Index`), entity index (`lookup_entities`), and the locked lane (`list_locked_beliefs`).
- `soft_delete_belief`'s FTS prune (`DELETE FROM beliefs_fts WHERE id = ?` guarded on `rowcount > 0`) matches `delete_belief` exactly, and the FTS table is manually managed (no content-sync triggers to conflict with), so the prune is safe and idempotent.
- Every `retrieve()` path that hydrates by id (`get_belief`) draws its ids from one of the now-filtered lanes; graph/cluster expansion operates over the already-filtered `l1` set via `l1_by_id`, so no soft-deleted belief can re-enter through edges. Audit rows (`beliefs`, `belief_entities`) are correctly retained.
- Rebased onto `github/main` (was 1 commit behind: the `uv.lock` pyjwt bump — disjoint files, clean rebase). Both commits re-signed; new `tests/test_soft_deleted_retrieval_exclusion_980.py` (9 tests) passes locally.

### One follow-up (non-blocking, out of this PR's stated scope)

`MemoryStore.search_peer_beliefs` (the federated FTS lane, reached via `aelf search` when peers are configured) is **not** filtered — its peer query has no `b.valid_to IS NULL` predicate:

```
FROM beliefs b
JOIN beliefs_fts f ON f.id = b.id
WHERE beliefs_fts MATCH ?
{scope_sql}
```

So a *peer's* soft-deleted belief can still surface through federation. It's a fifth FTS read path beyond the four enumerated here, low-severity (opt-in federation, cross-store — the host can't soft-delete a peer's row anyway), and a distinct concern from local soft-delete hygiene. Suggest tracking it as a separate item under #980 rather than expanding this PR. @robotrocketscience flagging for disposition.

LGTM on the local-lane fix; proceeding to merge once CI is green on the rebased SHA.

robotrocketscience:
[claim:review:Setr:2026-06-23T03:54:10Z]

robotrocketscience:
[release:review:Setr:2026-06-23T03:54:14Z]

robotrocketscience:
[release:review:Gylf:2026-06-23T04:03:11Z]

robotrocketscience:
[release:review:Gylf:2026-06-23T04:57:59Z]

Issue #606, opened 2026-05-10
v3.0: sentiment-feedback hook production wire-up (evaluation #193 passed; integration pending)


robotrocketscience:
## Scope

Wire `aelfrice.sentiment` (the sentiment-feedback module shipped at v2.0.0 per CHANGELOG) into a production `UserPromptSubmit` / `Stop` hook lane so corrections detected from prose (e.g. "no", "stop doing that") feed `apply_feedback` against the active retrieval window automatically.

## Background

- v2.0.0 CHANGELOG: "sentiment-feedback module (hook integration pending)". Module exists; nothing calls it on a real hook event.
- #193 evaluation closed COMPLETED 2026-05-03 — the privacy + determinism review passed. The hook-integration wire-up never got its own ticket; the README v2.2 line listed it but no actionable scope existed.

## Acceptance (sketch — needs spec)

1. **Spec memo.** Pick hook lane: `UserPromptSubmit` (corrections in the next prompt feed back against the prior turn's retrieval) vs `Stop` (corrections at end of turn). Pick decay policy: which retrieval window's beliefs receive the α/β bump.
2. **Determinism property.** Sentiment classifier is deterministic stdlib (matches the existing `aelfrice.sentiment` contract); no embedding/LLM in the hot path.
3. **Privacy property.** Hook respects existing transcript-ingest opt-out (`--no-transcript-ingest`); no new PII surface beyond what the transcript-logger already touches.
4. **Bench.** Synthetic two-session fixture: session A locks correction; session B replays the same query; retrieval ranks the correction higher. No regression on the existing posterior-ranking eval harness.
5. **Audit.** Every sentiment-driven `apply_feedback` row writes an `audit_log` entry tagged `sentiment_feedback_hook`.

## Out of scope

- Cross-session sentiment propagation (sentiment in session N affects session M ranking). Single-session for v3.0; cross-session is a follow-on.
- Multi-language sentiment. English-only mirror of existing module's surface.

## Refs

- `src/aelfrice/sentiment.py` — module
- `docs/v2_sentiment_feedback.md` — design memo (covers determinism + privacy review for #193)
- #193 — evaluation gate (CLOSED COMPLETED)

robotrocketscience:
[claim:issue:leibniz:2026-05-10T23:33:06Z]

robotrocketscience:
[claim:issue:godel:2026-05-10T23:33:39Z]

robotrocketscience:
[...]

Issue #265, opened 2026-04-28
[v2.x] View-flip — beliefs/edges become materialized; ingest_log is canonical


robotrocketscience:
## Summary

The terminal step of the spec at
`docs/design/write-log-as-truth.md`: flip view authority. After this
issue merges, `ingest_log` is canonical state; `beliefs`/`edges` are
materialized views maintained by the worker. The contract change
the whole chain has been preparing.

## What "flip authority" means concretely

- `MemoryStore.insert_belief()` becomes private (or removed from
  the public surface). Direct belief inserts are forbidden outside
  the derivation worker.
- All retrieval reads continue to query `beliefs` (the materialized
  view). User-facing latency is unchanged.
- `aelf doctor` runs `replay_full_equality` (#262) on every open as
  a sanity check — drift should always be zero on a healthy store.
- A new CLI surface, `aelf rebuild --rule-set <hash>`, drops
  canonical state and re-derives from log under the named (or
  current) rule set. This is what makes "fix the classifier, replay
  the log" actually work.

## Why this is the payoff

Four user-visible capabilities open up:

1. **Classifier bug fix → automatic re-derive.** Update the rules,
   run `aelf rebuild`, every belief in the store reflects the new
   rules. No re-onboard.
2. **Audit trail.** `aelf doctor --explain <belief_id>` prints the
   `ingest_log` row(s) that produced the belief — source path,
   classifier version, rule-set hash, raw text, timestamp.
3. **Time-travel queries.** `aelf rebuild --as-of 2026-03-01` drops
   canonical and rebuilds from log up to that timestamp under the
   rule set in effect at that time. Answers "what would the agent
   have retrieved on this query last March."
4. **Federation atom.** v3's cross-project replication ships log
   rows between scopes; each scope materializes locally. Without
   the flip, federation has no atomic event to ship.

## Why queen / why this is the riskiest issue

The chain's correctness has been verified up to this point against
canonical-as-authority. Flipping authority means canonical is now a
function of the log; if the worker has a bug we missed in #262/#264,
canonical drifts silently. Mitigations:

1. **Land behind a feature flag for at least one minor version.**
   `aelfrice.config.WRITE_LOG_AUTHORITATIVE=False` by default;
   opt-in for early adopters; flip the default after one stable
   release with no drift reports.
2. **Drift alarm in `aelf doctor`.** Continuous self-check: if
   `replay_full_equality` ever reports drift on a flag-enabled store,
   surface as a P0.
[...]

PR #992, opened 2026-06-23
feat: mirror claude-memory writes into the belief graph (write-through hook) (#985)

### LINKED ISSUE #985: feat: mirror Claude Code memory writes into the aelfrice belief graph (write-through hook)
## Motivation

Claude Code ships an on-by-default auto-memory: it writes one-fact markdown
files (with `name`/`description`/`metadata.type` frontmatter) into
`~/.claude/projects/<project>/memory/` plus a `MEMORY.md` index. aelfrice
already acknowledges this coexistence — `/aelf:audit-claude-memory` does a
cross-store dedup audit between locked aelfrice beliefs and the `MEMORY.md`
index.

But today the two stores **drift**: a fact written to Claude Code memory is *not*
in the aelfrice belief graph unless it is separately `aelf lock`'d / ingested by
hand. The audit skill can *detect* the divergence after the fact, but nothing
keeps them in sync at write time. (Concretely: durable findings get written to
Claude Code memory and then have to be manually re-locked into the graph to be
permanent — the audit's "store-exclusive" class is exactly this gap.)

## Proposal

Add a write-through so that **every Claude Code memory write is mirrored into the
aelfrice belief graph automatically**. Two candidate surfaces:

1. A Claude Code `PostToolUse` hook keyed on `Write`/`Edit` whose path is under
   the `…/memory/` dir → ingest the written file's body as a belief.
2. The existing aelfrice `SessionStart`/`Stop` hook scans the memory dir for
   new/changed files (diff against a last-seen manifest) and ingests them.

Either way it is a **one-way mirror** (Claude Code memory → aelfrice graph);
aelfrice is not authoritative over the memory files.

## Design questions to settle

- **Frontmatter → belief mapping.** `name`/`description` → belief content/anchor;
  `metadata.type` (`user`/`feedback`/`project`/`reference`) → belief `type` /
  `origin`. Does `type: user` imply a lock (L0), or just `origin=user_validated`?
- **Idempotency.** The `content_hash` UNIQUE constraint already dedupes exact
  matches; re-writes / edits of the same memory must not spawn duplicates
  (update-in-place vs SUPERSEDES edge).
- **Opt-in/out.** A flag (`[memory] mirror_claude_memory` in `.aelfrice.toml`
  and/or `AELFRICE_MIRROR_CLAUDE_MEMORY`), default TBD.
- **Relationship to `/aelf:audit-claude-memory`.** With write-through live, the
  audit becomes a reconciliation/repair tool for pre-existing drift rather than
  the only bridge.

## Acceptance

- Writing a memory file under the Claude Code memory dir results in a
  corresponding belief in the aelfrice graph, idempotently.
- The frontmatter→belief mapping is documented (likely under `docs/design/`).
- Behavior is flag-gated.

## Related

- `/aelf:audit-claude-memory` (existing cross-store audit).
- #529 (auto-capture, closed) — related but distinct: that was about capturing
  *session* content; this is about mirroring the *Claude Code memory store*.

robotrocketscience:
## What

Closes #985. A one-way write-through that mirrors each per-memory fact file
written under `~/.claude/projects/<encoded>/memory/` into the aelfrice belief
graph, so the two stores stop drifting at write time (today they only
reconcile after the fact via `/aelf:audit-claude-memory`).

## Design decisions (operator-ratified 2026-06-23)

- **Trigger surface:** `PostToolUse:Write|Edit|MultiEdit` hook, real-time,
  path-filtered to the memory dir (skips `MEMORY.md`).
- **`metadata.type` → origin:** `user`/`feedback` → `user_validated`
  (undeflated prior); `project`/`reference`/absent → `agent_inferred`
  (deflated). **Never locks** — L0 stays reserved for explicit `aelf lock`.
- **Default OFF / opt-in:** `is_mirror_enabled` resolves
  `AELFRICE_MIRROR_CLAUDE_MEMORY` env > kwarg > `[memory] mirror_claude_memory`
  TOML > `False`. The hook entry installs by default but is **inert** until the
  flag is set (cheap path-shape reject on every other Write/Edit). Consistent
  with #605 narrow-surface and the ADR-0003-dec-4 / #606 opt-in posture.

## How it stays replay-safe

`replay.py` nulls `raw_meta` before re-deriving, so origin logic that reads the
frontmatter can't live in the deterministic `derive()` path. Instead the hook
freezes the `user_validated` decision as a `route_overrides` (#265) — which
replay already round-trips — and the deterministic path yields `agent_inferred`.
`derive()` therefore carries **no** `claude_memory`-specific branch.

## Idempotency

Content-derived belief id + `insert_or_corroborate`: a byte-identical re-write
corroborates instead of duplicating. Edits mint a new belief; supersession is
left to the consuming agent (#605). `update-in-place`/`SUPERSEDES` are out of
scope for v1.

## Tests

- `tests/test_claude_memory_mirror.py` — parser, path classifiers, flag
  precedence, hook end-to-end (origin mapping, idempotency, flag-off and filter
  no-ops). All fixtures **synthetic**; no `~/.claude` content committed.
- `claude_memory_v0_1.jsonl` replay-soak corpus (deterministic path).
- Enum/table assertions extended (corroboration sources, retention defaults,
  auto-capture opt-out set).
- Full suite green (5247 passed).

## Acceptance (#985)

- [x] Writing a memory file results in a corresponding belief, idempotently.
- [x] Frontmatter→belief mapping documented (`docs/design/claude_memory_mirror.md`).
- [x] Behaviour is flag-gated (default off).

## Summary by Sourcery

Mirror writes from the upstream auto-memory tool’s per-memory markdown files into the local belief graph via an opt-in PostToolUse hook, including configuration, ingestion source wiring, and documentation.

New Features:
- Introduce parsing and classification utilities for upstream auto-memory per-memory markdown files and a write-through mirror flag.
- Add a PostToolUse hook and CLI/auto-install wiring to mirror eligible memory writes into the belief graph as a new claude_memory ingest source.

Enhancements:
- Extend models and derivation worker enums to support the new claude_memory ingest and corroboration source types with appropriate retention defaults.
- Document the claude-memory write-through behavior and frontmatter-to-belief mapping in a new design doc.

Tests:
- Add comprehensive tests for memory file parsing, path detection, mirror flag resolution, and end-to-end hook behavior, plus replay-soak and enum/retention coverage updates.


robotrocketscience:
[claim:review:idnn:2026-06-23T04:24:05Z]

robotrocketscience:
**Review: pass (rebased to unblock merge).**

Verified:
- Replay-safety is correct. The `user`/`feedback` → `user_validated` decision is frozen as a `route_overrides` (round-tripped by replay per #265); `project`/`reference`/absent fall through to the deterministic classifier → `agent_inferred`. `derive()` carries no `claude_memory` branch and never reads `raw_meta`, so it stays replay-equality-stable. The replay-soak corpus rows confirm `raw_meta: null` → `agent_inferred`.
- Opt-in default-off (env > kwarg > `[memory] mirror_claude_memory` > False); hot path is a cheap path-shape + flag reject. Non-blocking contract (always exit 0) holds.
- Discretion-clean: the only `Haiku`/`subagent` tokens in the diff are pre-existing context lines already on `main` (cli.py, onboard.md); this PR adds none.
- Fixtures synthetic (`tmp_path`/`monkeypatch`, public-domain replay corpus); no `~/.claude` content committed.
- Rebased onto `main` (#980 had landed, overlapping `models.py`/`cli.py` enum additions — auto-merged clean). All 6 commits signed; `tests/test_claude_memory_mirror.py` + corroboration/retention/cli-setup suites green (85 passed) locally. Full CI re-running on the rebased head.

Non-blocking nit (follow-up, not gating): the `INGEST_SOURCE_CLAUDE_MEMORY` docstring in `models.py` says it "Carries `raw_meta["memory_type"]` ... so derive() can map it to an origin." That's not what the implementation does — the origin decision is frozen in `route_overrides` and `raw_meta` is never passed to `derive()`. The comment as written suggests exactly the `raw_meta`-in-`derive()` pattern that would break replay equality; worth correcting to match the (accurate) `derivation.py` comment.

Adding `ready-to-merge` once CI is green.

robotrocketscience:
[release:review:idnn:2026-06-23T04:30:50Z]

robotrocketscience:
[claim:review:Gylf:2026-06-23T04:31:35Z]

robotrocketscience:
Reviewed — approving for merge-train.

**Discretion (the load-bearing check for this feature): clean.**
- Targeted grep over added lines for session/model/subagent markers: empty. The literal `claude` / `claude-memory` / `~/.claude/projects` references are the public feature's own domain naming, not leaked private content.
- Replay corpus (`claude_memory_v0_1.jsonl`) is synthetic — generic facts tagged `provenance: public-domain-v0.1`, `raw_meta: null`. Test fixtures are synthetic (`_SYNTHETIC_FILE`, `build-tool`, `x`). No `~/.claude` content committed.
- No hardcoded real path/username anywhere in the diff; `derive_memory_dir` docstring uses a synthetic `/Users/alice/...` placeholder; the design doc uses `<encoded>`/`<x>` and explicitly cites the `~/.claude`-never-to-a-public-remote rule. The hook reads `~/.claude` only at **runtime** on the user's machine.

**Correctness — verified against source:**
- **Default-off / opt-in** resolver (`is_mirror_enabled`: env > kwarg > TOML > False) mirrors the established retrieval-flag pattern; hook short-circuits on three cheap checks before importing the store.
- **Replay-safe**: the `user_validated` decision for `type:user|feedback` is frozen as a `RouteOverrides` (#265, round-trips replay) rather than living in `derive()` — so the deterministic path (raw_meta nulled) yields `agent_inferred`, no `claude_memory` branch in `derive()`.
- **Non-blocking**: `main` always returns 0, catches all exceptions; OSError/parse-fail return silently. A Write/Edit can never feel broken.
- **Idempotent**: content-derived id + `insert_or_corroborate`.
- **Path classifier** `is_memory_fact_path` is a pure structural match (`.md`, parent `memory/`, under `.claude/projects/`, excludes `MEMORY.md`) — worktree-safe.
- CI green (5247 passed); FF on main; all 6 commits signed.

**On the sourcery edge-case suggestion (resolving as non-blocking):** I reproduced the three cases against the branch's parser — no-closing-fence → `None`, CRLF → `None`, duplicate-keys → first-wins. All **fail safe** (no crash, no malformed belief). The CRLF case means a Windows-written memory file silently doesn't mirror; moot in practice since the upstream tool writes LF on the same host. Worth a one-line follow-up if you want Windows parity, but not a merge blocker.

robotrocketscience:
[release:review:Gylf:2026-06-23T04:36:57Z]

Issue #655, opened 2026-05-11
feat(federation): read-only mechanics — knowledge_deps.json + SQLite ATTACH + foreign-ID rejection (#650 sub-task)


robotrocketscience:
Federation sub-issue of umbrella #650. Per #661 ratification of
Option B (read-only federation), this issue is re-scoped from
multi-writer transport to **read-only mechanics**.

## Summary

Define the read-only federation mechanics: how a local DB opens a
peer DB read-only, where peers are listed, how foreign belief IDs are
distinguished from local ones, and how foreign-ID mutations are
rejected.

### 1. Wire format / transport

**SQLite ATTACH** (or `MemoryStore(f"file:{path}?mode=ro")` for an
equivalent read-only handle). The peer DB file is just a SQLite file
on local disk; no over-the-wire protocol. Federation requires the
peer DB file to be reachable on the local filesystem (NFS / shared
folder / committed-in-repo for tiny global scopes / etc).

### 2. Cadence

N/A under read-only. Every retrieval call opens (or reuses) the peer
DB and reads. There is no "sync" step — data is read directly from
the peer file at query time.

### 3. Auth

None. Read-only access requires only filesystem permissions; aelfrice
adds no auth layer. Trust model is "if you can read the file, you can
see its beliefs."

### 4. Discovery — `knowledge_deps.json`

A committed file at the repo root listing peer DBs:

```json
{
  "version": 1,
  "deps": [
    {"name": "global", "path": "~/.aelfrice/shared/global/memory.db"},
    {"name": "team-shared", "path": "/path/to/team/aelfrice.db"}
  ]
}
```

Path resolution: absolute paths are honored; `~` expands to `$HOME`;
relative paths resolve from the repo root. Missing files at retrieval
time produce a warning and skip; aelfrice does not fail-closed when a
peer is unreachable.

### 5. Conflict-resolution UI

N/A under read-only. There are no conflicts to resolve — foreign
beliefs are read-only display content. Surface display in
`aelf:search` / `aelf:reason` output annotates foreign beliefs with
their origin scope id (e.g. `[scope:global]` prefix or footnote) so
the user knows where they came from.

## Foreign-ID mutation rejection

`apply_feedback(belief_id, ...)`, `lock(belief_id)`, `delete(belief_id)`,
and any other mutation entry point: at the start of each, resolve
which scope owns the belief. If the owning scope is not the current
DB, raise `ForeignBeliefError` with a message naming the owning scope.
[...]

PR #993, opened 2026-06-23
fix(store): exclude soft-deleted beliefs from the federation peer-search lane (#980)

robotrocketscience:
## Summary

Follow-up to #980 / PR #987. Surfaced by the release-prep code audit.

PR #987 added `valid_to IS NULL` at the four **local** retrieval lanes
(FTS `search_beliefs`/`_scored`, BM25 `list_beliefs_for_indexing`, entity-index
`lookup_entities`, locked-L0 `list_locked_beliefs`) and pruned the FTS row on
`soft_delete_belief`. It missed the **fifth** lane: `search_peer_beliefs`
(federation read-only overlay, #655) runs its own FTS query against peer DBs:

```sql
FROM beliefs b JOIN beliefs_fts f ON f.id = b.id
WHERE beliefs_fts MATCH ? AND b.scope IN (...)
```

— no `valid_to` filter, so a soft-deleted belief on a **peer** store still
surfaces through the overlay.

## Fix

Add `AND b.valid_to IS NULL` to the peer query (`store.py:2575`). The local
`soft_delete` FTS-prune can't be relied on across the federation boundary —
a peer may run unpatched code or carry legacy pre-#980 soft-deletes whose FTS
rows persist — so the read-side filter is the robust guarantee.

## Test

`test_search_peer_beliefs_excludes_soft_deleted` sets `valid_to` directly on
the peer (leaving the peer's FTS row intact) to isolate the read-side filter.
Non-vacuity verified: the test **fails** without this change, passes with it.
Full `test_federation_readonly.py` green (13 passed).

## Scope

Completes the #980 item-1 invariant ("soft-deleted ⇒ not retrievable") across
**all five** lanes. Latent today (zero soft-deleted beliefs in stores) but a
correctness gap that should land before the release tag.

## Summary by Sourcery

Ensure federation peer search excludes soft-deleted beliefs from results.

Bug Fixes:
- Filter peer belief search results to omit beliefs that have been soft-deleted in remote stores.

Tests:
- Add a federation read-only test verifying that soft-deleted peer beliefs do not appear in search results.

### DIFF
--- a/src/aelfrice/store.py
+++ b/src/aelfrice/store.py
@@ -2591,6 +2591,7 @@ class MemoryStore:
                     FROM beliefs b
                     JOIN beliefs_fts f ON f.id = b.id
                     WHERE beliefs_fts MATCH ?
+                      AND b.valid_to IS NULL
                     {scope_sql}
                     ORDER BY bm25(beliefs_fts)
                     LIMIT ?
--- a/tests/test_federation_readonly.py
+++ b/tests/test_federation_readonly.py
@@ -98,6 +98,41 @@ def _wire_peer(tmp_path: Path, peer_path: Path, monkeypatch) -> Path:
     return deps_file
 
 
+def test_search_peer_beliefs_excludes_soft_deleted(
+    tmp_path: Path, monkeypatch
+):
+    """A soft-deleted belief on a peer must not surface via the federation
+    overlay (#980 invariant extends to the peer-search lane).
+
+    valid_to is set directly here, leaving the peer's FTS row intact, to
+    simulate a peer that has not run the FTS-pruning soft_delete (or a
+    legacy pre-#980 soft-delete). That isolates the read-side
+    `valid_to IS NULL` filter in search_peer_beliefs: without it, the
+    soft-deleted row still MATCHes and surfaces.
+    """
+    peer_path = tmp_path / "peerA.db"
+    peer = MemoryStore(str(peer_path))
+    try:
+        peer.insert_belief(_belief("peer-active", "shared cat token active"))
+        peer.insert_belief(_belief("peer-removed", "shared cat token removed"))
+        peer._conn.execute(
+            "UPDATE beliefs SET valid_to = ? WHERE id = ?",
+            ("2026-05-12T00:00:00+00:00", "peer-removed"),
+        )
+        peer._conn.commit()
+    finally:
+        peer.close()
+    _wire_peer(tmp_path, peer_path, monkeypatch)
+
+    local = MemoryStore(str(tmp_path / "local.db"))
+    try:
+        ids = [b.id for b, _, _ in local.search_peer_beliefs("shared cat token")]
+        assert "peer-active" in ids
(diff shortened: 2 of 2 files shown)


robotrocketscience:
[claim:review:Gylf:2026-06-23T04:37:35Z]

robotrocketscience:
Resolving the SQL-string-concatenation finding as a **false positive**: `scope_sql = f"AND b.scope IN ({scope_placeholders})"` interpolates only `?` placeholders (`scope_placeholders = ", ".join("?" * len(visible_scopes))`) — the scope *values* are bound as query parameters, and the user `query` goes through `_escape_fts5_query(query)` and `MATCH ?`. There is no untrusted-data path into the SQL text; the interpolation is placeholder-only. This is a pre-existing pattern the diff merely sits adjacent to (the one-line change adds `AND b.valid_to IS NULL`). Sourcery's static rule can't see that the interpolated content is `?`-only.

robotrocketscience:
Heads up (reviewer): your 04:41 rebase onto the #985-inclusive main clobbered a one-line typo fix I'd pushed. **Re-applied on top of your head as f6efa846** — the only change is the docstring `MATCHes` → `matches the FTS query` (the `typos` CI check splits `MATCHes` at the camel-case boundary into `MATC`+`Hes` and hard-fails on `MATC`→`MATCH`). Please don't drop it on a re-rebase; `typos` is red without it. The SQL-concat Sourcery finding is resolved above as a false positive (placeholder-only interpolation). Code fix itself looks correct.

robotrocketscience:
Note for reviewers: the Sourcery `sqlalchemy-execute-raw-query` "SQL injection"
flag on the `{scope_sql}` f-string is a **false positive** and not a required
check.

`scope_sql` is built as `"AND b.scope IN (" + ", ".join("?" * n) + ")"` —
placeholders only, no values interpolated. All values (the escaped MATCH query,
the visible scopes, the limit) are bound as parameters via
`conn.execute(sql, (escaped, *visible_scopes, limit))`. The dynamic part is the
placeholder *count*, which is the standard safe pattern for a variable-length
`IN` clause. This f-string predates this PR; the only line this PR adds inside
it is the constant literal `AND b.valid_to IS NULL`. No untrusted input reaches
the SQL string.

All required staging-gate checks (pytest 3.12/3.13, secrets-scan, history-scan,
migration-policy, release-docs-check) are green.

robotrocketscience:
Merged to main via FF-push (operator-approved bypass of the merge-train). The sole remaining blocker was a red `Sourcery review` check — a verified false positive on pre-existing idiomatic code (line 2597 `{scope_sql}` is placeholder-only `", ".join("?" * len(...))` with values bound as params; the same pattern is on main in ~6 other spots, none suppressed). All real checks (pytest 3.12/3.13, typos, CodeQL, secrets, history) were green; branch FF on main; both commits signed; discretion grep clean.

robotrocketscience:
[release:review:Gylf:2026-06-23T04:55:04Z]

Issue #1011, opened 2026-06-29
feat(ingest): passive conversational capture — ingest on session-end, not PreCompact-only; fix doc-chunk drowning of user facts


robotrocketscience:
## Summary

aelfrice does not passively capture conversational facts into the belief store. A user can state a fact in a session, and a **fresh session asking "what's going on" cannot recall it** — the core promise of the system. Root cause: **conversational ingestion (`aelf ingest-transcript`) fires only on the PreCompact hook (`_handle_pre_compact`).** Sessions that end *without* compacting log every turn to `turns.jsonl` but **never fold those turns into beliefs.** Most sessions end before they compact, so most conversational facts are never captured.

The code already half-acknowledges this: `cadence.py:7` — *"PreCompact-only ('P0' in #749) misses most of the [sessions]"* — but the gap was parked (#749) for the **rebuilder** context block, never closed for the **ingestion** path.

## Evidence (diagnostic run 2026-06-29, local stores)

Each repo gets its own store at `<repo>/.git/aelfrice/memory.db` (worktrees correctly resolve to the shared repo store — not a fragmentation bug). Across 5 independent stores, the most recent **conversation-derived** belief (`origin IN (agent_inferred, user_transcript)`) is days-to-weeks stale, while only explicit-command writes are recent:

| store | last conversation-derived belief | recent writes since |
|---|---|---|
| non-dev project A | 2026-06-19 | none |
| aelfrice | 2026-06-23 | only `speculative` (/wonder) |
| HOME | 2026-06-12 | only `user_stated` (explicit /remember) |
| project B | 2026-06-15 | none |
| project C | 2026-06-12 | none |

**Logged-but-not-ingested, proven live:** in project A, `…/.git/aelfrice/transcripts/turns.jsonl` holds 517 turns, newest timestamped **the same day as this diagnostic** — yet the store's last belief write was **10 days earlier**. The turns are captured by `transcript_logger` in real time and never ingested. The facts the user wanted recalled are sitting in `turns.jsonl`, invisible to retrieval.

[...]

Issue #981, opened 2026-06-23
Restore the deterministic HRR vocabulary-bridge expansion lane behind a default-off flag, then ablate (revisits #605/#897)


robotrocketscience:
## Summary

Restore the deterministic, embeddings-free **HRR vocabulary-bridge expansion lane** (FFT bind/probe over FTS5 seeds) as an explicit **default-off** flag `use_hrr_expand` on `retrieve_v2`, so the single retrieval lane that the v3.0.1 calibration notes name as the cause of the LoCoMo delta can finally be measured in isolation. This issue is scoped to implement-and-ablate, not to flip a default. Flipping the default would reverse #605 and requires the explicit re-decision that #897 deferred.

Cross-refs: #977 (gated-lanes ablation), #605 (determinism ratification), #897 (decision to reaffirm #605), #152 (HRR structural lane, the deterministic precedent).

## What the lane is (and is not)

The predecessor's `retrieve(..., use_hrr=True)` ran an L3 step the docstring labels "HRR vocabulary-bridge expansion" (predecessor `retrieval.py:456-462`): take the top-5 FTS5 hits as seeds, load an HRR graph, probe for top-k semantic neighbors, merge into the candidate set before scoring. Vectors are numpy FFT over a fixed seed (`hrr.py` random_vector + `default_rng`). It is fully deterministic and uses no learned model.

Three distinct artifacts have been conflated under "vocabulary bridge." This issue is **only** about the first:

1. **HRR FFT expansion lane** (THIS issue) — deterministic, produced the 66.1% LoCoMo, currently a no-op in `retrieve_v2`.
2. **`vocab_bridge.py` query-rewrite cascade** (#433/#536) — NOT this. Failed bench (-0.024 NDCG, 0/20 precondition rows, PR #535), removed in v3.0, `use_vocab_bridge` now raises `TypeError`. Do not revive.
3. **`(subject,predicate)`-collision SUPERSEDES** (lab prototype behind the MAB SH90/MH60 projection, #895) — NOT this. Closed wontfix-via-philosophy.

Also distinct: `docs/design/vocabulary_bridge.md` (#227) is a BM25-cosine `find_similar`, unrelated to the HRR lane.

## Evidence the lane is the documented cause

From `benchmarks/results/v3.0.1.json` `_calibration_notes` (verbatim):

[...]

Issue #977, opened 2026-06-23
Benchmark the gated retrieval lanes ON; revisit conservative default-off flags (BFS, type-aware compression, γ/ζ rerank)


robotrocketscience:
## Hypothesis
Several retrieval features ship **default-off**, gated on a benchmark-uplift flip that does not appear to have been run against the full suite. The conservative defaults may be suppressing benchmark scores. We should measure with the gated lanes ON and flip the defaults where uplift is demonstrated.

## Default-off today
- `bfs_enabled` — L3 typed-edge multi-hop graph walk. Off since v1.3.0; flip gated on bench uplift (`docs/design/bfs_multihop.md`). Caveat: temporal coherence for *audit-shaped* queries (`docs/user/LIMITATIONS.md`); does **not** affect recall-mode benchmarks.
- `use_type_aware_compression` (#434) — off until the A2/A4 bench gate clears.
- `use_gamma_posterior_temperature` (γ rerank, #796) — off until a labeled relevance corpus + bench panel show uplift.
- `use_zeta_posterior_rerank` (ζ rerank, #817/#800) — off, mirrors γ.
- #758 adaptive-learning delivery — off.

Default-on for contrast: entity-index, BM25F, HRR-structural, heat-kernel, intentional-clustering.

## Important: documented cause of the v3.0.1 LoCoMo drop
Per the v3.0.1 calibration notes (`benchmarks/results/v3.0.1.json`), the LoCoMo 66.1% → 40.88% drop is attributed primarily to the **#605 ratification** (deterministic, no-embeddings substrate): the original writeup ran a *full HRR-augmented* retrieval path, while v3.0.1 ships **HRR-structural-only**. BFS-off is **not** cited as the cause, so flipping BFS on is unlikely by itself to recover LoCoMo. The MAB SH 90% / MH 60% writeup figures were **lab projections** of a (subject,predicate)-collision SUPERSEDES mechanism not present in the public substrate. Treat all of this as an empirical question, not an assumed win.

## Proposed work
1. Run the canonical suite (LoCoMo, LongMemEval, MAB, StructMemEval) under: current defaults, each gated lane individually ON, and all-ON. One results table.
2. Flip the default for any lane showing uplift without a determinism or latency regression past gate budget.
3. Evaluate `bfs_enabled = true` specifically as a default candidate; scope/document the temporal-coherence caveat.
[...]

PR #1013, opened 2026-06-29
feat(retrieval): default-off origin-tier rerank lane de-ranking doc chunks below user facts (#1011)

robotrocketscience:
## What

Second #1011 AC: a **default-OFF** retrieval lane that de-ranks bulk document-chunk beliefs below user-stated facts. (Sibling of #1012, which fixed capture.)

## Why

The store floods with `filesystem`-origin document chunks — ingested article paragraphs. In one real store, 25,701 beliefs held only **12** `user_stated` facts vs **29,416** document chunks, so retrieval surfaces boilerplate instead of what the user actually said. Even with capture fixed (#1012), the user's facts can be out-ranked by keyword-overlapping article text.

## Changes

- New lane gated by `AELFRICE_USE_ORIGIN_TIER_RERANK` (env) / `[retrieval] use_origin_tier_rerank` (TOML), resolved by `resolve_origin_tier_rerank` (mirrors `resolve_use_zeta_posterior_rerank`).
- When on, `_l1_hits` applies `_origin_tier_boosted` after `_hash_n_boosted` in **both** the BM25F and FTS5 rerank loops — a log-additive per-origin multiplier (`user_stated`/`user_validated` ×3, `user_transcript` ×1.5, `document_recent` ×0.5; unmapped origins ×1.0).
- **Inert by default.** Both byte-identical short-circuits gate on `origin_tier is None`, so when off the lane is a strict no-op.

## Verification

- New suite (10 tests): resolver precedence, `_origin_tier_boosted` unit cases, and end-to-end through `retrieve()` — off keeps the BM25 baseline (document chunk first), on lifts the `user_stated` fact above it.
- **Regression**: full retrieval/ranking suites (`-k "retriev or rank or zeta or gamma or posterior or bm25 or hash_n or l0 or bayesian or heat"`) → **488 passed, 10 skipped, 0 failures**. Byte-identical contracts hold.

## Not in scope — ablation gate

The weight map is **not validated on LoCoMo**. Per the repo's lane discipline (#981 HRR, #977 γ/ζ), the keep/kill decision and weight-tuning require a LoCoMo ablation; this PR ships the lane **off** so it is safe to land and ablate. The ablation is the follow-up before any default-on flip.

Addresses #1011 (signal/noise de-rank). Remaining ACs (`aelf doctor` gap detector, `--since` backfill) stay on the umbrella.

## Summary by Sourcery

Introduce an opt-in origin-tier rerank lane in retrieval that prioritizes user-stated facts over bulk document-chunk beliefs while remaining inert by default.

New Features:
- Add an origin-tier rerank lane that applies per-origin weighting in L1 scoring so user-stated and validated facts outrank bulk document chunks when enabled.

Enhancements:
- Integrate origin-tier weighting into both BM25F and FTS5 L1 rerank paths in a way that preserves byte-identical behavior when the lane is disabled.

Documentation:
- Document the origin-tier rerank lane, configuration flags, and default-off behavior in the v3 changelog.

Tests:
- Add a dedicated test suite covering origin-tier lane resolution, per-origin boosting behavior, and end-to-end retrieval ordering with the lane on and off.

### DIFF
--- a/CHANGELOG/v3.md
+++ b/CHANGELOG/v3.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **Origin-tier rerank lane — de-rank document-chunk beliefs below user-stated facts ([#1011](https://github.com/robotrocketscience/aelfrice/issues/1011)).** Default-OFF opt-in (`AELFRICE_USE_ORIGIN_TIER_RERANK` env / `[retrieval] use_origin_tier_rerank` TOML). When enabled, the L1 rerank (`_l1_hits`) adds a log-additive per-origin multiplier so `user_stated` / `user_validated` facts outrank bulk `document_recent` chunks that merely share keyword overlap — the signal/noise failure where a store flooded with ingested article chunks drowns the few user-stated facts. Inert (byte-identical) when off; the weight map is ablation-pending (LoCoMo) before any default-on flip, mirroring the #981 HRR-lane pattern.
+
 ## [3.7.0] - 2026-06-23
 
 The graph-substrate and phantom-lifecycle wave. The belief graph gained an
--- a/src/aelfrice/retrieval.py
+++ b/src/aelfrice/retrieval.py
@@ -200,6 +200,23 @@ USE_GAMMA_POSTERIOR_TEMPERATURE_FLAG: Final[str] = (
 # decision per issue #817 §"Out of scope" deferred composition).
 USE_ZETA_POSTERIOR_RERANK_FLAG: Final[str] = "use_zeta_posterior_rerank"
 
+# #1011 origin-tier rerank lane. Default-OFF. When ON, the L1 rerank adds a
+# log-additive per-origin multiplier so user-stated facts outrank bulk
+# document-chunk beliefs that merely share keyword overlap — the signal/noise
+# failure where a store flooded with ingested article chunks (origin
+# document_recent / agent_inferred) drowns the few user_stated facts. Inert
+# (byte-identical) when off; the weight map is ablation-pending (LoCoMo)
+# before any default-on flip — mirrors the #981 HRR-lane pattern. Origins
+# absent from the map keep multiplier 1.0.
+ORIGIN_TIER_RERANK_FLAG: Final[str] = "use_origin_tier_rerank"
+ENV_USE_ORIGIN_TIER_RERANK: Final[str] = "AELFRICE_USE_ORIGIN_TIER_RERANK"
+DEFAULT_ORIGIN_TIER_WEIGHTS: Final[dict[str, float]] = {
+    "user_stated": 3.0,
+    "user_validated": 3.0,
+    "user_transcript": 1.5,
+    "document_recent": 0.5,
+}
+
 PLACEHOLDER_FLAGS: Final[tuple[str, ...]] = (
     SIGNED_LAPLACIAN_FLAG,
     POSTERIOR_RANKING_FLAG,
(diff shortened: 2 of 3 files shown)


robotrocketscience:
**LoCoMo cannot measure this lane — empirically confirmed on the real dataset (do not run a full off/on F1 ablation).**

Built the real store (LoCoMo10 `conv-26`, 1240 beliefs via `ingest_conversation` + `retrieve_v2`):

- **Origin distribution: `{'agent_inferred': 1240}`** — 100% one origin. Zero `user_stated`, `user_validated`, `user_transcript`, or `document_recent`.
- The lane only touches those four origins; `agent_inferred` is unmapped → ×1.0. So every belief gets an identical multiplier.
- **Across 30 real questions, the lane changed the retrieved set in 0.** Off == on, byte-identical. A QA F1 ablation would be a guaranteed no-op.

Mechanism: `locomo_adapter.ingest_conversation` calls `ingest_turn` with `role=None`, so the classifier files every turn as `agent_inferred`. LoCoMo has no user-fact-vs-document-chunk contrast, which is the exact thing this lane reranks. (Same class of result as the TTL/HRR "unmeasurable-surface" findings.)

**Right instrument instead:** a pollution-recovery eval — ingest a LoCoMo conversation, inject off-topic `document_recent` chunks into the same store, and measure whether lane-on recovers QA-F1 lost to that injected noise. That exercises the lane's actual function; LoCoMo-as-is does not. Until then the lane correctly ships **default-off**.

Design note surfaced by this: real captured conversational facts (#1012 path) are `user_transcript` (×1.5) or `agent_inferred` (×1.0), and explicit `/remember` facts are `user_stated` (×3). The lane helps the `user_transcript`/`user_stated`-vs-`document_recent` contrast — not the all-`agent_inferred` case. Worth weighing when tuning weights.

robotrocketscience:
**Pollution-recovery eval result: the lane is inert in practice — recommend NOT merging as-is.**

Built the correct instrument (5 LoCoMo convs, evidence-belief recall@budget, lexically-overlapping `document_recent` noise injected; reader-free, deterministic). The instrument is valid — pollution measurably degrades recall:

| budget | recall OFF | recall ON | Δ | best-evidence rank OFF→ON |
|--------|-----------|-----------|---|------|
| 2000 | 0.7513 | 0.7513 | 0.0000 | 7.80 → 7.78 |
| 400  | 0.6466 | 0.6492 | +0.0026 (2 q / 759) | 4.60 → 4.68 (worse) |
| 100  | 0.4869 | 0.4869 | 0.0000 | 1.02 → 1.02 |

(At 1500 noise chunks, recall falls 0.75→0.49 as the budget tightens — the failure mode is real.) **But lane-on does not recover it**: Δrecall ≤ 0.0026 (within noise), rank unchanged.

Why it fails: a ×0.5 demote is −log 2 ≈ −0.69 in the log-additive score, too weak to overcome the BM25 term-frequency advantage of keyword-stuffed noise; and a chunk of retrieval reaches the answer via L0/L2.5/BFS tiers the L1 rerank doesn't touch.

**Recommendation:** don't merge #1013 — a soft origin multiplier at L1 is the wrong lever. The signal/noise AC is better served by **ingest-side separation** (a retention-class / scope so document chunks don't compete in conversational L1 retrieval at all) or a **hard origin gate** (exclude `document_recent` from L1 for conversational queries), not a gentle rerank. The eval here is a reusable instrument for whichever approach replaces it.

Caveat: synthetic noise on LoCoMo; the real-world corpus may differ, but the mechanism (weak demote vs TF, multi-tier bypass) is corpus-independent.

robotrocketscience:
Closing per the pollution-recovery ablation above: the soft origin-rerank is inert in practice (Δrecall ≤ 0.0026 even under heavy targeted document_recent pollution). The signal/noise AC will be pursued via ingest-side separation (retention-class / scope so document chunks don't compete in conversational L1 retrieval) instead — a stronger lever than an L1 multiplier. The pollution eval is the reusable instrument to validate the replacement.

PR #1046, opened 2026-07-01
fix(lifecycle): gate hook auto-install on the running process, not disk presence (#1044)

### LINKED ISSUE #1044: [bug] #834 regression: worktree `uv run aelf` still downgrades installed hooks — `_is_uv_tool_install()` detects install *presence*, not the *running* process
## Summary

#834 was supposed to stop a worktree-local `uv run aelf` from rewriting the user's global `~/.claude/settings.json` hooks. It does not, in the common case where the user **also** has a `uv tool` install. Running any `aelf` subcommand from a source worktree silently **downgraded** installed hooks from **v3.8.0 → v3.6.0**.

Observed message (stderr):
```
aelfrice: hooks updated to v3.6.0 (was v3.8.0) — added: user_prompt_submit_retrieval, transcript_ingest, commit_ingest, session_start, stop_lock_prompt, search_tool, search_tool_bash, pre_issue_guard
```

## Repro

- Global install: `uv tool` at v3.8.0 (`~/.local/bin/aelf --version` → `3.8.0`; `~/.local/share/uv/tools/aelfrice/` exists).
- A source worktree checked out at an **older** release (here `feat/hrr-expand-lane` @ `release: v3.6.0`).

```
cd ~/projects/aelfrice        # worktree @ v3.6.0
uv run aelf wonder "..."      # runs the worktree's v3.6.0 source
# -> rewrites global hooks down to v3.6.0
```

The running interpreter here was **not** the uv-tool env (`sys.prefix` = `/usr/local/opt/python@3.14/...`), yet the merge still ran and clobbered the newer installed surface.

## Root cause

`auto_install_at_cli_entry` gates on `is_running_from_uv_tool_install()` → `lifecycle._is_uv_tool_install()`. That function's **primary** signal is a filesystem existence check:

```python
uv_tools_dir = Path.home() / ".local/share/uv/tools" / PACKAGE_NAME
if uv_tools_dir.exists():
    return True   # <-- fires whenever a uv-tool install exists ANYWHERE on the machine
```

This answers "does a uv-tool aelfrice exist on this box?" — **not** "is *this running process* the uv-tool install?", which is what the docstring on `is_running_from_uv_tool_install()` promises ("True iff the running aelfrice is the user's installed `uv tool` copy"). The `.exists()` branch short-circuits to `True` before the `sys.prefix.startswith(uv_tools_root)` secondary check (which *would* correctly return `False` for a worktree venv) is ever reached.

So for any user who has a uv-tool install, the #834 guard is inert and the original #834 bug is fully reintroduced from any worktree.

`src/aelfrice/lifecycle.py:344-355`, gate at `src/aelfrice/auto_install.py:567`.

## Suggested fix

1. **Make the gate about the running process, not disk presence.** In the `is_running_from_uv_tool_install()` path, drop the `uv_tools_dir.exists()` short-circuit and rely on `sys.prefix`/`sys.executable` resolving *under* the uv tools root. (Keep the `.exists()` check for `upgrade_advice()`, which legitimately asks "how is aelfrice installed" — the two callers want different questions, so they shouldn't share one predicate. Consider splitting into `_uv_tool_install_exists()` vs `_running_from_uv_tool()`.)
2. **Defense in depth: never downgrade.** In `maybe_install_manifest`/`auto_install_at_cli_entry`, if `installed_version` (running) is older than the stamped `prev_version`, skip the merge and emit a one-line notice instead of silently rewriting. A running binary should never stamp the hook surface *backwards*.

## Impact

Silent, not just cosmetic: it flipped 8 default-on hooks back to the v3.6.0 definitions and reset the version stamp, so the user's session loses whatever v3.8.0 changed until they re-run `aelf setup`/`upgrade`. Any contributor with a worktree + a uv-tool install hits this on every `aelf` invocation from the tree.

## Workaround

`AELFRICE_NO_AUTO_INSTALL=1 uv run aelf ...` when invoking from a worktree, and `aelf setup` (from the uv-tool copy) to restore.

robotrocketscience:
Fixes #1044.

## Problem

The #834 guard (a worktree's `uv run aelf` must not rewrite the user's global `~/.claude/settings.json` hooks) was **inert for any user who also had a `uv tool` install**. `lifecycle._is_uv_tool_install()` short-circuits `True` on a filesystem-presence check — `~/.local/share/uv/tools/aelfrice/` existing *anywhere* on the box — **before** the correct `sys.prefix`-under-tools-root check runs. So a source worktree (whose `sys.prefix` is its own venv, not the uv-tool env) was misclassified as the uv-tool install, and every `aelf` invocation from the tree silently rewrote the global hooks **backwards** (observed v3.8.0 → v3.6.0, 8 default-on hooks reset).

## Fix

1. **Gate on the running process, not disk presence.** New `lifecycle._running_from_uv_tool()` returns True iff `sys.prefix`/`sys.executable` resolves under the uv tools root. `auto_install.is_running_from_uv_tool_install()` now delegates to it. `_is_uv_tool_install()` keeps its exists-or-running semantics for `upgrade_advice()`, which legitimately asks "is a uv-tool install present on this box?" (the two callers want different questions, so they no longer share one predicate).
2. **Defense in depth — never downgrade.** `maybe_install_manifest` now skips the merge (non-force path) when the running version is older than the on-disk stamp, emitting a one-line notice instead of stamping backwards. An explicit `aelf setup` (`force=True`) still re-stamps for a genuine downgrade.

## Tests

- `test_running_from_uv_tool_distinguishes_process_from_presence` — the exact repro: uv-tool dir exists on disk but `sys.prefix` is a worktree venv → `_is_uv_tool_install()` True (presence), `_running_from_uv_tool()` False (process), gate stays **False**.
- `test_running_from_uv_tool_true_under_tools_root` — process under tools root → True.
- `test_maybe_install_manifest_never_downgrades` — older running version leaves stamp + settings untouched.
- Updated the delegation test to the new seam. Full affected suites pass (111 tests).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by Sourcery

Guard hook auto-installation based on the currently running uv-tool process instead of mere disk presence, and prevent non-forced runs from downgrading previously installed hook versions.

Bug Fixes:
- Ensure worktree `uv run aelf` invocations no longer rewrite or downgrade the user's globally installed hooks when a separate uv-tool install exists on disk.
- Fix uv-tool detection by distinguishing between the running process being under the uv tools root and a uv-tool install merely existing on the filesystem.
- Prevent `maybe_install_manifest` from downgrading the on-disk hook version when a lower version of the tool is running, unless explicitly forced via setup.

Enhancements:
- Refactor lifecycle uv-tool detection into separate predicates for process-based gating and filesystem-based presence checks to better support upgrade advice logic.

Documentation:
- Document the hook downgrade regression and its fix in the v3 changelog, including the new process-based gating and never-downgrade behavior.

Tests:
- Add regression tests covering process-vs-presence uv-tool detection, the auto-install gate behavior, and the never-downgrade manifest installation guard.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed auto-install behavior so running from a source worktree no longer overwrites existing hook settings.
  * Added protection to prevent older versions from downgrading installed configuration during non-forced runs.
  * Improved detection of when the app is actually running from a uv-managed environment, reducing false matches from files left on disk.

* **Tests**
  * Expanded regression coverage for environment detection and the no-downgrade safeguard.

### DIFF
--- a/CHANGELOG/v3.md
+++ b/CHANGELOG/v3.md
@@ -8,6 +8,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Worktree `uv run aelf` no longer downgrades the user's installed hooks ([#1044](https://github.com/robotrocketscience/aelfrice/issues/1044)).** The #834 guard was inert for anyone who also had a `uv tool` install: `lifecycle._is_uv_tool_install()` short-circuited `True` on a filesystem-presence check (`~/.local/share/uv/tools/aelfrice/` existing *anywhere* on the box) before the correct `sys.prefix`-under-tools-root check ran — so a source worktree's `uv run aelf` was misclassified as the uv-tool install and silently rewrote `~/.claude/settings.json` hooks backwards (observed: v3.8.0 → v3.6.0, 8 hooks reset). Split the predicate: the auto-install gate now delegates to a new `lifecycle._running_from_uv_tool()` (the running-**process** check only), while `_is_uv_tool_install()` keeps its disk-presence semantics for `upgrade_advice()` (which legitimately asks "is a uv-tool install present?"). Defense in depth: `maybe_install_manifest` now refuses to stamp the hook surface **backwards** in the non-force path — a running older version skips the merge with a one-line notice instead of downgrading (an explicit `aelf setup` still re-stamps). Regression tests cover the presence-vs-process split and the never-downgrade guard.
+
 ## [3.8.0] - 2026-06-30
 
 The locks-hygiene and passive-capture wave. Locks gain a frozen/reference tier
--- a/src/aelfrice/auto_install.py
+++ b/src/aelfrice/auto_install.py
@@ -356,6 +356,45 @@ def _result_added_anything(result: object) -> bool:
 # --- main entry ----------------------------------------------------------
 
 
+def _version_key(v: str) -> tuple[int, ...]:
+    """Parse ``'X.Y.Z...'`` into a comparable int tuple.
+
+    Leading digits of each dot-segment; a non-numeric segment contributes
+    0. This is deliberately not full PEP 440 — it only needs to answer
+    "is A older than B" for the never-downgrade guard (#1044).
+    """
+    key: list[int] = []
+    for seg in v.split("."):
+        digits = ""
+        for ch in seg:
+            if ch.isdigit():
+                digits += ch
+            else:
+                break
+        key.append(int(digits) if digits else 0)
+    return tuple(key) or (0,)
+
+
+def _is_downgrade(installed_version: str, prev: str) -> bool:
+    """True iff running ``installed_version`` is strictly older than the
+    on-disk stamp ``prev`` (and prev is a real stamp, not the sentinel).
+    """
+    return prev != _UNSTAMPED and _version_key(installed_version) < _version_key(prev)
+
+
+def _downgrade_skip_result(installed_version: str, prev: str) -> AutoInstallResult:
+    return AutoInstallResult(
+        ran=False,
+        prev_version=prev,
+        new_version=installed_version,
+        message=(
+            f"aelfrice: skipped hook auto-install — running "
+            f"v{installed_version} is older than the installed v{prev}; "
+            f"not downgrading (run `aelf setup` to force)"
+        ),
+    )
+
+
 def maybe_install_manifest(
     *,
     installed_version: str,
@@ -395,6 +434,12 @@ def maybe_install_manifest(
         return AutoInstallResult(
             ran=False, prev_version=prev, new_version=installed_version
         )
+    # Defense in depth (#1044): a running binary must never stamp the hook
+    # surface backwards. The primary gate already excludes worktrees, but
+    # if an older aelfrice ever reaches here (non-force path), skip rather
+    # than silently downgrade the user's installed hooks.
+    if not force and _is_downgrade(installed_version, prev):
+        return _downgrade_skip_result(installed_version, prev)
     target_path = settings_path if settings_path is not None else USER_SETTINGS_PATH
 
     # Acquire exclusive lock on the stamp's parent dir (the stamp file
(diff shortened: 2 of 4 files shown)


robotrocketscience:
[claim:review:Setr:2026-07-01T23:32:54Z]

robotrocketscience:
Reviewed — LGTM, merging.

Both parts of the issue's suggested fix land cleanly:
- **Predicate split**: gate now delegates to `_running_from_uv_tool()` (running-process check via `sys.prefix`/`sys.executable` under the uv tools root), while `_is_uv_tool_install()` keeps disk-presence semantics for `upgrade_advice()`. The two callers ask different questions and no longer share one predicate — this is what fixes the #834 regression for users who also have a uv-tool install.
- **Never-downgrade defense-in-depth**: `_is_downgrade()` uses int-tuple version keys (avoids the `"3.10" < "3.9"` string-compare trap), and the guard is re-checked *after* the under-lock stamp re-read, so it's TOCTOU-safe. `force=True` (explicit `aelf setup`) correctly bypasses it.

Regression coverage is on point (process-vs-presence split, under-root true case, never-downgrade). FF-clean on main, single signed commit, all CI green.

robotrocketscience:
[release:review:Setr:2026-07-01T23:34:35Z]

robotrocketscience:
Review-thread dispositions (reviewer: unblocking for merge):

**Path-containment (Sourcery + CodeRabbit, \`lifecycle.py\`)** — valid, fixed in \`eed247b8\`. \`_running_from_uv_tool()\` now anchors the uv-tools-root comparison with a trailing separator, so siblings like \`.../uv/toolshed\` and \`.../uv/tools-worktree\` no longer satisfy the prefix test. Added a regression test for the sibling case.

**PEP 440 version comparison (CodeRabbit, \`auto_install.py\`)** — declined, with reason:
- \`packaging\` is **not** a declared dependency here — it is transitive-only in \`uv.lock\` and imported nowhere in \`src/\`. Importing \`packaging.version\` directly would either require declaring a new runtime dep (a posture change this project avoids) or trip \`deptry\`'s imported-but-not-declared check.
- \`_version_key\` is a *deliberate, documented* stdlib choice ("deliberately not full PEP 440 — it only needs to answer 'is A older than B' for the never-downgrade guard").
- The flagged trigger (prerelease/partial-version stamps such as \`3.8.0rc1\` or \`3.8\`) does not occur: aelfrice stamps the running \`X.Y.Z\` release version, and releases are full three-part versions. The guard is also defense-in-depth behind a primary gate that already excludes worktrees.

Resolving all three threads.

PR #1069, opened 2026-07-04
feat: temporal spine — ingest-time chronological edges + dedicated retrieval lane, default-off (#1064)

robotrocketscience:
Lands deliverables 1–4 of #1064 — the temporal spine, everything default-OFF. Four atomic commits:

1. **`feat(temporal_spine)` — spine writer.** After each belief insert, link to the session predecessor (`created_at` order, insertion-order tie-break) with `TEMPORAL_NEXT`, src = successor, dst = predecessor, weight 0.8. New `store.session_predecessor_id()` does the indexed O(log n) lookup. Wired into `_ingest_turn_ids` (covers both `ingest_turn` and `ingest_jsonl`) behind `AELFRICE_TEMPORAL_SPINE_WRITE` env → kwarg → `[ingest] write_temporal_spine` TOML → False, mirroring the #988 writer's posture: off-path ingest is byte-identical. Idempotent per `(src, dst, type)`; soft-deleted predecessors stay eligible so chains never sever under GC.

2. **`feat(cli)` — `aelf spine backfill` + doctor row.** Idempotent per-session chain build over existing stores (`--dry-run` supported), hidden subcommand (migration surface, registered in `HIDDEN_SUBCOMMANDS`). `aelf health` / `doctor` graph scope reports `temporal spine: present (N edges) / absent`. Equivalence pinned by test: backfill output == writer output on the same corpus. Live-store dry-run: 21,476 edges across 521 sessions ≈ 0.98 edges/belief, matching the issue's ~1.0 estimate.

[...]

Issue #1064, opened 2026-07-04
Temporal spine: ingest-time chronological edges + dedicated retrieval lane (confirmed +14.6pp coverage on LoCoMo)


robotrocketscience:
## Summary

Add a **temporal spine**: ingest-time chronological `TEMPORAL_NEXT` chains per session, plus a dedicated default-off retrieval lane that traverses them. Confirmed in a two-stage internal R&D campaign (dev + one-shot fresh-corpus confirmatory) as the largest retrieval-coverage gain measured on this codebase, deterministic and embeddings-free — inside the #605 philosophy.

## Evidence

Metric: gold-set coverage = |gold ∩ retrieved| / |gold| per question (the lens that matters for aggregation/temporal questions with diffuse gold; single-hit recall stays saturated and misses this).

**Dev (LongMemEval, 475 questions, wide config l1=200/budget=8000):** spine + graph traversal lifts overall coverage 0.531 → 0.658 (**+12.7pp**); temporal-reasoning +13.6pp; multi-session +11.1pp. A shuffled control — identical edge count, endpoints permuted — gains ~nothing (311× ratio). Deterministic rerun byte-identical.

**Confirmatory (LoCoMo, 1,979 QA with evidence-belief gold sets, one shot, criteria pre-registered):** overall **+14.6pp** (0.460 → 0.606); temporal **+17.2pp**; multi-hop **+10.4pp**; 10× the shuffled control; all-evidence rate 0.133 → 0.260. Out-of-sample gain *exceeded* dev — no winner's curse. Expansion-node budget curve is monotone (0.605 / 0.631 / 0.659 at 32/64/128, ~+2.5pp per doubling, no plateau) — the effect is budget-limited, not substrate-limited. All gating arms byte-identical across fresh ingests.

The mechanism is complementary to lexical matching: prior rounds established that ~84% of the gold missing at wide config shares **zero** salient terms with the question — unreachable by any lexical means. The spine reaches that gold through chronological adjacency to beliefs that *do* match. The shuffled control isolates the cause: it is the chronology, not the extra connectivity.

Historical note: the predecessor codebase added an equivalent writer (`link_temporal`) five days *after* its last quantified benchmark run — this mechanism has never been benchmarked before now, on either codebase.

### 1. Spine writer (ingest-time)
[...]

PR #1074, opened 2026-07-04
test(temporal_spine): G2 top-rank invariance evidence for the #1064 flip gate

robotrocketscience:
## What

Part of #1064 (temporal-spine flip gate). Produces the missing **G2
top-rank invariance** evidence and records the result. The lane + writer
already landed default-off in #1069; this is flip-gate evidence tooling, not
a product change — both flags stay default-off and the issue stays open.

## The G2 question

G2 requires, at the production hook budget (1500 tokens), both a coverage
delta ≥ +3pp **and** *no top-rank regression*. The coverage half was already
measured; the "no top-rank regression" half had no measurement. This adds it.

The spine lane appends its hits **after** the `[locked, l25, l1, hrr]` core
and **before** BFS (`retrieval.py` § temporal-spine lane), so the claim under
test is: turning the lane on must never displace or reorder a core belief — it
may only insert its own hits below the core (and, under budget, evict
BFS-tail items, which is by-design since BFS is the lowest-priority lane).

## Change

- `benchmarks/temporal_spine_ablation.py` gains `--rank-invariance`: a paired
  lane-off/lane-on retrieval per question on the same store. It reads
  `last_lane_telemetry()` to locate the core boundary
  (`locked + l25 + l1 + hrr_expand`) **exactly**, then asserts the core prefix
  is identical between arms and that no dropped baseline belief sits inside the
  core. `RankInvarianceAccumulator` carries the aggregates + a `passed()` gate.
- `tests/test_temporal_spine_ablation.py` (8 tests): pins the accumulator math
  — head invariance, core displacement, core reorder, BFS-tail eviction,
  core-length mismatch, empty-guard, LCP tracking — on hand-built id lists, no
  dataset needed.
- `docs/design/feature-temporal-spine.md`: records the result and splits G2
  into the completed bench half and the still-open shadow-eval half.

## Result — LoCoMo10, `--budget 1500 --l1-limit 50 --rank-invariance`

Coverage (production trim):

| arm | coverage | Δ vs baseline |
|---|---|---|
| baseline | 0.3457 | — |
| +spine | 0.5402 | **+19.45pp** |
| shuffled-control | 0.3504 | +0.47pp |

The gain *survives the production trim* and in fact exceeds the wide-budget
dev figure; the seeded shuffled control recovers almost nothing, so the value
is the chronology, not the added density.

Top-rank invariance (1,986 questions):

- core prefix invariant: **1,986 / 1,986 (100%)**
- top-rank displacements: **0**
- core-length mismatches: **0**
- BFS-tail evictions: 0 (BFS is off here via the #977 expansion gate)
- spine contributed on every question (17,110 spine beliefs added, all below
  the core)
- **G2 top-rank invariance: PASS**

## What this does *not* close

- **G2 shadow-eval half** — aggregate-only run on a real hook-ingested
  backfilled store (open question 1: chain-length distribution under
  production `session_id` semantics).
- **G3 latency** — p50/p95 at ≥10k beliefs; needs a quiet box (the p95 gate
  false-negatives under concurrent load).
- The **default-ON flip** itself is the operator's call once G2-shadow + G3
  clear.

## Reproduce

```
uv run python -m benchmarks.temporal_spine_ablation \
  --data /path/to/locomo10.json --out /tmp/g2.json \
  --budget 1500 --l1-limit 50 --rank-invariance
```

## Summary by CodeRabbit

* **New Features**
  * Added an optional rank-invariance check to the temporal-spine benchmark, with summary metrics shown in the console and included in the JSON report.
  * Improved benchmark output to make core-prefix stability easier to inspect.

* **Documentation**
  * Updated the temporal-spine design notes with current evaluation status and results.

* **Tests**
  * Added coverage for ranking stability, prefix behavior, and regression cases to help prevent changes that alter the expected result order.

### DIFF
--- a/benchmarks/temporal_spine_ablation.py
+++ b/benchmarks/temporal_spine_ablation.py
@@ -38,7 +38,7 @@ import json
 import random
 import time
 from dataclasses import dataclass, field
-from typing import Final
+from typing import TYPE_CHECKING, Final
 
 from aelfrice.ingest import _ingest_turn_ids
 from aelfrice.models import EDGE_TEMPORAL_NEXT
@@ -46,13 +46,14 @@ from aelfrice.retrieval import retrieve_v2
 from aelfrice.store import MemoryStore
 from aelfrice.temporal_spine import backfill_temporal_spine
 
-from benchmarks.locomo_adapter import (
-    CATEGORY_NAMES,
-    DEFAULT_DATA_PATH,
-    LoCoMoConversation,
-    _parse_locomo_datetime,
-    load_locomo,
-)
+# The LoCoMo dataset stack (``locomo_adapter`` → ``nltk``) is imported
+# lazily inside the functions that actually run the bench, so this module's
+# pure logic (``RankInvarianceAccumulator``, ``_lcp_len``) can be imported —
+# and unit-tested — without the benchmark dependency set. Annotations are
+# strings under ``from __future__ import annotations``, so the type-only
+# import lives under ``TYPE_CHECKING``.
+if TYPE_CHECKING:
+    from benchmarks.locomo_adapter import LoCoMoConversation
 
 ARM_BASELINE: Final[str] = "baseline"
 ARM_SPINE: Final[str] = "+spine"
@@ -106,6 +107,8 @@ def ingest_with_evidence_map(
     internal ``_ingest_turn_ids`` because the public ``ingest_turn``
     returns a count only.
     """
+    from benchmarks.locomo_adapter import _parse_locomo_datetime
+
     evidence_map: dict[str, list[str]] = {}
     for session in conv.sessions:
         am_session = store.create_session(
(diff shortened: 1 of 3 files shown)


robotrocketscience:
[claim:review:Gylf:2026-07-04T07:59:54Z]

robotrocketscience:
## Review: APPROVE

Test + docs evidence for the #1064 G2 flip gate. Verified end-to-end:

**Refactor (lazy imports).** `locomo_adapter` (→ `nltk`) is now imported inside the functions that need it, so the pure logic is importable under the ordinary pytest matrix. Confirmed `from __future__ import annotations` is present and the `TYPE_CHECKING`-only `LoCoMoConversation` is used solely as a string annotation — no eager-evaluation `NameError`. Module imports clean; no runtime use of a deferred symbol leaks out.

**Accumulator semantics are correct.** `RankInvarianceAccumulator.add` compares `baseline_ids[:core] == spine_ids[:core]` (catches both drops *and* reorders inside the core), counts core-belief displacements separately, and excludes BFS-tail eviction (rank ≥ core) from regressions — matching the retrieval lane order `[locked, l25, l1, hrr] → temporal_spine → BFS`. `passed()` requires all-invariant ∧ 0 displacements ∧ 0 core-length mismatch ∧ n>0. Core boundary is read from `last_lane_telemetry()`, not inferred.

**API surface verified against the branch.** `retrieve_v2(budget, include_locked, l1_limit, use_temporal_spine)` and `LaneTelemetry.{locked,l25,l1,hrr_expand,temporal_spine}` + `last_lane_telemetry()` all exist.

**Tests.** All 8 pass locally (pure-append invariant, core drop/reorder = regression, BFS-tail eviction = not a regression, core-length mismatch fails, empty-acc fails, LCP tracking). Correctly mirror the implementation.

**Scope.** Docs mark the bench-pool half DONE and correctly leave shadow-eval + G3 + the flip OPEN; PR body is "Part of #1064", not "Closes" — issue stays open. Correct.

CI green (CodeRabbit/Sourcery rate-limited, no substantive automated findings). Discretion grep clean. FF-able, both commits signed. Over the advisory 200-line soft cap but cohesive — non-blocking.

Adding `ready-to-merge`.

robotrocketscience:
[release:review:Gylf:2026-07-04T08:04:12Z]

Issue #1081, opened 2026-07-05
Stored beliefs are more conversational expected


dmayman:
### What happened

When browsing my beliefs for a project, a lot are fragments of "what if"s from conversations that probably don't belong in a belief - or at least needed some follow through to see what was ultimately decided and to have the loose ends discarded. 

Claude's assessment:
 - It's a conversation transcript, not curated knowledge — many entries are half-thoughts ("or should there be other ways to trigger a re-eval?"), and  the score reflects recurrence, not truth or resolution. You can't tell from a belief whether a decision was made or just floated.

There are also lots of out of context fragments that don't really make sense on their own, and I worry that if Claude seeks those out for clarity it'll do weird things. Here are some examples of fragments:

- Use the json, but add prose that describes the template.
- There shouldn't ever be any future.
- We probably need a new concept of a calendar, which is a little bit different than a schedule.
- All averages are not useful.
- Maybe we need one more layer before targets — conceptual targets → weekly/biweekly numerical targets → templates.

### What you expected

idfk man

### Minimal reproduction

```shell
hard to say how to repro this
```

### aelfrice version

3.8.0

### Python version

3.12.4

### Operating system

macOS

### Which surface

aelf CLI, MCP server (Claude Code)

robotrocketscience:
[claim:issue:Toug:2026-07-05T01:29:57Z]

robotrocketscience:
[claim:issue:Kulili:2026-07-05T02:29:12Z]

robotrocketscience:
[release:issue:Kulili:2026-07-05T02:29:20Z]

robotrocketscience:
[claim:issue:Setr:2026-07-05T02:36:44Z]

robotrocketscience:
[release:issue:Setr:2026-07-05T02:36:52Z]

robotrocketscience:
Update after R&D on the root cause: the "beliefs read as a conversation transcript / score reflects recurrence not truth" symptom traces to a **one-directional scoring model** — the belief posterior only moves up (retrieval-exposure `+0.1`), while every demotion/decay sink is dead code or default-off. Measured on a live 24,883-belief store, junk-shaped beliefs mean μ=0.554 vs clean μ=0.446 (junk scores *higher*), 75% frozen at prior, 79% never retrieved.

[...]

Issue #1027, opened 2026-06-30
ingest: ephemeral conversational capture — broad language filter not viable; scope to pure interrogatives


robotrocketscience:
## Investigation (deterministic-only, real store)

Following the #1011 "ephemeral chatter captured as durable beliefs" concern, I characterized the `agent_inferred` beliefs that survive the current `is_transcript_noise` filter: **18,352 survivors**. A read of representative samples shows the population is heavily ephemeral — imperatives ("Run r&d rounds", "Check whether #880 merged"), transient status ("OPEN, MERGEABLE, all CI green", "R0 bench running"), action-logs ("Cleaned up stale worktree"), plan fragments, and questions — with a ~20% minority of durable facts ("Belief() without project_context defaults to ''").

## Why a broad deterministic filter is NOT viable

- **Durable and ephemeral content co-occur in a single belief.** The best language signals still misfire: "contains ?" and even "ends with ?" hit ~10% false positives — multi-sentence beliefs that carry durable info but close with a question (e.g. "**Status:** shipped … landed at v1.5 … ?").
- **No embeddings / LLM-judge** (determinism rule) leaves only surface patterns, which can't separate "transient task" from "durable preference" reliably.
- **The noise is dormant** (R5): only ~0.3% of *retrieved* hits are noise — BM25 suppresses term-poor chatter. So a lossy filter would discard real facts to remove content that doesn't hurt retrieval.
- **The high-precision wins are already shipped:** #785 (assistant-role gate — the biggest structural source), #1025 (harness tags), #1022 (corroboration dedup — de-fanged the `core` channel that noise leaked through).

## Scoped fix that IS clean

**Single-sentence pure interrogatives** (a belief that is one sentence ending in `?`): 430 in the store (2.3% of survivors), **0 durable facts in a 20-sample** — "Want me to run it?", "Which way?", "are there any PRs that need review?". A question is never a durable belief. Multi-sentence beliefs and mid-sentence `?` (abbreviations, quoted questions) are excluded, keeping precision near 100%.

## Plan

Add a single-sentence-interrogative category to `is_transcript_noise` (the ingest-applied filter). Prevents future capture of conversational questions; existing ones are dormant (a backfill GC is tracked separately). Explicitly NOT attempting a broad ephemeral-language filter — documented above as net-negative.

PR #1083, opened 2026-07-05
fix(classification): drop hedged conversational floats at ingest (#1081)

robotrocketscience:
Part of **#1081** — direction A of the four approved (ingest gate, resolution lifecycle, retrieval suppression, backlog cleanup). This is the ingest-gate piece; #1081 stays open as the umbrella for D/C/B.

## Problem

Passive transcript capture stored statement-form musings as `factual` beliefs. `classify_sentence` dropped only trailing-`?` questions (`_is_question`), so hedged declaratives — the reporter's own examples *"Maybe we need one more layer"*, *"We probably need a new concept of a calendar"* — passed every filter and became beliefs. The store filled with half-thoughts that read as conversation, not durable knowledge.

## Fix

`_is_speculative_float`, a deterministic keyword predicate mirroring `_is_question`:

- **Leading hedge** — the sentence *opens* with a hedge marker (`maybe …`, `what if …`, `should we …`, `not sure …`, `i wonder …`, …). A sentence that starts with a hedge is almost always a float; one that merely *contains* "maybe" mid-sentence is not matched.
- **Proposal-hedge phrase** — a small set of unambiguous phrases (`we probably need`, `maybe we`, `we might want`, …) matched anywhere.

Placed **last** in the pipeline (step 6, before default-factual), so a typed requirement / correction / preference is never dropped for being hedged — the type signal wins.

## Scope boundary (honest)

High-precision by design. Two classes are intentionally **kept** and left to #1081's downstream lanes (C retrieval-suppress / D curation):

- mid-sentence hedge words in genuine assertions (*"The test probably fails on slow hardware."*)
- bald hedge-free musings (*"All averages are not useful."*) — no deterministic signal catches these without semantics, which #605 keeps out of aelfrice.

Consistent with the locked #605 philosophy (deterministic, narrow surface, no embeddings) — same class as the existing question filter.

## Behavior / compatibility

- Default-on, unflagged — matches the `_is_question` / #1027 precedent (both drop `persist=False` content with no flag).
- Pure and replay-stable: re-runs identically on `aelf rebuild`. Existing stored floats prune on the next rebuild-from-log (complements direction D); no migration in this PR.

## Tests

10 new atomic tests in `test_classification.py`: per-arm drops, false-positive guards (mid-sentence hedge kept, `should be` vs `should we`, bald musing kept, typed-belief-kept-when-hedged), case-insensitivity, determinism. An end-to-end `ingest_jsonl` check (run locally) confirms a float + a question are dropped while a real assertion + requirement persist. Full suite green (5,639 passed).

## Summary by CodeRabbit

* **New Features**
  * Improved sentence classification to recognize tentative or hedged declarative statements and treat them as non-persistent factual items.
  * Better distinguishes genuine statements from “speculative float” phrasing, including several edge cases and case variations.

* **Bug Fixes**
  * Prevents hedged statements from being incorrectly locked in as persistent classifications.
  * Preserves stronger intent categories when hedging is present, instead of letting the hedge override the main meaning.

### DIFF
--- a/src/aelfrice/classification_core.py
+++ b/src/aelfrice/classification_core.py
@@ -81,6 +81,63 @@ _QUESTION_PREFIXES: Final[tuple[str, ...]] = (
     "could ",
 )
 
+# Statement-form speculative floats (#1081). Unlike `_is_question` these do
+# NOT end in `?` — they are declarative in *form* but tentative in *force*:
+# a floated idea or musing ("Maybe we need X", "What if we tried Y") rather
+# than an assertion. Passive transcript capture was storing these as
+# `factual` beliefs; they read as conversation, not durable knowledge.
+#
+# Two high-precision arms keep false positives low:
+#  - LEADING hedge: the sentence *opens* with a hedge marker. A sentence
+#    that starts with "Maybe " / "What if " / "Should we " (no `?`) is almost
+#    always a float; a genuine assertion that merely *contains* "maybe"
+#    mid-sentence is not matched.
+#  - a small set of unambiguous proposal-hedge PHRASES ("we probably need",
+#    "maybe we") that signal a floated proposal wherever they appear.
+#
+# Applied only to the default-factual bucket (see `classify_sentence` step
+# 6), so a typed requirement / correction / preference is never dropped for
+# being hedged. Bald hedge-free musings ("All averages are not useful.")
+# have no deterministic signal and are intentionally out of scope for this
+# ingest gate — they are handled by the retrieval / curation lanes
+# (#1081 directions C/D), not here. Pure/deterministic: replay-stable.
+_FLOAT_LEADING_HEDGES: Final[tuple[str, ...]] = (
+    "maybe ",
+    "maybe,",
+    "perhaps ",
+    "perhaps,",
+    "what if ",
+    "how about ",
+    "what about ",
+    "i wonder ",
+    "i wonder if",
+    "wondering if ",
+    "not sure ",
+    "i guess ",
+    "i suppose ",
+    "or maybe ",
+    "or perhaps ",
+    "or should ",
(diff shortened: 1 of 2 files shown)


robotrocketscience:
[claim:review:kulili:2026-07-05T05:14:17Z]

robotrocketscience:
Reviewed. Targeted ingest-classification fix for #1081; `classification_core.py` + tests only.

**Verified**
- CI green (pytest 3.12/3.13 + all scans); discretion grep on the diff: clean.
- **Scope / no over-filtering:** the float check is step 6 — *after* requirement/correction/preference detection — and only sets `persist=False` (never changes type, never deletes). `test_typed_belief_kept_even_when_hedged` pins that a hedged requirement/preference survives.
- **False-positive guards** are tested: `test_midsentence_hedge_word_persists` (contains-a-hedge-word ≠ float), `test_leading_should_be_is_not_a_float` (`should be` vs `should we`), and bald hedge-free musings intentionally out-of-scope (`All averages are not useful.` persists) — the honest boundary matching the issue's direction split.
- **Downstream:** `persist=False` routes to `skipped_non_persisting` + the rejection ledger (classification.py:411) — not inserted as a belief, recoverable, mirroring the existing question path.
- **Deterministic / replay-stable:** pure lowercase string prefix/substring matching, no clock or randomness.

**Note:** rebased onto current `main` (advanced when #1085 merged) — clean, no file overlap; single commit re-signed. Good to merge.

robotrocketscience:
[claim:review:Setr:2026-07-05T05:18:42Z]

robotrocketscience:
[release:review:Setr:2026-07-05T05:18:47Z]

robotrocketscience:
[release:review:kulili:2026-07-05T05:22:54Z]

PR #1087, opened 2026-07-05
test(temporal_spine): G5 determinism + ablation-green evidence (#1064)

robotrocketscience:
## G5 — determinism / repro (the last evidence gate before the temporal-spine flip)

Advances **#1064**. Completes **G5**, the final pre-registered evidence gate in the temporal-spine flip criteria (§ Flip gate). With this, **G1, G2 (both halves), G3, and G5 are DONE**; the default-ON flip is blocked only on **G4** — the auto-vs-prompted backfill decision, an operator call. This PR does **not** flip anything and does **not** close #1064.

Both halves are pinned in `tests/test_temporal_spine_repro.py`, stdlib-only and wall-clock-free, so they run in the ordinary pytest matrix (latency/full-bench gates stay run-on-demand — they flake on shared runners, #739/#754).

### Half 1 — two-build byte-identity
The backfill writer is stdlib-only and sampling-free (#605), so a fixed corpus in a fixed insertion order must produce the same `TEMPORAL_NEXT` table every build. `test_two_build_byte_identity` feeds the identical corpus to two independent stores and asserts the serialized spine tables are byte-identical — compared in **stored/rowid order, not sorted**, so any ordering non-determinism (the `created_at` tie-break case is in the fixture) would fail. `test_rebuild_on_built_store_is_byte_identical` asserts a re-backfill changes zero bytes.

### Half 2 — ablation bench green
The full ablation (`benchmarks/temporal_spine_ablation.py`) scores gold-set coverage on LoCoMo and stays run-on-demand. This pins the **mechanism the bench measures** on a controlled fixture, driving the real `retrieve_v2` lane through the bench's **own** scoring accumulators (`CoverageAccumulator`, `RankInvarianceAccumulator`): a gold belief with zero lexical overlap with the query is unreachable lane-off and reachable lane-on via one `TEMPORAL_NEXT` hop → coverage `0.0 → 1.0`, core-prefix invariant, 0 top-rank displacements, lane non-vacuous (candidates ≥ 1).

### Verification
- `pytest tests/test_temporal_spine*.py` → 54 passed locally.
- Discretion grep on the diff: clean.
- Two atomic signed commits (test, then docs).

PR #1090, opened 2026-07-05
feat(temporal_spine): G4 migration — auto-once backfill + reversible spine clear (#1064)

### LINKED ISSUE #1064: Temporal spine: ingest-time chronological edges + dedicated retrieval lane (confirmed +14.6pp coverage on LoCoMo)

robotrocketscience:
## G4 — temporal-spine flip migration (auto-once backfill + reversible `spine clear`)

Resolves **G4** of the #1064 flip gate. Operator call (2026-07-05): **auto-once backfill, not prompted.** This readies the migration; it does **not** flip the defaults (that stays a separate operator release call) and does **not** close #1064.

### Why
The default-ON flip starts the ingest writer chaining *new* beliefs, but existing stores have no spine over their history — so a pure-prompted migration would ship a flip that's inert for exactly the stores that need it (79% of beliefs never retrieved; junk-vs-clean is a *historical* problem). Auto-once delivers day-one coverage; the objections to "auto" (silent, repeating, irreversible) are closed by the sentinel + a clear command.

### Mechanism
- **`maybe_backfill_temporal_spine`** (`temporal_spine.py`) — runs the backfill exactly once per host, sentinel-gated on `~/.aelfrice/spine-backfilled`, wired into `aelf setup` beside the #733 uv-migration. **Gated on the writer flag**, so it's inert while the spine is default-off (pre-flip) and **re-arms**: the first `aelf setup` after the flip turns the writer on fires it once, then the sentinel makes it a no-op. A host that opts the writer on early (env/TOML) triggers it early too. Best-effort store open + never-raises, so it can never break `setup`.
- **`aelf spine clear`** + `MemoryStore.delete_edges_by_type` — reversibility: deletes every `TEMPORAL_NEXT` edge (beliefs untouched). Deterministic backfill (G5) means a later re-backfill rebuilds the identical spine.

### Notes for the reviewer
- **Inert today:** the writer flag is default-off on `main`, so this ships dormant. It self-activates on the flip PR (which changes the default). Verified: `write_enabled=False` → deferred, no sentinel written (re-arms); `True` → runs once + sentinel.
- `spine` gains a `clear` action choice (no new subparser — already in `HIDDEN_SUBCOMMANDS`).
- Sibling PR **#1087** (G5) is a *different* deliverable of the same umbrella, not a duplicate of this.

### Verification
- `pytest tests/test_temporal_spine.py` 47 passed; `test_slash_commands` + setup tests 172 passed.
- New: clear (edge removal / beliefs-intact / clear-then-rebuild byte-identity / type-scoped delete), auto-backfill (deferred-when-off / runs+sentinel / one-shot / re-arms-after-flip / default-gate-reads-flag), CLI `spine clear` dispatch (AELFRICE_DB-isolated).
- Ruff clean on changed files; discretion grep clean.
- Two atomic signed commits (feat, then docs).


robotrocketscience:
[claim:review:Garsecg:2026-07-05T06:44:27Z]

robotrocketscience:
**Review: LGTM — approving via merge-train label.**

Verified against the G4 ratification (auto-once backfill, sentinel-gated, writer-gated, reversible):

- **Scope correct — mechanism only, no default flip.** `is_temporal_spine_write_enabled()` and the lane resolver are untouched; both flags stay default-OFF. The writer/lane default-ON flip remains the separate operator release call. ✅
- **Auto-once semantics.** `maybe_backfill_temporal_spine` short-circuits sentinel → writer-disabled → run, in cheapest-first order. Writer-off path does *not* write the sentinel, so it correctly **re-arms** and fires on the first `setup` after the flip; the sentinel then makes it a no-op. Tests pin all four paths (deferred / runs+writes-sentinel / one-shot / re-arm-after-flip / default-gate-reads-resolver). ✅
- **Reversibility.** `aelf spine clear` → `delete_edges_by_type(TEMPORAL_NEXT)`, beliefs untouched, and `clear + re-backfill` proven byte-identical (leans on the G5 determinism property). Scoped-delete test confirms a non-spine edge survives. ✅
- **Never breaks `aelf setup`.** Backfill wrapped (BLE001), sentinel write wrapped (OSError → still reports `ran=True` since the backfill is idempotent), and the setup call site wraps the whole thing again. Store opened best-effort. ✅
- **CI:** all 21 checks green; both commits signed; mergeState CLEAN; no unresolved review threads. Discretion grep on added lines clean.

Minor (non-blocking): `spine clear --dry-run` silently ignores `--dry-run` (clear always executes). Help text already scopes `--dry-run` to backfill, so acceptable as-is.

robotrocketscience:
[release:review:Garsecg:2026-07-05T06:47:09Z]

PR #1091, opened 2026-07-05
fix(feedback): make retrieval-exposure audit-only by default (#1086)

### LINKED ISSUE #1086: Belief scoring is one-directional: exposure promotes, no organic sink (junk percolates up, not down)
## Problem

The belief posterior (Beta-Bernoulli α/β) only moves **upward** in normal operation. The single wired organic signal is retrieval-**exposure** (`+0.1`), applied every time a belief is surfaced. Every demotion/decay **sink** is dead code or default-off. So a belief that recurs — and therefore gets surfaced — accumulates positive evidence and rises, while one-off real knowledge freezes at its prior. **Junk percolates up, not down.** The Bayesian ranking is one-directional.

Origin: #1081 (belief store reads as a conversation transcript, not curated knowledge — "the score reflects recurrence, not truth or resolution").

## Evidence (measured on a live 24,883-belief store)

- **Posterior μ: junk-shaped beliefs mean 0.554 vs clean beliefs mean 0.446** — junk scores *higher* than real knowledge.
- **75%** of beliefs sit at exactly their initial (deflated) prior — the posterior never moved.
- **79%** were never retrieved → zero feedback → inert forever.
- `feedback_history`: **76,035 positive** events (all `+0.1`, `source=hook` exposure) vs **~979 negative** (all from the `clamp_ghosts` manual batch). ~78:1 promote:demote, and the only demote is a manual sweep.
- mean evidence mass `α+β = 3.0` (median 1.6) — almost nothing accumulates, so decay-toward-prior barely bites.

## Mechanism

| Force | Role | Status |
|---|---|---|
| `HOOK_RETRIEVAL_VALENCE = +0.1` per surfacing (`hook_search.py`; `deferred_feedback.py`, #191) | promote | **LIVE** — its own comment notes *"a retrieval is exposure, not endorsement"* |
| `uri_baki.py` recency-decay + supersession-demote | sink | **dead** — benchmarked honest-negative for retrieval NDCG (#153), never wired; and it was a query-time re-ranker, not posterior percolation |
| `retrieval.py` temporal decay (`temporal_sort`) | sink | **default-off** |
| `clamp_ghosts` | sink | **manual batch only** |

## Scope — four fixes (this umbrella)

1. **Fix exposure-as-endorsement.** Exposure alone must not move μ. Only real signals should: explicit `used`/`harmful`, the agent acting on a belief, user lock/correction, contradiction. This stops recurrence from inflating the truth-posterior.
2. **Add an organic sink.** Cold/unused beliefs decay toward — and below — the prior over time and hibernate out of retrieval, so the store self-prunes without a manual sweep. The symmetric counterpart to the promote signal.
3. **Separate recurrence from truth.** μ means "is this true/useful" (fed only by real signals); recurrence (`corroboration_count`) is tracked as a separate, visible axis, never conflated into the posterior.
4. **Resolution / decided-vs-floated status — with active re-evaluation.** A lifecycle axis recording whether an idea was decided, superseded, or just floated — *and* actively **re-evaluating a floated "maybe" when it later becomes a fact/decision** (promote it; mark the float resolved-by the decision). A common workflow floats many options in conversation and converges on a decision over several turns; today the store captures every float but never the resolution, and can't tell which float became the decision. `EDGE_RESOLVES` exists as a constant but has no producer.

## Acceptance criteria

- On a representative store, junk-shaped beliefs sink **below** clean beliefs in μ (reverse the current 0.554 > 0.446).
- Cold/unused beliefs demote over time with no manual intervention.
- Recurrence no longer inflates the truth-posterior; recurrence remains visible as its own axis.
- A belief's decided-vs-floated status is recorded and queryable.
- A belief floated as a "maybe" and later decided is re-evaluated and promoted, with the float marked resolved-by the decision.
- Determinism preserved (#605): every signal deterministic, no embeddings.
- No regression in retrieval quality on the existing bench pools.

## Non-goals / notes

- **Not** a re-litigation of `uri_baki` query-time re-ranking — #153's honest-negative stands. This is about the stored posterior, which `uri_baki` never touched.
- Ingest-side noise filtering (dropping specific scaffolding shapes at capture) is a separate, lower-leverage lever that overfits to a single workflow's vocabulary; this umbrella is the vocabulary-independent scoring fix. The small deterministic hedge-drop already in review (#1083) is complementary but marginal.
- The fix should be prototyped and **measured against a real store** (does junk sink below clean? how much real knowledge is affected?) before each component lands.

Closes-with: this reframes the scoring half of #1081; #1081 stays for the ingest/curation half.

robotrocketscience:
First step of **#1086** (belief scoring is one-directional). Fixes the *dominant* driver: retrieval-exposure was treated as positive evidence.

## Problem

A hook retrieval bumped the belief's Bayesian posterior (`+0.1` to α via `apply_feedback`). But **a retrieval is exposure, not endorsement** — its own module docstring said so. Counting every surfacing as truth-evidence let whatever *recurs* float above genuine knowledge, with no counter-signal (all decay/demote sinks are dead or default-off).

## Evidence (measured on a real 24,883-belief store)

- Junk (session scaffolding, fragments) carried **~3× the exposure-α of clean beliefs** and scored **higher**: junk mean μ **0.554** vs clean **0.446**.
- The most-exposed junk were recurring prompts surfaced **~240× each**.
- **Factorial counterfactual** (exposure {on/off} × decay-sink {on/off}): removing exposure removes **~60%** of the junk-above-clean gap (+0.053 → +0.021); a recency/decay *sink* is nearly inert here (junk is hot, not stale) and cold-hibernation is counterproductive — so the sink half of #1086 is empirically deprioritised; exposure is the lever.
- **Dual-matrix** (each surfacing = a bet, payoff asymmetric): useful-hit-rate WR ≈ 0.62–0.75; break-even at **k≈2–3**, i.e. the store is net-negative at retrieval once acting-on-junk costs more than ~2–3× a good surface's value.

## Fix

`apply_feedback` gains `update_posterior: bool`. When `False`, the event is still written to `feedback_history` (exposure frequency stays recoverable for the recurrence axis, #1086 fix #3) but the posterior update **and** valence propagation are skipped. `record_retrieval` passes it through, gated by `AELFRICE_EXPOSURE_UPDATES_POSTERIOR` — **default off = the fix**; set to `1` to restore the pre-#1086 behaviour (benchmark A/B and rollback).

## Verification

Controlled: a belief retrieved **50×** no longer outranks a once-seen belief — both stay at prior μ=0.5 (gap **+0.000**); under the legacy flag the recurred belief inflates to μ=**0.857** (gap +0.333). The built fix reproduces the counterfactual prediction.

## Tests

Two tests that encoded the old behaviour (exposure moves α) are updated to the new default and given legacy-flag coverage; three focused `apply_feedback(update_posterior=False)` unit tests added (posterior untouched, audit row still written, no propagation). Full suite: **5,649 passed**.

## Scope

- Does **not** close #1086. Remaining: separate recurrence from truth (#3), decided-vs-floated status with active maybe→decision re-evaluation (#4), and the residual ingest-prior misclassification.
- The `#191` deferred-feedback sweep applies the same exposure-as-evidence pattern (currently inactive — 0 rows) and should get the same treatment in a follow-up.

### DIFF
--- a/src/aelfrice/feedback.py
+++ b/src/aelfrice/feedback.py
@@ -85,6 +85,7 @@ def apply_feedback(
     source: str,
     now: str | None = None,
     propagate: bool = True,
+    update_posterior: bool = True,
 ) -> FeedbackResult:
     """Apply one feedback event to one belief.
 
@@ -92,8 +93,12 @@ def apply_feedback(
     2. Reject zero valence: a no-update event is not a successful update,
        and pre-commit #5 says feedback_history records every successful
        update — so a zero call has no row to write.
-    3. Bayesian-update alpha or beta by valence sign.
-    4. Persist the new posterior on the belief row.
+    3. Bayesian-update alpha or beta by valence sign — UNLESS
+       `update_posterior` is False, in which case the posterior is left
+       untouched (audit-only; #1086). Retrieval-exposure records an event
+       for the recurrence axis without being treated as truth-evidence.
+    4. Persist the new posterior on the belief row (skipped when
+       `update_posterior` is False — nothing changed).
     5. Append one row to feedback_history (created_at = `now` or UTC now).
     6. Propagate the signal through outbound edges (#1058): each
        attenuated delta from `store.propagate_valence` is applied via a
@@ -124,11 +129,18 @@ def apply_feedback(
 
     prior_alpha: float = b.alpha
     prior_beta: float = b.beta
-    new_alpha, new_beta = _bayesian_update(b, valence)
-
-    b.alpha = new_alpha
-    b.beta = new_beta
-    store.update_belief(b)
+    if update_posterior:
+        new_alpha, new_beta = _bayesian_update(b, valence)
+        b.alpha = new_alpha
+        b.beta = new_beta
+        store.update_belief(b)
+    else:
+        # Audit-only (#1086): record the event so exposure frequency stays
+        # recoverable (the recurrence axis), but do NOT move the posterior.
+        # A retrieval is exposure, not endorsement; counting every surfacing
+        # as positive evidence inflated whatever recurs. The posterior is
+        # left exactly as-is; the feedback_history row is still written.
+        new_alpha, new_beta = prior_alpha, prior_beta
 
     timestamp: str = now if now is not None else _utc_now_iso()
     event_id: int = store.insert_feedback_event(
@@ -149,7 +161,7 @@ def apply_feedback(
         source=source,
     )
 
-    if propagate and _propagation_enabled():
+    if update_posterior and propagate and _propagation_enabled():
         deltas = store.propagate_valence(belief_id, valence)
         # Sorted for a deterministic feedback_history row order
         # regardless of edge-iteration order inside the BFS.
--- a/src/aelfrice/hook_search.py
+++ b/src/aelfrice/hook_search.py
@@ -12,11 +12,14 @@ This module exists because:
    to the agent". The README's Bayesian-memory claim depends on this
    loop closing.
 
-2. Hook-driven valence is implicit and weaker than explicit user
-   feedback. A retrieval is exposure, not endorsement. Use a small
-   positive valence (`HOOK_RETRIEVAL_VALENCE = 0.1`) so a thousand
-   retrievals over a few months don't dominate the posterior the way a
-   handful of explicit user thumbs-up should.
+2. A retrieval is exposure, not endorsement. Since #1086 the hook
+   records the exposure event to `feedback_history` (so surfacing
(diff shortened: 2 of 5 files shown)


robotrocketscience:
[claim:review:kulili:2026-07-05T06:43:29Z]

robotrocketscience:
[claim:review:Garsecg:2026-07-05T06:44:17Z]

robotrocketscience:
[release:review:Garsecg:2026-07-05T06:44:22Z]

robotrocketscience:
[release:review:kulili:2026-07-05T06:45:15Z]

PR #1092, opened 2026-07-05
feat(temporal_spine): flip defaults ON (#1064) — HOLD: merge after #1090 + operator go

robotrocketscience:
## Temporal spine → default-ON (the #1064 flip)

⚠️ **HOLD — do not merge yet.** This is staged at the operator's request. Merge only:
1. **after #1090** (the G4 auto-once backfill) lands on `main` — otherwise the flip delivers no historical spine for existing stores; and
2. **on the operator's explicit go** — flipping a global default is a release decision.

This PR is opened as a **draft** to enforce (1)+(2). It will need a rebase once #1090 merges (the design-doc Status line reconciles against #1090's G4 edit).

### What it does
Every pre-registered evidence gate passed — G1 (+14.6pp LoCoMo), G2 (production-budget trim survival + 1,986/1,986 top-rank invariance), G3 (latency delta in band), G5 (two-build determinism + ablation-green) — and G4 migration is resolved (#1090). So both flags flip default-ON:

- `is_temporal_spine_write_enabled` default `False → True` — ingest chains new beliefs by default.
- `is_temporal_spine_enabled` (lane) default `False → True` — retrieval traverses the spine by default.

Per-flag opt-out is unchanged (`AELFRICE_TEMPORAL_SPINE_WRITE=0` / `AELFRICE_TEMPORAL_SPINE=0`, or the `.aelfrice.toml` keys). Existing stores get their historical spine via #1090's sentinel-gated auto-backfill on the first `aelf setup` after upgrade.

### Blast radius (measured)
Full suite: **5651 passed** after updating the tests that pinned the default-off posture. The lane flip is contained because a store with **no `TEMPORAL_NEXT` edges makes the lane a no-op** (empty-lane guard) — every retrieval test on a spineless store is byte-identical. The only fallout was in `test_temporal_spine.py`: default-flag assertions flipped, the ingest off-path became an explicit opt-out (+ a new default-on ingest test), and the two "unrecognised/malformed value not decisive" tests now prove non-decisiveness against a lower rung rather than the default.

### Verification
- Full `pytest`: 5651 passed / 69 skipped / 75 xfailed.
- Discretion grep clean; two atomic signed commits (feat, then docs).

## Summary by CodeRabbit

* **New Features**
  * Temporal spine now defaults to ON for both ingest and retrieval.
  * New stores automatically use the spine behavior after upgrade, with existing stores backfilled on first setup.
* **Bug Fixes**
  * Invalid or malformed settings no longer force the feature off; the system falls back to the default ON behavior.
  * Explicit opt-out remains available for both writing and retrieval.
* **Documentation**
  * Updated release notes and design docs to reflect the new default-ON behavior and opt-out settings.


robotrocketscience:
**Operator-directed split (2026-07-06): the writer half is being carved out into #1111.**

Per operator decision, the temporal-spine flip is split so the write-side ships now while the read-side waits for the retrieval cutover:

- **#1111** flips the ingest **writer** default-ON (`is_temporal_spine_write_enabled` False→True) so stores start accumulating `TEMPORAL_NEXT` edges + G4 auto-backfill fires — de-risking a dense, production-proven spine ahead of the cutover.
- This PR (#1092) should **reduce to the lane-only flip** (`is_temporal_spine_enabled`), which folds into #1107 Phase 2 (garsecg's cutover), where the lane actually reaches the production `retrieve()` path. After #1111 merges, the writer lines here will already be on main.

Rationale in #1111's description. Flagging so this HOLD draft isn't merged as 'both flags' — the writer half is now #1111.

robotrocketscience:
Superseded by **PR #1114** — the #1107 Phase-2 cutover delivers the lane-only flip this draft was reduced to (shim exposes `use_temporal_spine` resolver-driven on the production `retrieve()` path + resolver default flipped ON). Suggest closing this draft once #1114 merges; operator's disposition.

robotrocketscience:
Closing — superseded by #1114, which delivered the lane-half of the flip as #1107 Phase 2 (resolver-driven `use_temporal_spine=None` on the production `retrieve()` path + resolver default flipped ON). The writer-half shipped separately in #1111. Nothing left here to do.

PR #1093, opened 2026-07-05
fix(noise_filter): prune stranded capture noise — orphan headers + shell echoes (#1081)

robotrocketscience:
## What

Passive-captured stores accumulate **stranded capture scaffolding** as standalone belief rows: orphan section headers (a short capitalised phrase ending at a bare colon — `Recommendation:`, `Two paths forward:`) and shell-output echoes (`$ python run_all.py`). This adds `is_stranded_capture_noise` and wires it into the `doctor --prune-noise` GC so they can be soft-deleted (reversible via `valid_to`, dry-run by default).

Closes the ingest-hygiene half of #1081.

## Why it lives on the cleanup path, not `is_transcript_noise`

This was the key finding from the R&D. `is_transcript_noise` runs at the **ingest sentence boundary**, where the sub-floor detector (#785/#809) needs to see a header *in context* — an inline header between two full sentences is demoted to edge `anchor_text`, and only *unanchored* headers are dropped. Adding the header shape to `is_transcript_noise` pre-empts that and destroys anchors (it broke `test_ingest_subfloor_noise` outright).

So the forward path already handles headers correctly. The residue is **historical / non-ingest-path rows** already sitting in the store. `is_stranded_capture_noise` is scoped to exactly that: the cleanup path, where every candidate is an isolated row with no neighbour to anchor, so a header/echo shape is unambiguously stranded.

## Precision (measured on a real 26k-belief store, R&D)

- **High-precision by construction**: a real assertion never ends at a bare colon (`Recommendation: use the cache` survives), and the header pattern requires ≥2 leading chars so kappa formulas (`K:`, `K = 2*N …`) are safe; capped at four words.
- **Deliberately excluded**: the conversational-lead class (`ok …` / `yeah …`). Measurement showed it is FP-prone — in a casual-writing store a real belief can open with a marker (`OK Computer is a great album`), and the bare `k` token matches kappa formulas. That class is left to a caller-side, store-local cleanup, consistent with this module's FP-averse posture.
- End-to-end: `aelf doctor --prune-noise` (dry-run) flags 215 rows on the dev store, all clean orphan headers/echoes in the sample.

## Scope

Ships the **detector + GC wiring + tests** only. Running `--apply` against a live store is a separate, operator-gated step. Workflow-specific coordination noise (PR/rebase/claim chatter) is intentionally **not** shipped here — it is vocabulary-specific and belongs in a local cleanup, not a universal filter.

## Tests

Synthetic only. `is_stranded_capture_noise` unit tests (headers, echoes, and the FP guards: colon-with-content, single-letter label, formulas, lowercase, >4 words, price strings, conversational leads), a separation test pinning that `is_transcript_noise` still leaves the shape to sub-floor, and a `doctor --prune-noise` integration test proving a stored header/echo is soft-deleted while a colon-bearing real assertion survives. Full suite green.

## Summary by CodeRabbit

* **Bug Fixes**
  * `doctor prune-noise` now removes additional stranded capture artifacts, including isolated section headers and shell-style echo lines.
  * These matches are soft-deleted during cleanup, and the command remains safe to preview with dry-run before applying changes.

* **Documentation**
  * Updated the changelog with the new cleanup behavior.

* **Tests**
  * Added coverage for the new noise-detection and prune-noise behavior.


robotrocketscience:
[claim:review:kulili:2026-07-05T07:58:11Z]

robotrocketscience:
[release:review:kulili:2026-07-05T07:59:52Z]

PR #1094, opened 2026-07-05
feat(claude_memory): default-on claude-memory ingest — reconcile + consent gate (#1089)

### LINKED ISSUE #1089: Ingest claude-memory as a default-on source, ranked above conversational capture
## Why

The host's built-in auto-memory ("claude-memory") is only as good as the agent's discipline in *using* it — writing to it is optional, and nothing guarantees a fact the user stated ends up retrievable later. aelfrice's belief graph is the deterministic memory; the two should stay in sync automatically, with the **curated** claude-memory facts trusted at least as much as — and ideally above — beliefs auto-captured from raw conversation transcripts.

## Current state (what already ships)

`#985` (v3.7.0) added the `claude_memory_mirror` hook (`src/aelfrice/hook_claude_memory_mirror.py`), a **one-way** mirror claude-memory → belief graph:

- **Trigger:** `PostToolUse:Write|Edit|MultiEdit` on a per-memory fact file (`.../.claude/projects/<encoded>/memory/<name>.md`, not `MEMORY.md`). It fires only when the agent writes/edits a fact file *in-session*.
- **Provenance mapping:** `metadata.type` `user`/`feedback` → `origin=user_validated` (undeflated prior); `project`/`reference`/absent → `origin=agent_inferred` (deflated prior). Source kind `INGEST_SOURCE_CLAUDE_MEMORY`.
- **Never locks** — L0 stays reserved for explicit `aelf lock`.
- **Opt-in / default OFF:** inert until `AELFRICE_MIRROR_CLAUDE_MEMORY` (env) / `[memory] mirror_claude_memory` (`.aelfrice.toml`) is set.

So the mechanism exists but the *posture* is not what a "memory should just work" default wants: it is off by default, and it is write-event-triggered rather than a regular full reconciliation, so **pre-existing memory files and any writes the hook didn't observe are never ingested.**

## The gap (requested)

1. **Automatic, as a matter of course.** claude-memory should be ingested without the operator flipping a flag — i.e. flip the mirror **default-on** (subject to the collision constraints below) and/or add a **regular full-set reconciliation** (e.g. on `SessionStart`, or a periodic sweep) that ingests the entire current `MEMORY.md` fact-file set, not just in-session write deltas. This closes the "the hook never saw that write" and "the file already existed" holes.
2. **Rank curated memory above conversational capture.** A belief sourced from a claude-memory fact file (deliberately curated) should outrank a belief auto-captured from a chat transcript when both are retrieval candidates.

## Design tensions to resolve in the spec (don't repeat dead ends)

- **Ranking mechanism — the naive lever is already refuted.** A general **origin-tier rerank** lane was built and *refuted* on LoCoMo (`#1013`): reranking candidates by origin/recency did not move retrieval, because weak-match user facts fail as a BM25 *recall* limit (never a candidate), which a reranker can't fix. The deterministic lever that *did* work for "make trusted facts win" was **locking (L0)** / origin+prior, not a rerank weight. So axis 2 must route through provenance→origin/prior (partly already done for `user`/`feedback`) and/or a deterministic pack-order tie-break, **not** a resurrected origin rerank lane. Options to weigh: (a) extend the source→origin/prior mapping so all claude-memory beliefs get an undeflated (or boosted) prior; (b) a deterministic source-priority tie-break in the pack loop favouring `INGEST_SOURCE_CLAUDE_MEMORY` over conversational-capture sources; (c) reconsider the deliberate *never-lock* stance for `user`/`feedback`-typed memory (auto-lock or auto-promote) — currently ruled out by #985, revisit explicitly.
- **Note:** there is currently **no `INGEST_SOURCE_TRANSCRIPT` constant** — conversational/passive capture does not have a first-class source kind the way claude-memory does. Axis 2 may first require *naming* the conversational-capture source so "rank X above Y" is even expressible.
- **Collision avoidance.** aelfrice must complement, not fight, the host's on-by-default auto-memory (the two stores writing over each other's intent). Default-on mirroring must preserve the one-way, non-authoritative contract (#985) and stay idempotent under the regular reconciliation.
- **Default-on gating.** Flipping default-on is a posture change with a privacy surface (it reads memory files unprompted). Confirm the flip is acceptable, or gate it behind first-run consent rather than silent-on.

## Acceptance criteria (draft)

- [ ] claude-memory fact files are ingested **without** an operator flag flip — via default-on mirror and/or a regular full-set reconciliation that also captures pre-existing / previously-missed files.
- [ ] The reconciliation is idempotent (re-running ingests no duplicates; content-hash/dedup holds) and preserves the one-way non-authoritative contract.
- [ ] A belief sourced from claude-memory deterministically outranks an equivalent conversational-capture belief in retrieval, via an origin/prior/tie-break mechanism (NOT a resurrected origin rerank lane; cite #1013).
- [ ] Behaviour is covered by tests analogous to `tests/test_claude_memory.py` + the replay-soak `claude_memory` corpus; no reliance on wall-clock / LLM in the gate.
- [ ] Docs updated: `docs/design/claude_memory_mirror.md`, the hook manifest `default_on`, and the config/env surface.

## Out of scope

- Two-way sync (aelfrice → memory files). The mirror stays one-way; aelfrice is never authoritative over the memory files.
- Re-litigating the refuted origin **rerank lane** (#1013) as a retrieval-time reranker.

Refs: #985 (mirror hook), #1013 (origin-tier rerank refuted → locking/origin is the lever), `docs/design/claude_memory_mirror.md`.

robotrocketscience:
Axis 1 of **#1089** (ingest claude-memory automatically). Fixes the ingestion half: the #985 write-through mirror only saw *in-session* writes and was opt-in, so pre-existing memory files — and any write the hook didn't observe — never reached the belief graph.

Operator decisions this implements (ratified 2026-07-05): enablement = **first-run consent gate** (the G4 auto-once-announced pattern, not silent default-on, not stay-opt-in). Ranking (axis 2) = prior-boost + pack tie-break — **a separate follow-up; this PR does not close #1089.**

## What lands

- **Shared ingest core.** The per-file frontmatter → origin/prior mapping is lifted out of `hook_claude_memory_mirror` into `claude_memory_reconcile.ingest_memory_text`, so the hook and the new sweep share one code path and can't drift from the frozen #985 mapping. Pure refactor — the existing mirror + end-to-end suites are unchanged.
- **Full-set reconcile sweep.** `reconcile_claude_memory` ingests the project's entire `.../memory/<name>.md` set (skips `MEMORY.md`), idempotent (content-derived ids → a re-run corroborates, never duplicates).
- **Consent-gated, default-on-post-consent.** `maybe_reconcile_claude_memory` follows the G4 temporal-spine one-shot pattern (#1064): sentinel-gated per project, safe to call every `aelf setup`, defers (without writing the sentinel) when the mirror is explicitly opted out. The first run is the **consent event** — the sentinel it writes also flips `is_mirror_enabled()` on for that project, so curated memory keeps syncing without a flag flip. An explicit `AELFRICE_MIRROR_CLAUDE_MEMORY=0` / `[memory] mirror_claude_memory=false` always wins.
- **Setup runs it only when there's memory to sync.** `aelf setup` calls the sweep only once the project has a claude-memory dir, so a fresh host that hasn't adopted the auto-memory tool is untouched (no consent granted prematurely, no side effects). It announces the consent event on stderr.
- **Surfaces.** `aelf doctor` gains a `claude-memory mirror: ON/off (reconciled?)` row; `aelf reconcile-claude-memory [--project] [--force]` re-runs the sweep manually (`--force` past the sentinel), for re-ingesting after bulk out-of-session edits.

## Design notes

- **Sentinel lives beside the store** (`db_path().parent/claude-memory-reconciled`), not under `~/.aelfrice`: per-project by construction, sandboxed by `AELFRICE_DB` in tests, and removed with the store on uninstall/rebuild (a fresh store then re-consents — the right behaviour).
- One-way and non-authoritative per #985 — aelfrice never writes back to the memory files.
- The `metadata.type` → origin mapping is unchanged (#985): `user`/`feedback` → `user_validated` undeflated; `project`/`reference` → `agent_inferred` deflated; never locks.

## Out of scope (axis 2, follow-up)

Ranking curated claude-memory above conversational capture. Recorded decision: prior-boost + a deterministic FTS5 origin tie-break (NOT the origin rerank lane refuted in #1013), gated default-OFF pending LoCoMo evidence. The default retrieval path (`posterior_weight==0`) short-circuits to `store.search_beliefs` (`ORDER BY bm25` only), so that's a core-SQL change warranting its own bench-gated PR.

## Verification

- Full suite: **5687 passed**, 69 skipped, 75 xfailed.
- Confirmed zero test pollution of the real `~/.aelfrice` / `.git` (the setup reconcile is gated on the memory dir existing + the db-adjacent sentinel is `AELFRICE_DB`-sandboxed; the migrate-order setup test neutralises the step with one mock).
- New tests: ingest-core mapping, reconcile sweep (all files / skip index / idempotent / missing dir), sentinel-gated `maybe_reconcile` (runs → sentinel; second call short-circuits; opted-out defers), consent-sentinel flips `is_mirror_enabled`, explicit opt-out beats the sentinel, and the CLI subcommand + setup helper (announces when a memory dir exists, defers otherwise).

Closes: part of #1089 (axis 1 only).


robotrocketscience:
[claim:review:Setr:2026-07-06T17:46:35Z]

robotrocketscience:
## Review — LGTM, one low-severity note on `--project`

Solid, well-scoped axis-1 of #1089. Verified against the ratified decisions and the known traps in this area:

- **Matches the ratified #1089 enablement** — first-run consent = the G4 auto-once-announced sentinel pattern (#1064), not silent default-on and not stay-opt-in. Opt-out (`AELFRICE_MIRROR_CLAUDE_MEMORY=0` / `[memory] mirror_claude_memory=false`) always wins; `mirror_opted_out()` vs `is_mirror_enabled()` are correctly distinct (never-consented ≠ opted-out).
- **Replay-safe** — user/feedback ingest freezes the origin/prior via `route_overrides` rather than recomputing from `raw_meta` (which replay nulls). This is the correct avoidance of the raw_meta replay-equality trap.
- **Refactor is behaviour-preserving** — the #985 `metadata.type → origin/prior` mapping moved verbatim into the shared `ingest_memory_text`; the hook now delegates, so the two paths can't drift.
- **Idempotent** — content-derived ids + `insert_or_corroborate`; a re-run corroborates, never duplicates.
- **Hygiene done** — new hidden subcommand registered in `HIDDEN_SUBCOMMANDS`; manifest description updated; setup step extracted to a mockable module-level helper; never-raises discipline throughout so `aelf setup` can't break.

### Note (low severity, non-blocking): `--project` is not store/sentinel-aware

In `_cmd_reconcile_claude_memory`, `--project` feeds only `derive_memory_dir(project_path)` — which memory *files* are read. But `_open_store()` and `reconcile_sentinel_path(db_path())` both resolve the **ambient** store (`db_path()` keys off `$AELFRICE_DB` / cwd's git-common-dir; it takes no project arg). So:

```
cd /project/A            # ambient store = A
aelf reconcile-claude-memory --project /project/B
# → reads B's memory files, ingests them into A's store, stamps A's sentinel
```

The flag help says "the project directory whose claude-memory **store** to reconcile," which over-promises: it selects the memory dir, not the store or sentinel. The common path (no `--project`, run from the target project) is correct, and the tests only exercise `--project` pointing at a dir consistent with the sandboxed ambient store — so the divergence is untested and latent.

Cheapest fix: reword the help to "selects which memory *files* to read into the current store (use `$AELFRICE_DB` or run from the target project to redirect the store)". A fuller fix resolves the store + sentinel from the `--project` git-common-dir. Either is fine as a follow-up — not a merge blocker for a hidden maintenance verb.

### Merge mechanics

Branch is **behind `main`** (needs a rebase before the merge-train FF; likely fell behind when #1097 landed). All commits signed, CI green (the `size-check` CANCELLED is a stale superseded run — the latest passed).

This is a behaviour-changing **default-on-post-consent** PR (unlike the default-off #1097), so deferring the merge call to the operator.

robotrocketscience:
[release:review:Setr:2026-07-06T17:58:23Z]

